PR #10, opened 2019-10-23
Button and HeaderWithWave should follow template and exist in storybook

badrange:
While peeking at the project I noticed that Button and HeaderWithWave were missing from the storybook. 

Added them just to see how this works.

### DIFF
--- a/packages/react/src/components/button/Button.tsx
+++ b/packages/react/src/components/button/Button.tsx
@@ -1,0 +1,9 @@
+import React from 'react';
+
+import styles from './Button.module.css';
+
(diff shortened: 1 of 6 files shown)


packages/react/src/components/button/Button.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
jimppa: Could you remove the `Button.js`-component from this directory?

badrange: Good spot! Removed.

PR #106, opened 2020-01-16
Add icons from events-helsinki-ui

Jokinen:
These icons were generic in nature, but not yet available through HDS.

Here's a list of icons being added  
* icon-calendar
* icon-calendar-add
* icon-directions
* ~~icon-external-link~~ (dropped)
* ~~icon-open-book~~ --> icon-read
* icon-angle-left 
* icon-angle-up
* icon-angle-down
* ~~icon-flag~~ (dropped)
* ~~icon-flag-information~~ (dropped)
* icon-house
* ~~icon-tag~~ (dropped)

I went through all the `svg` files and edited them to conform with patterns the older icons used.

Some of the new icons used `stroke` for colouring. For that reason, I made a slight modification into the story. Along with `fill`, it now also defines `stroke` in its css rules.


Jokinen:
Sure, no problem 👍

Jokinen:
I briefly discussed these icons with @jorilindell. He pointed out that there's an index of icons in Abstract. I cross referenced the icons in this PR with those and noticed that only some can be found from Abstract. I will drop the icons which are not in Abstract, save for a few.

I am thinking about keeping `calendar`, `calendar-add` and `directions` even though they are not in the index, because they look to like they have been designed to fit the HDS.

I am renaming `open-book` into `read` because that's the name for the icon in Abstract.

Jokinen:
Because the CSS styles rely on masks, I ended up runnings the stroked icons through a vector program to transform the strokes into paths. For that reason the  icons in the `react` package now also work with `fill` and I could revert the stroke/strokeWidth styles I previously added to the story file.

@lehtoinen The changes are now ready, could you look at this again? 🙂

lehtoinen:
> I briefly discussed these icons with @jorilindell. He pointed out that there's an index of icons in Abstract. I cross referenced the icons in this PR with those and noticed that only some can be found from Abstract. I will drop the icons which are not in Abstract, save for a few.
> 
> I am thinking about keeping `calendar`, `calendar-add` and `directions` even though they are not in Abstract, because they look to like they are more polished.
> 
> I am renaming `open-book` into `read` because that's the name for the icon in Abstract.

@JoakimSchonert: what's your opinion on adding these new icons? If added, they should be added to the Abstract files as well.

PR #115, opened 2020-02-04
Feature: Add sass library

marharyta:
## Description

Added SASS modules to the design system.
Do not freak out, most of them are just svg files.

## Motivation and Context
Sass is actively used by many Helsinki City Projects

## How Has This Been Tested?

Not tested much yet, please bare with the updates


packages/core-sass/README.md
```diff
@@ -0,0 +1,31 @@
...
+    |   ├── base/
+    |   │   ├── _reset.scss
+    |   |   ├── _typography.scss
+    |   │   ├── <component name>.css
+    |   │   :
+    |   ├── utilities/
+    |   |    ├── _color_variables.scss
+    |   ├── main.scss
+    |
+    └── icons/          # collection of HDS icon styles
```
midori0507: This should stay under dir name assets i think 
``` assets > icons > svg > **```

File: packages/core-sass/src/stylesheets/utilities/_color_variables.scss [hunk 2]
```diff
@@ -0,0 +1,82 @@
+/* BRAND COLORS */
+$hds-brand-color-coat-of-arms-blue: #0072c6;
+$hds-brand-color-coat-of-arms-blue-light-20: #cce2f3;
+$hds-brand-color-coat-of-arms-blue-light-50: #80b8e2;
+$hds-brand-color-coat-of-arms-blue-dark-50: #000098;
+
+$hds-brand-color-coat-of-arms-gold: #c2a251;
+$hds-brand-color-coat-of-arms-gold-light-20: #f2ecdc;
+$hds-brand-color-coat-of-arms-gold-light-50: #e0d0a8;
```
midori0507: This will make duplicated var, cause HDS defined vars in css vars also
--hds--black-80: #black80Hex
$--hds--black-80: #black80Hex

Not single source of truth

packages/core-sass/README.md
marharyta: It was not in the core module

packages/core-sass/src/stylesheets/utilities/_color_variables.scss
marharyta: well, this is something that is not bad actually, if we offer sass partials, the sass variables should be exposable. I used bootstrap-sass as a reference.

PR #120, opened 2020-02-26
[Feature] Minify all icons with svgminify.com

badrange:
## Description

I ran all the svg files through svgminify.com and they are now between 2 and 50% smaller. 

## Motivation and Context

By reducing the size of the svg files downloads of our projects will be faster and we will save CO2.

## How Has This Been Tested?

By opening storybook locally.


badrange:
I will close this PR as I believe all the icons will be replaced anyway. I propose that we minify the new icons when they arrive!

PR #121, opened 2020-03-18
Add tip into documentation for solving missing styles

Jokinen:
## Description

Adds tips into README and site for developers seeking to integrate this library into their project. Currently the documentation doesn't mention any dependencies, but there's at least one requirement for the environment where this package is used: `*.module.css` files should be loaded with CSS modules toggled on.

I do not know if this is by design. If it is and the idea is that this project is only guaranteed to work with Ratkis preferred tooling, the better fix may be to state that the `hds-react` lib is compatible with `create-react-app` based projects.

## Motivation and Context

I attempted to integrate this project with [Varaamo](https://github.com/City-of-Helsinki/varaamo), but failed to get components showing correctly based on the instructions in the documentation.

### DIFF
--- a/site/docs/get_started/for_developers.mdx
+++ b/site/docs/get_started/for_developers.mdx
@@ -49,6 +49,31 @@ Import components:
 import { TextInput } from "hds-react";
 ```
 
+**I am not seeing any styles**  
+`hds-react` uses CSS modules under the hood. If your tooling, for instance webpack, is missing support for CSS modules, components may be left without styles altogether. Here's a quick example of how you can declare support for CSS modules without changing behavior for regular CSS files. These rules assume that the files meant for CSS module support have been named like so: `*.module.css`.
+
+```js
+{
+  test: /\.css$/,
+  exclude: /\.module\.css$/,
+  use: [
+    'style-loader',
+    'css-loader',
+  ],
+},
+{
+  test: /\.css$/,
+  include: /\.module\.css$/,
+  use: [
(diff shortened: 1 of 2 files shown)


site/docs/get_started/for_developers.mdx
```diff
@@ -49,6 +49,31 @@ Import components:
 import { TextInput } from "hds-react";
 ```
 
+**I am not seeing any styles**  
+`hds-react` uses CSS modules under the hood. If your tooling, for instance webpack, is missing support for CSS modules, components may be left without styles altogether. Here's a quick example of how you can declare support for CSS modules without changing behavior for regular CSS files. These rules assume that the files meant for CSS module support have been named like so: `*.module.css`.
```
badrange: To avoid duplication of documentation I suggest we only reference the instructions in hds-react from here. Unless there is a reason that won't work?

badrange: Ah, one of them is in the site, another the README. Hmm! Never mind, I think!

Jokinen: Still a good point imo! I'd rather have click links and have up to date information than read instructions that have gone stale because they have been forgotten.

But yeah, maybe not the right moment to do it 👍

Jokinen:
As far as I understand, this issue is resolved here: https://github.com/City-of-Helsinki/helsinki-design-system/pull/122

So I am closing this PR for now.

PR #131, opened 2020-04-28
Add TextArea and NumberInput components

ahie:
## Description

Refactors the old TextInput component into InputWrapper,
which is used to wrap input fields of different types.

A new TextInput component using InputWrapper is also added,
with an identical interface to the old component.

## Motivation and Context

Venepaikka needs forms with number fields as well as text areas.

## How Has This Been Tested?

Stories added for both components. Verified that the styles work correctly with the added components.


packages/react/src/components/textinput/InputWrapper.tsx
```diff
@@ -0,0 +1,98 @@
...
+  labelledBy?: string;
+  alternative?: boolean;
+  className?: string;
+  defaultValue?: string;
+  disabled?: boolean;
+  helperText?: string;
+  hideLabel?: boolean;
+  invalid?: boolean;
+  invalidText?: string;
+  onChange?: (event: ChangeEvent) => void;
```
jorilindell, line 20: It would be very handy to accept onBlur as prop. For example formik is setting touched value to true after onBlur event. Now it's not possible to show error messages only after touching the field

niglu1: I'm currently updating the TextInput component. With the updated component you're able to pass props to the native input, so this will be possible then.

jorilindell: Great! That's event better.

niglu1:
Support for `number` type inputs and TextArea component will be added in #132. It uses the `InputWrapper` approach used in this PR.

PR #133, opened 2020-05-13
Increase label size for the TextInput component

niglu1:
## Description

Changed the label size for the TextInput component from 14px to 16px and added a small margin.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -72,7 +72,7 @@
     "typescript": "^3.6.3"
   },
   "dependencies": {
-    "hds-core": "^0.6.1",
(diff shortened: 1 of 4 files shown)


packages/react/package.json
```diff
@@ -72,7 +72,7 @@
     "typescript": "^3.6.3"
   },
   "dependencies": {
-    "hds-core": "^0.6.0",
+    "hds-core": "^0.6.1",
```
lehtoinen, line 75: Should this be version locked?

niglu1: Changed.

PR #136, opened 2020-05-14
Fix/input submit and select height

### LINKED ISSUE #134: [Question] Button and input[type="submit"] support
**Is your feature request related to a problem? Please describe.**
I am theming a keycloak theme which comes with a set of ready made templates. At times I need to overwrite these themes to accomplish changes in content--but I try to avoid it as much as I can. Changing the template increases the complexity of version upgrades, because they will require more "merging".

In my case, these templates include forms which use `<input type="submit" />` as the control for submitting the form. The current button component in this library consists of two elements: the button and a label. Naturally I wouldn't be able to add a child node for a submit input.

**Describe the solution you'd like**
When I apply button classes to a `input[type="submit"]` I would get the same button I can find from Abstract.

**Describe alternatives you've considered**
For now, I've applied the button and label classes to the input like so

`<input type="submit" class="hds-button hds-button--primary hds-button__label" />`

and it looks to be working well enough. I am able to inject these classes through a configuration file which means that I don't need to overwrite the template itself, which is in line with my goals.

Is this a good approach? I suspect the child node is a child node for a purpose--maybe it has something to do with icons? Is there something I should look out for when using this approach?

### LINKED ISSUE #135: Checkbox component does not support a multi-row label
**Describe the bug**
When the label of a checkbox spans multiple rows, the whitespace in the layout may be applied improperly.

**To Reproduce**
Steps to reproduce the behaviour:
1. Start your debugging environment
2. Create a an element with a bottom margin of 20px
3. Add a checkbox after it in a manner which causes its label to span multiple rows
4. Notice that the 20px margin between the checkbox and its preceding element is not respected

**Expected behaviour**
When the checkbox label spans multiple rows, whitespace is still respected and the checkbox is aligned to the top of its container (in other words, the checkbox element and its label are "both on the first row") 

A designer asking for this alignment: https://helsinkisolutionoffice.atlassian.net/browse/HP-108?focusedCommentId=13604

**Screenshots**
Screenshot 2020-05-14 at 9 24 29

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser: Chrome
 - Version: 81

niglu1:
## Description
This fixes the issue where the height for the `Checkbox` and `RadioButton` components was fixed, causing problems with long labels.
This also adds support for `type="submit"` inputs to be styled as buttons using `hds-core` classes.

## Related Issues

Fixes #135 
Closes #134 

## How Has This Been Tested?
Tested on storybook

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hds-core",
-  "version": "0.6.1",
+  "version": "0.6.2",
   "description": "Core styles for the Helsinki Design System",
   "author": "Anssi Lehtonen <lehtovaaralainen@gmail.com>",
   "contributors": [
--- a/packages/core/src/components/button/button.css
+++ b/packages/core/src/components/button/button.css
@@ -87,10 +87,16 @@
 
 /* LABEL */
 
+input[type="submit"].hds-button,
 .hds-button__label {
   font-weight: 500;
   line-height: 1.25em;
   padding: var(--spacing-s);
+}
+
+input[type="submit"].hds-button {
+  cursor: pointer;
+  padding: var(--spacing-s) var(--spacing-l);
 }
 
 /* no icons */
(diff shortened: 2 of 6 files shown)


niglu1:
@Jokinen, I fixed the alignment issue. The `text-align` was redundant. I also fixed the focus state by adding a `box-shadow` for submit inputs. It isn't possible to make the space between the focus outline and the button transparent like it is for buttons, but I added a css var (`--submit-input-focus-gutter-color`) that can be used to define the color. It's white by default, but can be changed with the var

Jokinen:
@niglu1 Works nicely now! Thank you.

PR #141, opened 2020-05-20
[Feature] Dropdown component

niglu1:
## Description

Adds Dropdown component to `hds-react`

[Supported props](https://niglu1.github.io/?path=/docs/components-dropdown--default)

The component can be tested [here](https://niglu1.github.io/?path=/story/components-dropdown--default).


packages/react/src/components/dropdown/Dropdown.module.css
```diff
@@ -0,0 +1,333 @@
+@import "hds-core";
+
+.root {
+  --border-radius: 2px;
+  --border-width: 2px;
+  --outline-width: 3px;
+  --dropdown-height: 52px;
```
lehtoinen, line 7: Pixels for borders and outline is a design decision (we use relative units for everything), but should the dropdown-height be relative instead of pixels? That way it would scale with the font-size.

packages/react/src/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,379 @@
...
+  /**
+   * Hides the label above the dropdown
+   */
+  hideLabel?: boolean;
+  /**
+   * Item that should be selected when the dropdown is initialized
+   */
+  defaultValue?: OptionType;
+  /**
+   * Item(s) that should be selected when the dropdown is initialized. Use this instead of `initialSelectedItem` when `multiselect` is enabled
```
lehtoinen, line 52: Should this refer to `defaultValue` instead of `initialSelectedItem`?

packages/react/src/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,379 @@
...
+  disabled?: boolean;
+  /**
+   * If `true`, the dropdown menu can be filtered
+   */
+  filterable?: boolean;
+  /**
+   * Function used to get `aria` attributes and the `id` prop for menu options (`li`)
+   * @param index Data item index
+   */
+  getItemId?: (index: number) => string;
```
lehtoinen, line 38: This gets me confused. This return a single `id` as string, right? I think the `aria` part here is misleading, at least to me.

packages/react/src/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,379 @@
...
+   */
+  label?: string | ReactNode;
+  /**
+   * Used for `aria` attributes and the `id` prop for the input element when `filterable` is `true`
+   */
+  inputId?: string;
+  /**
+   * Used for `aria` attributes and the `id` prop for the menu (`ul`)
+   */
+  menuId?: string;
```
lehtoinen, line 90: `aria` throws me off in these as well. If I understood correctly, the id here is used for stuff like `aria-labelledby` so you probably don't need to state it here if the accessibility just works under the hood :)

Same goes for `toggleButtonId` a few lines below.

packages/react/src/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,379 @@
...
+  /**
+   * Enables selecting multiple values if `true`
+   */
+  multiselect?: boolean;
```
lehtoinen, line 94: Should contain a comment about `multiselect` not working in tandem with `filterable`..?

packages/react/src/components/dropdown/Dropdown.tsx
niglu1: Good catch!

Jokinen:
As a question/comment: is this meant to be used in forms?

niglu1:
> As a question/comment: is this meant to be used in forms? In other words, I would replace `select` with this component?

Yes, I believe that is the primary use case for this component. This component uses [Downshift](https://github.com/downshift-js/downshift) under the hoods and more specifically the [useSelect](https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect) hook for the default behaviour (without filtering or multi-select).

niglu1:
>Version bump required
We're planning on releasing this with some other stuff in the alpha release next Tuesday, so the version bump will be done then.

lehtoinen:
> > Version bump required
> 
> We're planning on releasing this with some other stuff in the alpha release next Tuesday, so the version bump will be done then.

Fair enough 👍

PR #145, opened 2020-05-25
[Feature] Add required field indicator and rich labels

### LINKED ISSUE #140: Consider allowing richer content for checkbox labels
**Is your feature request related to a problem? Please describe.**
We use the checkbox component as a control the user can use to communicate compliance with out instructions or requests. For instance, on one page we a button that allows the user to delete their profile. Before they can do that, they must tick a checkbox

Screenshot 2020-05-18 at 16 23 20

Our designs suggest us to include links directly within the label. When I attempt to do this by using the `<Checkbox />` component in a TS project, I will face a type error. Only text is allowed for the label. This is further highlighted by the naming of the prop which controls content of the label: `labelText`.

**Describe the solution you'd like**
I would like to be able to add more rich content as a description, or label, for a checkbox, that's the type of input. More specifically, I would like to add a link within the label.

**Describe alternatives you've considered**
* Rebuild `<Checkbox />` with classes from `hds-core`
* Add 
```ts
// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
// @ts-ignore
```
 to ignore error

**Additional context**
I was able to find discussion that briefly talked about the accessibility with this approach: https://github.com/FreedomScientific/VFO-standards-support/issues/76

A document is linked, but the section talking about anchors, labels and accessibility has since been removed. I was not able to confirm that an anchor within a label element does not negatively impact accessibility.

niglu1:
## Description

- Added a `required` prop for `TextInput`, `TextArea` and `Dropdown`
  - If `true`, the label is displayed as required. The native elements will be required for text fields.
- Deprecated the `labelText` prop in favour of a new `label` prop (for `TextInput`, `TextArea`, `Dropdown`). The `label` prop accepts strings and react nodes, so you could e.g. include a link in a label (#140)

## Related issues

Closes #140


packages/react/src/internal/field-label/FieldLabel.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC, ReactNode } from 'react';
+
+import styles from './FieldLabel.module.css';
+import RequiredIndicator from '../required-indicator/RequiredIndicator';
+
+type FieldLabelProps = {
+  hideLabel?: boolean;
```
lehtoinen, line 7: Would `hidden`be a better prop name here?

niglu1: Sure would!

PR #153, opened 2020-05-28
[Feature] Breakpoint tokens and documentation

ronijaakkola:
## Description

- Added breakpoint values to tokens
- Added container widths (in each breakpoint) to tokens
- Updated token package build script to include new tokens

- Added documentation for breakpoint and container width tokens
- Added documentation for grid system

We decided to add grid documentation even though HDS does not include implementation of the grid. Reason for this is that grid values that HDS uses can be helpful for developers that use common grid implementations.

## Motivation and Context

Breakpoints and container widths is a highly requested feature. When services utilize these tokens we can achieve better consistency across different services.

## How Has This Been Tested?

Tokens has been tested working by running `yarn build` in design-tokens package.
Documentation has been tested working by running the documentation site locally.


ronijaakkola:
Some fixes done in the latest commit:
- Added mention and link to Grid guidelines to Breakpoints page
- Fixed link naming. Links now better specify what is behind the link.

PR #154, opened 2020-05-28
Add release badge to sidebar

niglu1:
## Description
Added a [shields.io](https://shields.io/) GitHub release badge to the sidebar, so that the version number is automatically updated every time a new release is published.

(thanks for the tip @lehtoinen!)

Screenshot 2020-05-29 at 11 20 44


ronijaakkola:
Could the alpha text be added after "Design System" text? I think it would be important that it is emphasized somewhere. For example:
- Design System Alpha
- Design System (alpha)

PR #202, opened 2020-07-01
HDS-200 | [Feature] Add option to show selected options as buttons

badrange:
## Description

This may or may not solve HDS-200, but at least it could be a step in the right direction. The idea is to show the selected options as buttons (tbd how they look), which makes it easy for the user to see what they have selected and to unselect. 

Closes HDS-200 

## How Has This Been Tested?

This is in Draft

### DIFF
--- a/packages/react/src/components/dropdown/Dropdown.module.css
+++ b/packages/react/src/components/dropdown/Dropdown.module.css
@@ -35,6 +35,8 @@
   --menu-item-icon-color-selected: var(--color-white);
   --menu-item-icon-color-disabled: var(--color-black-40);
   --placeholder-color: var(--color-black-60);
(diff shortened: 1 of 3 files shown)


Jokinen:
We had a chat about this component on Slack in brief. I'll copy some of the things here for transparency:
1) This component should probably be looked at by a designer. The current implementation doesn't conform to any obvious UI pattern.
2) Something that will likely be spotted by the designer, but I still want to point out in more detail is that using <kbd>space</kbd> does not select an option in this implementation. Instead, a space will be written into the filter box, which will in turn likely return different (even zero) results. This will likely make this component hard to use for keyboard navigators who may try to use <kbd>space</kbd> to select options.

PR #227, opened 2020-08-27
[Change] Breakpoint and container width tokens

ronijaakkola:
## Description
Small tweaks were done to breakpoint and container width tokens to allow wider outside margins in edge cases between breakpoints. Other than breakpoint-xl, breakpoints are untouched while container widths between xs and l were made slightly smaller. To see exact values changed, please refer to the tables below.

breakpoint token | old | new
xl | 1224 | 1248

container width token | old | new
xs | 304 | 288
s | 560 | 544
m | 752 | 720
l | 968 | 944
xl | 1200 | 1200

## How Has This Been Tested?
This has been tested by running the documentation site locally.


ronijaakkola:
I assume @niglu1 has taken these new values into account in the Navigation component implementation.

niglu1 (approved):

PR #231, opened 2020-08-28
[Fix] Reference equality comparison in Dropdown multiselect mode

### LINKED ISSUE #228: Controlled multiselect Dropdown does not mark options as selected in the options list
## Describe the bug
When the `Dropdown` component is used in `multiselect` mode as a controlled component, it fails to mark selected options as selected in the dropdown list.

## How to reproduce
Here's a link to a reproduction: https://codesandbox.io/s/agitated-sunset-mh0xe?file=/src/App.js

## Expected behavior
When the `Dropdown` is used in `multiselect` mode as a controlled component, it marks an item as selected in the options list if it's identical to the selected option.

## Possible solution
The cause may be this: when `onChange` changes the state of the component, the references of the options (or options list?) change, which causes the `Dropdown` component to consider two identical options as different. In that case, the problem may be similar to the one that was fixed here: https://github.com/City-of-Helsinki/helsinki-design-system/pull/159

https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/components/dropdown/Dropdown.tsx#L356
Screenshot 2020-08-27 at 12 30 33

Jokinen:
## Description

Instead of reference equality comparison we use the same `isEqual` helper that's used in single select mode. This implementation uses the [`.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method which traverses the array until it finds a match.

## Related Issue

Closes #228

## Motivation and Context

This change allows the `Dropdown` to be used as a controlled component in `multiselect` mode with "variable" options list.

## How Has This Been Tested?

I changed the current "controlled Dropdown" story so that it broke, applied the fix and tested the behaviour using the story.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### React
+#### Fixed
+- [Dropdown] Reference equality comparison in multiselect mode
+    - Previously a selected option would be labeled as selected only if the selected items contained the exact same object. This could result in unexpected behavior for instance when a multiselect dropdown was used as a controlled component and its options list was not static.
+
 
 ## [0.12.1] - August 19, 2020
 ### Core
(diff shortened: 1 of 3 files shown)


niglu1 (approved):
Looks good, thanks a lot! This will be included in the next release.

Issue #224, opened 2020-08-26
onBlur for Dropdown component


badrange:
## Describe the bug

With React, if you create a form using Formik, you want field validation for required fields to happen *after* Formik marks the field as `touched`. Formik uses onBlur to see whether a user has touched a field, and this doesn't work with HDS' Dropdown, my guess is because the Dropdown is based on downshift and not a standard HTML form field. 

## How to reproduce

1. Have a project with Formik Form
2. Add Dropdown
3. Add required fields
4. See that either you get no validation/`invalid` for these fields after visiting them, or you get validation when first touching the form (as in video)

## Expected behavior
Validation works logically

## Possible solution
Somehow ensuring that HDS Dropdowns sends the onBlur event to Formik

## Screenshots

![onBlur-bug](https://user-images.githubusercontent.com/5602596/91282806-786ef300-e792-11ea-90c4-80738ae4985d.gif)

badrange:
I came up with a workaround for our project. We have a wrapper for the HDS dropdown called `FormikDropdown`, and I managed to trick it into triggering the touched by wrapping the Dropdown in a `div` with `onBlur` on it: 

`
  return (
    <div
      onBlur={() => {
        helpers.setTouched(true);
      }}
    >
      <Dropdown
        {...field}
        className={styles.formField}
        defaultValue={defaultValue}
        options={options}
        onChange={handleChange}
        invalid={meta.touched && Boolean(meta.error)}
        helper={meta.touched && Boolean(meta.error) && t(meta.error || '')}
        multiselect={false}
        {...rest}
      />
    </div>
  );
`

niglu1:
It would seem that exposing an `onBlur` callback for the dropdown would fix this issue?

badrange:
Haven't had the time to check, but it is definitely worth giving it a shot.

niglu1:
Fixed in #258

badrange:
Fantastic!

PR #258, opened 2020-09-29
[Feature] Refactor dropdown

niglu1:
## Description

WIP

Closes: #202, #224


eemelinieminen (commented):
Dokumentaation osalta 👍

Jokinen (approved):
Code wise these changes are extensive, but I could not spot any obvious instances of harmful code, although I may be a bit biased.

I also used the test stories of the component to test some basic usecases and some more specific behaviour when it comes tot he keyboard.

PR #288, opened 2020-10-26
[BreakingChange] - Navigation accessibility fixes and improvements

niglu1:
## Description
This is a rather big update to the `Navigation` component containing accessibility improvements and bug fixes.

- The `Navigation.Dropdown` component was completely re-written.
  - The old dropdown used a `role="listbox"` implementation that should only be used in forms. It now uses a simple "menu button" implementation.
- The `Navigation.LanguageSelector` is now always visible in the header.
  - It used to be displayed at the bottom of the menu in mobile.
- The search now needs to be activated when using a keyboard.
  - It used to be activated when the search button received focus.
- The search icon in the input for the `Navigation.Search` component was changed to a button.
- The mobile menu toggle now uses the correct aria attributes.
- The styles were pretty much re-written and now use `scss` instead of `css`
- Now uses `css` animations instead of `react-spring`.

## How Has This Been Tested?
On Storybook and a test app using the local react package as a dependency


aleksik (approved):
👍

PR #295, opened 2020-10-28
[Fix] Combobox accessibility fixes

niglu1:
## Description

- Add some missing aria attributes to the combobox toggle button
- The selection is now cleared if input value doesn't match the selected value

## How Has This Been Tested?
Storybook


aleksik (approved):
👌

PR #298, opened 2020-10-28
[Feature] Allow hiding copyright texts in the footer

niglu1:
## Description
Made it possible to hide the copyright texts in the base section

## How Has This Been Tested?
Storybook


packages/react/src/components/footer/footerBase/FooterBase.test.tsx
```diff
@@ -6,7 +6,9 @@ import { FooterWrapper } from '../../../utils/test-utils';
 
 describe('<Footer.Base /> spec', () => {
   it('renders the component', () => {
```
aleksik, line 8: Could we add a mock for the Date object? Without one we need to update the snapshot on every new year, since the copyright year is dynamic

aleksik (changes requested):

aleksik (approved):
👌

PR #299, opened 2020-10-28
[Feature] Clear the input value on blur if there's no selected item

niglu1:
## Description
Clear the input value on blur if there's no selected item


aleksik (approved):
👍

PR #300, opened 2020-10-29
[Accessibility] Add warning for non-string checkbox labels

aleksik:
## Description

This PR adds warning when using other than strings as a checkbox label.

## Motivation and Context

Links (anchor) embedded in label is poor usability and accessibility practice:

- First, it is very easy for keyboard and screenreader users to accidentally activate the link and/or the checkbox when attempting to click the other one, as users do not necessarily perceive the elements as separate. Secondly, such an unorthodox tag embedding will cause some screenreaders to describe the embedded link in a very unusual manner that is hard to comprehend.

- We recommend that the link be rather placed adjacent but separate to the checkbox. E.g. first the link “City data protection principles”, and below it as a separate checkbox: I have read the city data protection principles (link above).”

## Related Issue

#140

### DIFF
--- a/packages/react/src/components/checkbox/Checkbox.stories.tsx
+++ b/packages/react/src/components/checkbox/Checkbox.stories.tsx
@@ -56,17 +56,6 @@ export const Custom = () => {
 };
 Custom.storyName = 'With custom styles';
 
-export const RichLabel = () => {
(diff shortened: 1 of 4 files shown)


niglu1 (approved):
👍

PR #316, opened 2020-11-13
Bump hds-design-token package version

aleksik:
Bump `hds-design-token` package version in `hds-core` dependencies.


anmipa (approved):
LGTM.

PR #350, opened 2021-01-04
[FIX] Update footer reference images

aleksik:
## Description

There's a dynamic copyright year number in the Footer component. This PR updates the reference images for the new year.


laurakarhu (approved):
all new images here have correctly 2021 and nothing else seems to be broken

PR #364, opened 2021-01-20
[Feature] TimeInput

### LINKED ISSUE #185: Time input field
### What
An input field for inputting time (clock).

aleksik:
## Description

This PR implements the TimeInput component.

## Related Issue

Closes #185


ronijaakkola:
For some reason the TimeInput does not look correct in the documentation site Playground. It is lacking the surrounding border. Any ideas @minevala @VilleMiekkoja why that is?

Screenshot 2021-02-01 at 13 41 14

minevala:
> For some reason the TimeInput does not look correct in the documentation site Playground. It is lacking the surrounding border. Any ideas @minevala @VilleMiekkoja why that is?
> 
> Screenshot 2021-02-01 at 13 41 14

Good catch 👍  It seems that there is some difference in how Storybook processes CSS files. I have not yet figured out why the styles work in Storybook and not in the built library version.ok.

ronijaakkola:
I tested this and the examples now work properly in the documentation site.

packages/react/src/components/timeInput/TimeInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+            onChange={onChange}
+            ref={inputRef}
+            required={required}
+            type={type}
+            tabIndex={0}
+            value={time}
+            {...rest}
+          />
+          <label htmlFor={hourInputId} className={styles.partialInputLabel}>
+            hours
```
VilleMiekkoja, line 274: Could/should we provide this in different languages? fi, sv and en? We could pass a language prop to the component, and render this label based on it?

packages/react/src/components/timeInput/TimeInput.module.scss
```diff
@@ -0,0 +1,82 @@
...
+.partialInputLabel {
+  @extend %hiddenFromScreen;
+}
+
+.partialInput, .divider {
+  border: 0;
+  padding: 0;
+  -webkit-appearance: none;
+  box-sizing: border-box;
+  font-size: 1.125rem;
```
VilleMiekkoja, line 44: Do we have a font-size variable for this?

minevala: `var(--fontsize-body-l) ` is the same but I do not know if there is something more suitable 🤔

packages/react/src/components/timeInput/TimeInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+      inputRef.current.dispatchEvent(event);
+    };
+
+    /**
+     * Merge props.ref to the internal ref
+     */
+    useLayoutEffect(() => {
+      if (ref) {
+        if (isFunction(ref)) {
+          (ref as (instance: HTMLInputElement) => void)(inputRef.current);
```
VilleMiekkoja, line 121: What is happening here?

packages/react/src/components/timeInput/TimeInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+      tooltipLabel,
+      tooltipText,
+      tooltipButtonLabel,
+      type = 'text',
+      ...rest
+    }: TextInputProps,
+    ref?: React.Ref<HTMLInputElement>,
+  ) => {
+    const defaultValues = getDefaultValues(defaultValue);
+    const inputRef = useRef<HTMLInputElement>(null);
```
VilleMiekkoja, line 77: Why do we need to create a new inputRef here, instead of applying the one that is provided through props?

minevala: I guess this ref is for the hidden input.

packages/react/src/components/timeInput/TimeInput.tsx
minevala: I think it the hidden input is set as ref here. I think this can be a hook or a ref created with createRef.

packages/react/src/components/timeInput/TimeInput.tsx
VilleMiekkoja: Yeah I wonder if we could check here if the ref is provided, and then apply it here directly instead of new ref? Then we might not need the stuff done at line 131 where props.ref is merged with the internal ref 🤔

packages/react/src/components/timeInput/TimeInput.tsx
minevala: I added hoursLabel and minutesLabel props.

packages/react/src/components/timeInput/TimeInput.tsx
minevala: I'm not sure would it cause some problems 🤔

packages/react/src/components/timeInput/TimeInput.module.scss
minevala: I guess it would be better to have these as fixed values 🤔

packages/react/src/components/timeInput/TimeInput.tsx
minevala: It probably makes a code bit more complex because the we would need to know every time we use it if it is a Ref or MutableRefObject:
`inputRef || inputRef.current`

minevala: It would be complex or even impossible because the ref prop can be a function returning HTML Element or a RefObject:
`RefCallback<T> | RefObject<T> | null;`

packages/react/src/components/timeInput/TimeInput.tsx
```diff
@@ -0,0 +1,330 @@
...
+import { TextInputProps } from '../textInput/TextInput';
+import textInputStyles from '../textInput/TextInput.module.css';
+import styles from './TimeInput.module.scss';
+import classNames from '../../utils/classNames';
+
+export type TimeInputProps = TextInputProps & {
+  /**
+   * A visually hidden label for the hours. Helps to navigate the component with screen readers.
+   */
+  hoursLabel: string;
```
VilleMiekkoja, line 15: Should we make this not compulsory and default to some value?

minevala: Yes, it is possible but then it would probably require language attribute 🤔

VilleMiekkoja (approved):
Good job! 👍

ronijaakkola:
Made some small improvements and additions to the component documentation.

Issue #354, opened 2021-01-13
Safari shows extra border line between Koro and Footer


juha-ylonen-futurice:
## Describe the bug
Safari shows extra border line between Koro and Footer as it is shown here:
https://hds.hel.fi/components/footer

## How to reproduce
With Safari (Mac Version 14.0.2 (14610.3.7.1.10) Goto:
https://hds.hel.fi/components/footer

## Device information
**Desktop (please complete the following information)**:
>  MacOs
>  Safari
>  Version 14.0.2 (14610.3.7.1.10)

PR #366, opened 2021-01-26
[Fix] Remove unwanted footer "border" in safari

minevala:
## Description
It seems that Safari browser rounds down CSS calc results to pixels but it uses subpixels in measurements:
- koros-height  **14px** + CSS calculated background pseudo-element height **285px** = **299px**
- actual footer height with koros and content **299.53125px**

There is a less than a pixel wide gap ( **.53125px** ) between Footer's koros and content's background-color pseudo-element.
As a fix the pseudo-element and its styles are removed and the footer's content is wrapped with a background-color div.

## Related Issue
https://github.com/City-of-Helsinki/helsinki-design-system/issues/354

## Motivation and Context
The fix removes unwanted border between the footer koros and content in Safari browser

## How Has This Been Tested?
Manually in multiple platform Safaris (in MacOS and in OSX Simulator)

Before with the unwanted border:
Näyttökuva 2021-1-26 kello 12 28 37

After:
Näyttökuva 2021-1-26 kello 12 27 13


laurakarhu (approved):
Seems yo fix the issue :)

PR #372, opened 2021-02-15
[Fix] Added active link state to Navigation examples

ronijaakkola:
## Description
Navigation documentation and Storybook examples were lacking active link items. This was confusing for the reader since it was required to know that a such prop exists for NavigationItem component. 

This PR adds that prop to the first link of every Navigation example in the documentation site and React Storybook.

## How Has This Been Tested?
Tested by running the documentation site and React Storybook locally.

image


laurakarhu (approved):
Active state consistently in examples

PR #377, opened 2021-02-18
[Fix] HDS-595: Fix ssr bug in navigation component

nenonja:
## Description
I tried to fix the ssr bug in minimal changes to the component. Idea is to avoid showing different dom in mobile and desktop view. This can be done via css-styles and hiding & showing different parts of dom depending on media size. Background of the problem can be found forexample here: 

To prevent reinventing the wheel, I decided to use @artsy/frensel library: https://github.com/artsy/fresnel which does this in declarative way.

Work in Progress! Dont review this yet.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-595
https://helsinkisolutionoffice.atlassian.net/browse/TH-897

Closes #


packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hds-react",
-  "version": "0.21.0",
+  "version": "0.21.1",
```
VilleMiekkoja, line 3: This change can be omitted; we will update the version when we make a release

packages/react/package.json
```diff
@@ -16,7 +16,7 @@
     "base.min-*"
   ],
   "scripts": {
-    "build": "rimraf lib/ && yarn run lint && yarn run tcm && rollup -c",
```
VilleMiekkoja, line 19: Lint is removed?

Sartonon: Yeah my bad, deleted it for debugging purposes.. Maybe @nenonja you can add it back? :)

packages/react/src/components/navigation/navigationUser/NavigationUser.tsx
```diff
@@ -34,49 +35,57 @@ export const NavigationUser = ({
...
+  const userItems = (
     <>
-      {userName && (
-        <span className={styles.userName}>
-          <IconUser aria-hidden />
-          {userName}
-        </span>
-      )}
-      {children}
+      <Visible above="m">
```
VilleMiekkoja, line 45: Think this should be on mobile, so below="m"? Also row 54 should be on desktop, so above="m" there.

packages/react/src/styles/layout.scss
```diff
@@ -0,0 +1,90 @@
+//
+//  MEDIA QUERIES
+//––––––––––––––––––––––––––––––––––––––––––––––––––
+
+// A map of breakpoints.
+$breakpoints: (
```
VilleMiekkoja, line 6: It seems that breakpoints are already defined at least in 2 places already in hds, so maybe we could apply them from there instead of re-defining them here? Currently they are defined in these files:

https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/design-tokens/tokens/breakpoint/breakpoint.json (these are generated into all.scss and breakpoint.scss files)

https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/styles/breakpoints.css

packages/react/src/styles/layout.scss
```diff
@@ -0,0 +1,90 @@
...
+    @warn 'Invalid breakpoint: #{$breakpoint}.';
+  }
+}
+
+//
+//  RESPOND BETWEEN
+//––––––––––––––––––––––––––––––––––––––––––––––––––
+
+// @include respond-between(sm, md) {}
+@mixin respond-between($lower, $upper) {
```
VilleMiekkoja, line 64: Seems that this mixin is not applied at the moment?

packages/react/src/components/navigation/Navigation.tsx
```diff
@@ -182,7 +195,7 @@ export const Navigation = ({
 
   const [{ authenticated, navigationVariant }, dispatch] = useReducer(reducer, {
     authenticated: false,
-    navigationVariant: 'default',
+    navigationVariant: getNavigationVariantFromChild(children),
```
VilleMiekkoja, line 198: Is this called before the children are rendered? I wonder if it can find the children yet here?

Sartonon: On first render it is called before children are rendered for sure. But for react to find components children I can't see why it should render them before. They are basically just objects passed as children to Navigation component.

packages/react/package.json
nenonja: fixed

packages/react/src/components/navigation/navigationUser/NavigationUser.tsx
nenonja: fixed

packages/react/package.json
nenonja: reverted

packages/react/src/styles/layout.scss
nenonja: I have to check does `<Visible above="s" below="l">Hello</Visible>` work correctly or not, and do we need this. Have to write some unit tests :)

packages/react/src/styles/layout.scss
VilleMiekkoja: Could we take the variables from design-tokens and just apply them inside this file?

nenonja: Let's try! I have to improve my sass knowledge, it seems that json can be imported to the file.

packages/react/src/internal/visible/Visible.tsx
```diff
@@ -0,0 +1,27 @@
+import classNames from 'classnames';
```
VilleMiekkoja, line 1: I wonder if we could apply our own classNames util (https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/utils/classNames.ts) instead of adding this new library?

packages/react/src/components/navigation/navigationRow/NavigationRow.tsx
```diff
@@ -20,9 +19,9 @@ export type NavigationRowProps = React.PropsWithChildren<{
 }>;
 
 export const NavigationRow = ({ variant = 'default', children }: NavigationRowProps) => {
-  const { dispatch } = useContext(NavigationContext);
-
-  useEffect(() => dispatch({ type: 'NAVIGATION_ROW', value: variant }), [dispatch, variant]);
+  // Can be deleted??
+  // const { dispatch } = useContext(NavigationContext);
```
VilleMiekkoja, line 23: Now that we get the variant already inside the parent component, we should be able to remove this. Also we could probably change the reducer in the parent component (Navigation.tsx) to a more simple useState call, since we no longer have complex structure there?

VilleMiekkoja: This is done now

Sartonon: Should we handled situation where the variant is changed on the fly after first render?

VilleMiekkoja: Ah good point. As it is now if variant is changed parent component won't notice it. So we actually need to update the parent component here. I can work on it

VilleMiekkoja: Fixed

packages/react/src/styles/breakpoints.css
```diff
@@ -1,3 +1,6 @@
+
+
```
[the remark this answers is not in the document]
VilleMiekkoja: Removed these

packages/react/src/components/navigation/Navigation.tsx
```diff
@@ -173,15 +173,15 @@ export const Navigation = ({
   const [mobileMenuOpen, setMobileMenuOpen] = useState(menuOpen);
 
   const [authenticated, setAuthenticated] = useState(false);
-  const navigationVariant = getNavigationVariantFromChild(children);
+  const [navigationVariant, setNavigationVariant] = useState(getNavigationVariantFromChild(children));
```
Sartonon, line 176: Should we pass callback to useState to not call the func on every render?
`useState(() => getNavigationVariantFromChild(children));`

PR #395, opened 2021-03-12
Fix navigation logout SSR-bug

minevala:
## Description
When the user logouts, the server-side rendered Navigation throws an error. A menu button is missing the component ref. 

## Related Issue

Closes #


VilleMiekkoja (approved):
Looks good 👍

PR #404, opened 2021-03-23
Feature/dialog

minevala:
## Description
Dialog component

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-557
Closes #
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-557

## How Has This Been Tested?
- On developers machine locally
- Automated unit tests

Näyttökuva 2021-4-6 kello 14 06 26
Näyttökuva 2021-4-6 kello 14 06 46

## Demo
👉  [Storybook demo](https://city-of-helsinki.github.io/hds-demo/dialog/)


packages/react/src/components/dialog/Dialog.module.scss
```diff
@@ -0,0 +1,71 @@
+@import '../../styles/common';
```
VilleMiekkoja, line 1: The file inside styles folder seems to be named with underscore prefix: _common. But the import seems to work also without the underscore?

packages/react/src/components/dialog/Dialog.module.scss
```diff
@@ -0,0 +1,75 @@
...
+  right: 0;
+  top: 0;
+  width: 100%;
+}
+
+.dialog {
+  background: white;
+  box-sizing: border-box;
+  margin: auto;
+  max-height: max-content;
```
VilleMiekkoja, line 50: Just out of curiosity, what does this mean? :)

packages/react/src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { DialogHeader } from './dialogHeader/DialogHeader';
+import { DialogContent } from './dialogContent/DialogContent';
+
+export interface DialogCustomTheme {
+  '--accent-line'?: string;
+  '--color-overlay'?: string;
+  '--width'?: string;
+}
+
+export type DialogProps = React.PropsWithChildren<{
```
VilleMiekkoja, line 20: Is this React.PropsWithChildren necessary? I thought it helps in that we would not need to define the children prop ourself? But since we are defining it at line 56, I wonder if this adds value for us?

packages/react/src/components/dialog/__snapshots__/Dialog.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Dialog /> spec renders the component 1`] = `<DocumentFragment />`;
```
VilleMiekkoja, line 3: This feels like surprisingly short snapshot? One could expect that the size of Dialog snapshot would be bigger? Hmm.

packages/react/src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+  children: ReactNode | ReactNodeArray;
+}>;
+
+export const Dialog = ({ id, isOpen, children, close, className, focusAfterCloseId, theme, ...props }: DialogProps) => {
+  const customThemeClass = useTheme<DialogCustomTheme>(styles.dialogContainer, theme);
+  const dialogRef: RefObject<HTMLInputElement> = React.createRef();
+
+  const { 'aria-labelledby': ariaLabelledby, 'aria-describedby': ariaDescribedby } = props;
+
+  const onKeyDown = (event: KeyboardEvent): void => {
```
VilleMiekkoja, line 65: Small improvement to readability: We could maybe move this closer to the useEffect function where it is used? I see that setFocusAfterClose is also applied inside the useEffect though. Could this function come after the useEffect? There might be some lint rule that does not allow it though...

packages/react/src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+        ref={dialogRef}
+        role="dialog"
+        aria-modal="true"
+        id={id}
+        className={classNames(styles.dialog, className)}
+        aria-labelledby={ariaLabelledby}
+      >
+        <div aria-describedby={ariaDescribedby}>
+          {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+          <div onFocus={onTabBarrierFocus(true)} tabIndex={0} />
```
VilleMiekkoja, line 127: What determines the initially focused element?

packages/react/src/components/dialog/dialogActionButtons/DialogActionButtons.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactNode, ReactNodeArray } from 'react';
+
+import styles from './DialogActionButtons.module.scss';
+
+export type DialogActionButtonProps = React.PropsWithChildren<{
```
VilleMiekkoja, line 5: I wonder if the PropsWithChildren adds value here?

VilleMiekkoja:
Remember to add the component to rollup config:
https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/rollup.config.js

packages/react/src/components/dialog/Dialog.module.scss
minevala: https://ishadeed.com/article/intrinsic-sizing-in-css/#max-content
I will remove this, it probably has no effect.

packages/react/src/components/dialog/Dialog.tsx
minevala: It is now fixed in the Dialog.Header but I could get rid of the Header component. 
Then the focus handling would be in one place.

packages/react/src/components/dialog/dialogActionButtons/DialogActionButtons.tsx
minevala: No. I'll fix these.

packages/react/src/components/dialog/__snapshots__/Dialog.test.tsx.snap
minevala: Yes, there are some problems with how the React Testing library's render works with portals and snapshots. 
I have not yet figured out how to render without following warnings: 
```
 Warning: render(): Rendering components directly into document.body is discouraged, since its children are often manipulated by third-party scripts and browser extensions. This may lead to subtle reconciliation issues. Try rendering into a container element created for your app.
 ```

File: packages/react/src/components/dialog/Dialog.tsx (hunk 3 above)
Comment (minevala) [reply packages/react/src/components/dialog/Dialog.tsx:20]: I'll fix this.

File: packages/react/src/components/dialog/Dialog.tsx (hunk 5 above)
Comment (minevala) [reply packages/react/src/components/dialog/Dialog.tsx:65]: I'll fix.

File: packages/react/src/components/dialog/Dialog.module.scss (hunk 1 above)
Comment (minevala) [reply packages/react/src/components/dialog/Dialog.module.scss:1]: Good catch 👍
Hmm.., I think we should use the underscore name. IIRC, the underscore tells the SASS that there is no need to generate a new CSS file. But this must be confirmed.

File: packages/react/src/components/dialog/Dialog.tsx [hunk 8]
```diff
@@ -0,0 +1,251 @@
...
+  };
+
+  useEffect(() => {
+    if (isOpen) {
+      document.body.classList.add(styles.dialogVisibleBody);
+      document.addEventListener('keydown', onKeyDown, false);
+    }
+
+    return (): void => {
+      if (isOpen) {
```
VilleMiekkoja, line 212: Is there a possibility that these won't be removed if modal is closed?

minevala: Maybe if the dialog element is not closed gracefully (for example, when removing straight from DOM) 🤔

ronijaakkola:
Documentation has now been updated with examples.

packages/react/src/components/dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,164 @@
+import React, { RefObject, useState } from 'react';
```
VilleMiekkoja, line 1: Very good and comprehensive unit tests! 👍

site/docs/components/dialog.mdx
```diff
@@ -0,0 +1,613 @@
...
+- **Dialogs are very intrusive pattern and they should only be used when the immediate actions or focus from the user are needed.**
+- Be careful when including a separate close ("x") icon in the dialog. If there are more than one action available in the dialog, it can be ambiguous for the user which action is triggered when the close icon is pressed.
+    - Generally it is a good practise to omit the close icon if the dialog has more than one action available.
+- If your dialog contains form elements, follow [HDS form](/patterns/forms) and [form validation patterns](/patterns/form-validation) similarly as you would in a regular forms.
+- As dialogs always contain buttons, pay close attention that the button labels describe the action it is going to trigger. You can read more information about this in the [HDS Button documentation page](/components/button).
+    - Following the same guidelines as the HDS Form pattern, dialog action buttons are placed at the left side of the dialog and the primary action will be the first one from the left.
+- Opening a dialog should be always triggered by the user. Do not open diaglos without user action.
```
VilleMiekkoja, line 34: diaglos -> dialogs

site/docs/components/dialog.mdx
```diff
@@ -0,0 +1,613 @@
...
+| ------------------------------ | ------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------- |
+| [Info](#info-dialog)           | Important information needs to be conveyed to the user. Only requires acknowledgement and no choices from the user. | Informing the user about changed terms of use.
+| [Confirm](#confirm-dialog)     | An action is required from the user.                                                                                | Confirming that the user wants to continue even though all form fields are not filled.
+| [Danger](#danger-dialog)       | An action is required from the user while **the action results are destructive**.                                   | Confirming that the user really wants to delete a blog post.
+
+
+## Accessibility
+- When the dialog is closed, the focus should return to the element that triggered the original dialog context. This allows keyboard and screen readers users to continue the context where they originally opened the dialog.
```
VilleMiekkoja, line 46: screen readers users -> screen reader users?

VilleMiekkoja (approved):
Well done! 🏅

site/docs/components/dialog.mdx
```diff
@@ -0,0 +1,613 @@
...
+    - To emphasize this, a dark screen overlay must be used to cover the view content behind the dialog element.
+- **Dialogs are very intrusive pattern and they should only be used when the immediate actions or focus from the user are needed.**
+- Be careful when including a separate close ("x") icon in the dialog. If there are more than one action available in the dialog, it can be ambiguous for the user which action is triggered when the close icon is pressed.
+    - Generally it is a good practise to omit the close icon if the dialog has more than one action available.
+- If your dialog contains form elements, follow [HDS form](/patterns/forms) and [form validation patterns](/patterns/form-validation) similarly as you would in a regular forms.
+- As dialogs always contain buttons, pay close attention that the button labels describe the action it is going to trigger. You can read more information about this in the [HDS Button documentation page](/components/button).
+    - Following the same guidelines as the HDS Form pattern, dialog action buttons are placed at the left side of the dialog and the primary action will be the first one from the left.
```
laurakarhu, line 33: Should we add note here as well that on danger dialog's primary action is cancel so that the user doesn't accidentally destroy anything? It is mentioned on danger dialog but I guess it wouldn't harm to have it here as well?

laurakarhu (approved):
Good job!

VilleMiekkoja (approved):
Nice 👍

PR #412, opened 2021-04-06
[Fix] Content overlap in the Select component with long labels

pyryk:
## Description

This PR fixes the small layout bug in the `Select` component with labels of a specific length. Before this PR, the label text could overlap with the dropdown caret icon. See screenshots below for more details.

## Related Issue

The issue was briefly discussed in Slack: https://helsinkicity.slack.com/archives/CHCV3KTHA/p1617287264046900

## Motivation and Context

We encountered this bug in Kasko Asti.

## How Has This Been Tested?

This has been tested by hand in Storybook: I verified that the change fixes the bug in question. I also verified that this change does not affect the `Combobox` component or the multi-select variant of the `Select` component.

### Before

Screenshot 2021-04-06 at 9 00 13

### After

Screenshot 2021-04-06 at 9 00 23

### DIFF
--- a/packages/react/src/components/dropdown/select/Select.module.scss
+++ b/packages/react/src/components/dropdown/select/Select.module.scss
@@ -44,6 +44,13 @@
 }
 
 /**
+ * THE BUTTON LABEL (SELECTED VALUE OR PLACEHOLDER)
+ */
+.buttonLabel {
+  padding-right: var(--spacing-layout-xs);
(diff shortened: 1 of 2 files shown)


rjkokko (approved):
LGTM! 👍

VilleMiekkoja (approved):
Looks good, thx 🙏

PR #421, opened 2021-04-15
Use only the simplest dialog story in docs section

minevala:
## Description
Remove all but one dialog story from the docs tab. Three of them are used in Loki tests (open by default). The last one is a too complicated example with two dialogs.

## Motivation and Context
Three of five dialog story examples do not work in Storybooks docs tab.

## How Has This Been Tested?
- Locally 

👉 [Storybook dialog docs](https://city-of-helsinki.github.io/hds-demo/dialog/?path=/docs/components-dialog--default)

### DIFF
--- a/packages/react/src/components/dialog/Dialog.stories.tsx
+++ b/packages/react/src/components/dialog/Dialog.stories.tsx
@@ -81,6 +81,7 @@ export const Default = (args) => {
   );
 };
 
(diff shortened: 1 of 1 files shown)


ronijaakkola (approved):
Looks like this is the best solution for now. At least one story is visible in the Docs tab and as all variants have same props it should be sufficient.

VilleMiekkoja (approved):
Nice 👍

PR #423, opened 2021-04-16
Make doc examples more DRY

minevala:
## Description
- Use constants in dialog documentation examples

## Motivation and Context
- Developers often copy-paste examples, so it is good that the examples themselves follow DRY pattern.
- The constants also express the direct linkage between the element ids and the aria-attributes.

## How Has This Been Tested?
- Locally on dev's machine

### DIFF
--- a/packages/react/src/components/dialog/Dialog.stories.tsx
+++ b/packages/react/src/components/dialog/Dialog.stories.tsx
@@ -15,8 +15,6 @@ export default {
   },
   args: {
     id: 'example-dialog',
-    'aria-labelled-by': 'example-dialog-title',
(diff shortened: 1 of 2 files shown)


ronijaakkola:
@minevala Looks good! I assume these principles have been followed in Storybook code examples as well?

minevala:
> @minevala Looks good! I assume these principles have been followed in Storybook code examples as well?
Yes, I could use variable also there. I'll fix that

laurakarhu (approved):
Great! How could we make this a permanent practice for other new components as well? :)

ronijaakkola (approved):

PR #445, opened 2021-05-06
Fix/datepicker position

minevala:
## Description
- Position DateInput's DatePicker on the bottom right by default. But if there is no space available, it will open on the bottom left. Fixes DatePicker's positioning outside the viewport.
- Fix tests
- Use userEvent instead of fireEvent to simulate user interactions in DateInput tests as [recommended](https://testing-library.com/docs/dom-testing-library/api-events/)
- Add extra weeks if a month has less than six calendar weeks (max number of calendar weeks per month). It ensures the same height for all month-tables.

## Closes
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-779

## Motivation and Context
- In some cases, the DateInput locates on the left of the viewport. And when the DatePicker is open, some of its content is outside the viewport.

## How Has This Been Tested?
- Locally on dev's machine 

Näyttökuva 2021-5-12 kello 17 20 55
Näyttökuva 2021-5-12 kello 17 19 05

👉  [Storybook Demo](https://city-of-helsinki.github.io/hds-demo/datepicker-position/?path=/story/components-dateinput--default)

### DIFF
--- a/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
+++ b/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
@@ -61,6 +61,7 @@
 
 .hds-datepicker__day {
   align-items: center;
+  box-sizing: border-box;
   width: 100%;
   padding: 0 0 100% 0;
   margin: 0;
@@ -71,6 +72,7 @@
   display: inline-flex;
   line-height: 1;
   overflow: hidden;
+  vertical-align: middle;
 
   --background-color: var(--date-background);
   --color: var(--date-color);
--- a/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
+++ b/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
@@ -1,4 +1,5 @@
 @import "../../../../styles/common";
+@value small-up from "../../../../styles/breakpoints.css";
 
 :root {
   --date-background: transparent;
@@ -30,7 +31,7 @@
   min-width: calc(100vw - 2 * var(--spacing-s));
   padding: var(--vertical-spacing) var(--horizontal-spacing);
 
-  @media only screen and (min-width: 576px) {
+  @media small-up {
     min-width: calc(var(--table-width) + 2 * var(--spacing-s));
   }
 }
--- a/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
+++ b/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
@@ -60,6 +60,7 @@
 }
 
 .hds-datepicker__day {
+  align-items: center;
   width: 100%;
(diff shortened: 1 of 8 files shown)


minevala:
> The date picker seems to prefer opening upwards if possible. It also does a weird twitch from down to up when there is even a little amount of space available below the input.

I think I got it fixed now by changing the React-popper's root boundary config to document.

laurakarhu (approved):
Works as expected for me.

ronijaakkola (approved):
Looks good now!

packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
```diff
@@ -27,9 +27,12 @@
   box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.5);
   display: inline-block;
   width: 100%;
-  max-width: calc(var(--table-width) + 2 * var(--horizontal-spacing));
-  min-width: var(--container-width-xs);
+  min-width: calc(100vw - 2 * var(--spacing-s));
   padding: var(--vertical-spacing) var(--horizontal-spacing);
+
+  @media only screen and (min-width: 576px) {
```
VilleMiekkoja, line 33: Wonder if we could/should apply the breakpoint from design-tokens: https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/design-tokens/tokens/breakpoint/breakpoint.json ?

packages/react/src/components/dateInput/components/monthTable/MonthTable.tsx
```diff
@@ -15,10 +17,17 @@ export type MonthTableProps = {
...
+  const start: Date = startOfMonth(month);
+  const end: Date = endOfMonth(month);
+  const monthCalendarWeeks: number = getWeeksInMonth(month, { weekStartsOn: 1 });
+  const maxAmountOfCalendarWeeks = 6;
+  const weeksEnd: Date =
+    monthCalendarWeeks === maxAmountOfCalendarWeeks
+      ? end
+      : add(end, { weeks: maxAmountOfCalendarWeeks - monthCalendarWeeks }); // To ensure equal height month tables, we need to add extra week rows at the end of the table.
+  const weeksInterval: { start: Date; end: Date } = { start, end: weeksEnd };
+  const weeks: Date[] = eachWeekOfInterval(weeksInterval, { locale });
```
VilleMiekkoja, line 30: Wonder if it would be a good idea to extract the process of determining the weeks into its own function? The variables above the line `const weeks: Date[]...` all seem to relate to determining the weeks? Then again, the user would then have to jump more while reading the code, but perhaps the variables above would be more clearly scoped with a meaning 🤔

VilleMiekkoja:
Nicely done! 🎉

VilleMiekkoja (approved):
Nice 👍

PR #453, opened 2021-05-25
Release 1.1.0

minevala:
Release 1.1.0


laurakarhu (approved):
Ready to rock'n'roll!

ronijaakkola (commented):
> **Added**: Icon variations for tags and labels

Was this option added for Tags as well? In designs the custom icon is possible only for StatusLabels. Tags can have the close icon but that has been there since the beginning.

PR #454, opened 2021-05-31
Remove git-hooks

minevala:
## Description
- Remove git hooks

## Motivation and Context
- Making a commit takes about half a minute with all the operations in pre-commit hooks. This change removes git hooks. As a downside, we do not run all the same checks before commits as in the CI pipeline. But we only lose pre-commit type checking, and some IDEs check them automatically without any or with a minimal configuration.

## How Has This Been Tested?
- Locally on dev's machine


laurakarhu (approved):
Better that @VilleMiekkoja still takes a look on these, to me seems solid :)

VilleMiekkoja (approved):
Good job! 🏅

VilleMiekkoja (approved):

VilleMiekkoja (approved):

Issue #467, opened 2021-06-08
Error text alignment is broken


laurakarhu:
In some cases error text alignement breaks and the error text jumps higher than the icon.

laurakarhu:
Released in 1.2.0

PR #469, opened 2021-06-08
Fix/text field error and success text alignments

VilleMiekkoja:
## Description

Fix text field error and success text alignments by specifying explicit line height

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-703
https://github.com/City-of-Helsinki/helsinki-design-system/issues/467

Closes #

https://github.com/City-of-Helsinki/helsinki-design-system/issues/467

## How Has This Been Tested?

- Automated loki tests to ensure no breaking visual changes to the components itself

image

->

image

The issue can be reproduced by setting a line height of 15px to the whole component container. The error text (and success text) will inherit that line height which will break the vertical alignment.

### DIFF
--- a/packages/core/src/components/text-input/text-input.css
+++ b/packages/core/src/components/text-input/text-input.css
@@ -120,6 +120,7 @@
 .hds-text-input__helper-text {
   color: var(--helper-color-default);
   display: block;
+  line-height: var(--lineheight-l);
   font-size: var(--fontsize-body-m);
   margin-top: var(--spacing-3-xs);
 }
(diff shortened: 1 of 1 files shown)


laurakarhu (approved):
Well done!

packages/core/src/components/text-input/text-input.css
```diff
@@ -130,6 +130,7 @@
   color: var(--helper-color-invalid);
   display: block;
   font-size: var(--fontsize-body-m);
+  line-height: var(--lineheight-l);
```
minevala, line 133: Should we add line-height for helper-text too?

minevala (approved):
Good job!

packages/core/src/components/text-input/text-input.css
VilleMiekkoja: Good point. I'll add line-height for it too.

- [ ] Add line-height for helper-text too.

PR #472, opened 2021-06-14
Try to fix lsf-bandwidth issue in development

minevala:
## Description
- Decrease the bandwidth size of Github actions by replacing the file comparison with a file name-hash-based cache key.

## Closes
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-858

## Motivation and Context
- All Github actions are failing with error:  
  `batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data  packs to restore access`
- We create a file name-hash-based cache key without pulling the actual images every time. We can use the ones in Github's actions cache instead
- The solution: https://github.com/actions/checkout/issues/165#issuecomment-657673315
- The origin for solution: https://www.develer.com/en/avoiding-git-lfs-bandiwdth-waste-with-github-and-circleci/

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -11,11 +11,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
+
+      - name: Create LFS file list
+        run: git lfs ls-files -l | cut -d' ' -f1 | sort > .lfs-assets-id
+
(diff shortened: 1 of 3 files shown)


laurakarhu (approved):
I really hope this helps in the future!

VilleMiekkoja (approved):
Looking good 👍

PR #473, opened 2021-06-15
Side navigation component

jorilindell:
## Description
Side navigation component which has following features:
- Side navigation which can have main levels and sub-levels inside main level. Nested main levels are not supported
- Can have multiple ~~or only one~~ main levels open at same time. EDIT: We removed the autoCollapse-feature because the side-effect of clicking an item (closes other accordions) is not apparent for assistive technology. 
- Mobile version of the side navigation which is opened by clicking toggle button
- Supports custom theme 

## Motivation and Context
Linked Events is already using similar SideNavigation component at https://linkedevents.hel.fi/en/help/instructions/general. Motivation for this is to get same component available to other project as well.

## How Has This Been Tested?
Jest unit tests are written to cover at least most use cases

👉 [Storybook Demo](https://city-of-helsinki.github.io/hds-demo/side-navigation/?path=/story/components-sidenavigation--default)

Default side navigation
Näyttökuva 2021-8-23 kello 13 03 41

Mobile version of the side navigation
Näyttökuva 2021-8-23 kello 13 05 11

Side navigation with custom theme
Näyttökuva 2021-8-23 kello 13 03 30


minevala:
# Missing features
- [ ] divider
- [ ] design styles
- [ ] detailed story examples

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
```diff
@@ -0,0 +1,198 @@
...
+        open && styles.open,
+        className,
+      )}
+      style={style}
+    >
+      {hasSubLevels ? (
+        <>
+          <button
+            type="button"
+            aria-current={active ? 'page' : 'false'}
```
minevala, line 166: Accessibility issue: aria-current={true}

packages/react/src/components/sideNavigation/SideNavigation.module.scss
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+.level {
+  & > button {
+    @extend %buttonReset;
+  }
+
+  & > a,
+  & > button {
```
minevala, line 66: Accessibility note: no active style for opened accordion

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
minevala: - [x] use aria-current={boolean} for opened/closed accordion

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
```diff
@@ -0,0 +1,198 @@
...
+  ...rest
+}: MainLevelProps) => {
+  const menuId = `${id}-menu`;
+  const { autoCollapseOthers, openMainLevels, setMobileMenuOpen, setOpenMainLevels } = useContext(
+    SideNavigationContext,
+  );
+  const activeRef = React.useRef<boolean>(active);
+  const open = openMainLevels.includes(index as number);
+
+  const subLevels = React.Children.map(children, (child) => {
```
VilleMiekkoja, line 108: I wonder if this cloning could be avoided by providing context to subLevels? Not sure if it would just require more work? Also it might have bring additional renderings. Is there any risks related to cloning? If user provides a ref inside the sublevel, and we clone it here, then does the original ref still work?

packages/react/src/components/sideNavigation/subLevel/SubLevel.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { useContext, useEffect } from 'react';
+
+import styles from './SubLevel.module.scss';
+import SideNavigationContext from '../SideNavigationContext';
+import classNames from '../../../utils/classNames';
+
+export type SubLevelProps = React.ComponentPropsWithoutRef<'a'> & {
+  /**
+   * If `true`, the item will be marked as active
```
VilleMiekkoja, line 9: What does active mean? How does it differ from open?

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
VilleMiekkoja: Apparently the ref from the original component will be preserved: https://reactjs.org/docs/react-api.html#cloneelement

packages/react/src/components/sideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const container = React.useRef<HTMLDivElement>(null);
+  const customThemeClass = useTheme<SideNavigationCustomTheme>(styles.sideNavigation, theme);
+  const menuId = `${id}-menu`;
+  const [openMainLevels, setOpenMainLevels] = React.useState<number[]>(defaultOpenMainLevels ?? []);
+  const [mobileMenuOpen, setMobileMenuOpen] = React.useState(false);
+  const isMobile = useMobile();
+
+  const mainLevels = React.Children.map(children, (child, index) => {
+    if (isValidElement(child) && (child.type as FCWithName).componentName === 'MainLevel') {
+      return cloneElement(child, { index });
```
VilleMiekkoja, line 85: Why do we need to give this index to mainLevel component? Could we use the id of the component instead?

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
```diff
@@ -0,0 +1,198 @@
...
+   */
+  href?: string;
+  /**
+   * Optional icon of the side navigation main level.
+   */
+  icon?: React.ReactNode;
+  /**
+   * Index of the side navigation main level.
+   */
+  index?: number;
```
VilleMiekkoja, line 34: It seems this is overwritten by the parent component SideNavigation?

packages/react/src/components/sideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,168 @@
...
+          variant="secondary"
+          type="button"
+        >
+          {toggleButtonLabel}
+        </Button>
+        <ul
+          {...{
+            className: classNames(styles.mainLevelList, mobileMenuOpen && styles.open),
+            'aria-label': toggleButtonLabel,
+            'aria-hidden': isMobile && !mobileMenuOpen,
```
VilleMiekkoja, line 156: This might cause problems for server side rendering. As react requires the initially rendered html is the same as the server side rendered. And because in server we do not know if it's mobile, the html will differ for mobile screens.

packages/react/src/components/sideNavigation/mainLevel/MainLevel.module.scss
```diff
@@ -0,0 +1,48 @@
...
+.active {
+  composes: activeLevel from '../SideNavigation.module.scss';
+}
+
+.mainLevelListMenu {
+  display: none;
+}
+
+.mainLevelWithDivider {
+  composes: mainLevelWithDivider from '../SideNavigation.module.scss';
```
VilleMiekkoja, line 21: The mainLevelWithDivider class definition seems to be only applied inside this file? Should we move the definition completely to here in order to decrease code indirection?

packages/react/src/components/sideNavigation/mainLevel/MainLevel.module.scss
```diff
@@ -0,0 +1,48 @@
...
+  width: var(--side-navigation-icon-size);
+}
+
+.rightIcon {
+  margin-left: auto;
+}
+
+.mainLevel.open {
+  .arrowIcon {
+    transform: rotate(180deg);
```
VilleMiekkoja, line 40: Alternative to this might be to just apply another arrow icon in the tsx file. There should be arrow icons for all main directions.

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
minevala: I guess this is needed for that overwrite. But it could be mentioned in the property comment.
- [] add comment of index's purpose

packages/react/src/components/sideNavigation/SideNavigation.tsx
minevala: I think we could, but then the use of property defaultOpenMainLevels might be a bit more complex.
I'll have to test this.

packages/react/src/components/sideNavigation/mainLevel/MainLevel.module.scss
minevala: - [ ] Remove unnecessary CSS indirections

packages/react/src/components/sideNavigation/mainLevel/MainLevel.module.scss
minevala: 👍 
- [ ] Replace CSS rotations with up and down icons

packages/react/src/components/sideNavigation/subLevel/SubLevel.tsx
```diff
@@ -0,0 +1,77 @@
...
+   */
+  href: string;
+  /**
+   * Label of the side navigation sub level.
+   */
+  label: string;
+  /**
+   * Index of the containing main level
+   */
+  mainLevelIndex?: number;
```
VilleMiekkoja, line 31: Should we add a comment that this is generated automatically by the main level component, and user of the SubLevel should not provide this by themself?

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
```diff
@@ -0,0 +1,198 @@
...
+          </button>
+          <ul className={styles.mainLevelListMenu} id={menuId} aria-hidden={!open} aria-labelledby={id}>
+            {subLevels}
+          </ul>
+        </>
+      ) : (
+        <a
+          {...rest}
+          aria-current={active ? 'page' : 'false'}
+          aria-label={external ? `${label} ${openInExternalDomainAriaLabel || 'Siirtyy toiseen sivustoon'}` : label}
```
ronijaakkola, line 184: Examples are in English so this additional label probably should be in English too.

packages/react/src/components/sideNavigation/SideNavigation.tsx
minevala: Good question 👍 
I think it is ok to be the property false as long as it will render the component and won't throw errors. 
I'd guess if we wanted to support fully isMobile property(also in SSR), it should be configurable throughout HDS components.

packages/react/src/components/sideNavigation/subLevel/SubLevel.tsx
minevala: Yes 👍 
- [ ] add clarifying comment of mainLevelIndex

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
ronijaakkola, line 184: When clicked, external main level links also receive the active style. This is often unnoticable since the page changes immediately. However, if this only opens a new tab, it should not receive the active style.

In fact, even if it would not open a new tab, it should not receive the active style. In this case I assume it would look like active link if the user uses the browser's back feature (?).

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+        label={labels.mainLevel3}
+        onClick={handleSubLevelClick}
+      />
+    </SideNavigation>
+  );
+};
+AutoCollapseOtherOpened.args = { autoCollapseOthers: true };
+
+export const CustomTheme = (args) => Default(args);
+CustomTheme.args = {
```
ronijaakkola, line 255: I think the custom theme Storybook example has some contrast issues that we probably do not want to display as an example. I can alter the example so it is more in line with the Hel.fi color palette approach.

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+        label={labels.mainLevel4}
+        external
+        onClick={handleClick(setActive)}
+        withDivider
+      />
+    </SideNavigation>
+  );
+};
+
+export const AutoCollapseOtherOpened = (args) => {
```
ronijaakkola, line 186: Auto collapse feature has a strange functionality where all accordions collapse when a main level link is clicked. Now, this is mostly unnoticable because the page also changes. However, I assume this also happens if the main level link is external. In this case, it would be clearly visible to the user and it looks a bit weird. I think the accordions should not collapse if another main level link (internal or external) is activated.

ronijaakkola (changes requested):

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
minevala: - [ ] fix active styles

minevala: - [ ] use english label for openInExternalDomainAriaLabel

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
minevala: I guess we could remove the whole property?

laurakarhu (commented):
Otherwise looks really good and behaves as expected, but in real life, external link cannot really be with active status either. External link takes user outside the website. If it opens in the same tab, the navigation status is irrelevant. If the link opens in new tab, the original page and selected page should stay active.

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
ronijaakkola: Remove the whole `AutoCollapse` property? I would be fine with it being removed for now. It is practically only useful if the side navigation has a ton of elements (and accordions) in it. Even then, maybe the navigation hierarchy design was not so successful after all.

What does @laurakarhu think about this?

laurakarhu: I agree, auto collapse is not mandatory to have, at least at this stage. And like Roni said, mostly helpful for massive menus.

minevala: - [ ] remove autoCollapse-property

ronijaakkola:
A couple of questions related to the mobile variant.

The button seems to have a maximum width which makes it look odd on certain screen sizes (just on the edge of the next breakpoint). Can a dev make this full-width if wanted so?
image

It was specified that when a mobile side nav is opened, it **pushes** all of the content below it. So it does not open a top of content. I cannot verify this from the Storybook examples.

packages/react/src/components/sideNavigation/SideNavigation.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should not have basic accessibility issues', async () => {
+    const { container } = renderSideNavigation();
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
+  test('should open main level if sub level is active', () => {
```
VilleMiekkoja, line 51: Previous tests apply the _it_ instead of _test_. Im not too familiar with their differences, is there a reason why we wish to use _test_ here and onwards?

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
VilleMiekkoja: Finnish is probably the most common and default language in Helsinki services though. It might be better to let the default aria-label to be in Finnish language, and change the examples to be in Finnish too. 🤔

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+  mainLevel2: 'Main level accordion',
+  mainLevel3: 'Main level link',
+  mainLevel4: 'Main level external link',
+  subLevel1: 'Sub level link',
+  subLevel2: 'Sub level link',
+  subLevel3: 'Sub level link',
+  subLevel4: 'Sub level link',
+};
+
+const handleClick = (setActive) => (ev) => {
```
VilleMiekkoja, line 44: If the handleClick is defined outside of a story, it will not be shown in storybook under the **Story** tab. Not sure if showing this would bring that much additional value though. Also it would have to repeated every time to show it in storybook, not sure if it's worth it.

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+        external
+        openInExternalDomainAriaLabel="Opens a different website"
+        onClick={handleClick(setActive)}
+        withDivider
+      />
+    </SideNavigation>
+  );
+};
+
+export const WithIcons = (args) => {
```
VilleMiekkoja, line 107: Should we add one story where the side navigation is used with the normal navigation? It would be cool to see that they fit and work together without problems. Or maybe we will use this combination in the new docs site?

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+<LargeParagraph>
+  A side navigation is an additional navigation element which is located at the left-hand side of the viewport. It provides an extra level of navigation and also can be helpful in situations where the traditional top navigation feels too clumsy.
+</LargeParagraph>
+
+## Principles
+
+- **It is strongly recommended to always include the [main (top) navigation component](/components/navigation) in your service.**
+  - It is allowed to use the side navigation without the top navigation. However, please note that the top navigation includes important features such as language selection and login. If you rely on the side navigation only, you need to provide these functionalities somewhere else in the user interface.
+- If you use both main (top) and side navigation, do not repeat same navigation items in both of them.
```
VilleMiekkoja, line 26: do not repeat same navigation -> do not repeat **the** same navigation

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+- Always position the side navigation to the left-hand side of the page. Keep the navigation in place so the user can always easily locate it.
+- Keep navigation menu link labels clear and concise. Prefer max 1-2 word long labels. Avoid starting multiple menu link labels with the same word.
+- The side navigation allows the use of icons for the first level links.
```
VilleMiekkoja, line 29: first level -> first-level

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+- **Do use the Side navigation when**
+  - the service has more than three (3) levels in its navigation hierarchy
+  - the user needs to be able to easily scan through and navigate between pages of a specific level (e.g. admin interfaces or documentation sites)
+  - you are uncertain of the number of the navigation items and you need room for growth and changes
+- **Don't use the Side navigation when**
+  - you need space for your content (the side navigation uses a large portion of the content area)
+  - your service has a low amount of navigation items and only one level deep navigation hierarchy
+
+## Accessibility
+- **When designing a menu link hierarchy, always think about keyboard and screen reader users.** First, make sure top-level menu labels are clear, and submenu items contextually fit under it. Next, make sure the menu option order is logical and reasonable. Remember that keyboard users need to manually navigate to each element and thus placing a commonly used menu option last is not optimal.
```
VilleMiekkoja, line 47: and submenu items contextually fit under it -> and submenu items contextually fit under **the top-level**.

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+- **It is strongly recommended to always include the [main (top) navigation component](/components/navigation) in your service.**
+  - It is allowed to use the side navigation without the top navigation. However, please note that the top navigation includes important features such as language selection and login. If you rely on the side navigation only, you need to provide these functionalities somewhere else in the user interface.
+- If you use both main (top) and side navigation, do not repeat same navigation items in both of them.
+- Always position the side navigation to the left-hand side of the page. Keep the navigation in place so the user can always easily locate it.
+- Keep navigation menu link labels clear and concise. Prefer max 1-2 word long labels. Avoid starting multiple menu link labels with the same word.
+- The side navigation allows the use of icons for the first level links.
+  - If icons are used, all first level links should have an icon.
+  - Pay attention to the icon choices. The icons should be easily recognisable and not used for decorative purposes only.
+  - The second level links and external links must not include an icon.
```
VilleMiekkoja, line 32: second level -> second-level

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+## Accessibility
+- **When designing a menu link hierarchy, always think about keyboard and screen reader users.** First, make sure top-level menu labels are clear, and submenu items contextually fit under it. Next, make sure the menu option order is logical and reasonable. Remember that keyboard users need to manually navigate to each element and thus placing a commonly used menu option last is not optimal.
+- If you are using both the main (top) and the side navigation, it is a good practice to include an anchor link to the side navigation at the beginning of the main (top) navigation. This way assistive technologies can easily access the side navigation after a page load.
+
+## Usage and variations
+
+### Default
+
+The default side navigation has a support for two (2) levels of navigation, icons, external links and a divider to create logical groups for the navigation items.
```
VilleMiekkoja, line 54: a support -> support

VilleMiekkoja, line 54: external links and a divider.. -> external links**,** and a divider...

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+</Playground>
+
+#### React code example:
+```tsx
+
+```
+
+### Mobile
+
+In mobile screen sizes the side navigation turns into an expandable menu button.
```
VilleMiekkoja, line 67: In mobile screen sizes -> in mobile screen sizes, ...

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+</Playground>
+
+#### React code example:
+```tsx
+
+```
+
+### With icons
+
+Icons are allowed to be used with the main level navigation items and accordions. The second level navigation items do not have icon support. Also, icons are not visible in the mobile side navigation variant.
```
VilleMiekkoja, line 80: second level -> second-level

site/docs/components/side_navigation.mdx
```diff
@@ -0,0 +1,101 @@
...
+import LargeParagraph from "../../src/components/LargeParagraph";
+import Link from "../../src/components/Link";
+
+# Side navigation
+
+<StatusLabel type="alert">Draft</StatusLabel>
+<StatusLabel type="success" style={{marginLeft: 'var(--spacing-xs)'}}>Accessible</StatusLabel>
+
+<LargeParagraph>
+  A side navigation is an additional navigation element which is located at the left-hand side of the viewport. It provides an extra level of navigation and also can be helpful in situations where the traditional top navigation feels too clumsy.
```
VilleMiekkoja, line 19: which is -> that is

packages/react/src/components/sideNavigation/SideNavigation.stories.tsx
laurakarhu: Yes, we should have an example of using both, but like Ville said, new docsite should have both navigations and also there is a separate ticket https://helsinkisolutionoffice.atlassian.net/browse/HDS-917 to create pattern documentation for the navigation: what levels to present in which navigation, how to combine those and so on. So I'd say way will put a little more effort to examples of side navigation together with main navigation later on.

packages/react/src/components/sideNavigation/mainLevel/MainLevel.tsx
minevala: I fixed it by using English in the story. The default label is still in Finnish.

packages/react/src/components/sideNavigation/SideNavigation.test.tsx
minevala: Probably copy-paste thingy:
- [ ] use consistent test commands throughout tests

minevala:
> A couple of questions related to the mobile variant.
> 
> The button seems to have a maximum width which makes it look odd on certain screen sizes (just on the edge of the next breakpoint). Can a dev make this full-width if wanted so?
> image
> 
> It was specified that when a mobile side nav is opened, it **pushes** all of the content below it. So it does not open a top of content. I cannot verify this from the Storybook examples.

- [ ] Make mobile "Navigate to page" - button full-width
- [ ] Check mobile navigation pushes content down

packages/react/src/components/sideNavigation/SideNavigation.tsx
minevala: I can test this component in SSR context.

minevala: Seems to work in hds-next:
Näyttökuva 2021-8-24 kello 10 40 13

laurakarhu (approved):
Thanks!

packages/react/src/components/sideNavigation/SideNavigation.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        </SideNavigation.MainLevel>
+      </SideNavigation>,
+    );
+
+    expect(queryLink('subLevel1')).not.toBeInTheDocument();
+    userEvent.click(queryButton('mainLevel1'));
+    expect(queryLink('subLevel1')).toBeInTheDocument();
+  });
+
+  it('should open main level if sub level is active', () => {
```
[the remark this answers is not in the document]
minevala: It can be set programmatically, so we have to ensure its mainlevel is open.

VilleMiekkoja (approved):
Good work! 🎉

PR #486, opened 2021-06-28
Fix/inline-notification-z-index

VilleMiekkoja:
## Description

Removes z index (sets to default) from the inline version of the notification component. Having a z index for the inline version caused problems with overlapping layouts like navigation menus.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-763

## How Has This Been Tested?

- Local testing on developers machine


laurakarhu (approved):
Thanks!

minevala (approved):
Nice 👍

PR #498, opened 2021-08-04
Add "upcoming breaking changes"-page

minevala:
## Description
- Add new "Upcoming breaking changes" - page

## Motivation and Context
- HDS library users can see planned breaking changes from the documentation site too (in addition to Github)

## How Has This Been Tested?
- Locally on dev's machine

Näyttökuva 2021-8-5 kello 12 13 07

### DIFF
--- a/site/docs/about/breaking_changes.mdx
+++ b/site/docs/about/breaking_changes.mdx
@@ -10,7 +10,7 @@ import Link from "../../src/components/Link";
 # Upcoming breaking changes
 
 <LargeParagraph>
-  Here you will find summarized notes of breaking changes of HDS. The source can be found here, <Link href="https://github.com/City-of-Helsinki/helsinki-design-system/issues/495" external>Github breaking changes issue</Link>.
+  Here you will find summarized notes of upcoming breaking changes of HDS. The source can be found here, <Link href="https://github.com/City-of-Helsinki/helsinki-design-system/issues/495" external>Github upcoming breaking changes issue</Link>.
 </LargeParagraph>
 
 <BreakingChangesComments />
--- a/site/docs/about/breaking_changes.mdx
+++ b/site/docs/about/breaking_changes.mdx
@@ -1,0 +1,16 @@
+---
+name: Upcoming breaking changes
+route: /about/breaking
+menu: About
+---
+import BreakingChangesComments from '../../src/components/BreakingChangesComments';
+import LargeParagraph from "../../src/components/LargeParagraph";
+import Link from "../../src/components/Link";
+
+# Upcoming breaking changes
+
+<LargeParagraph>
+  Here you will find summarized notes of breaking changes of HDS. The source can be found here, <Link href="https://github.com/City-of-Helsinki/helsinki-design-system/issues/495" external>Github breaking changes issue</Link>.
+</LargeParagraph>
+
+<BreakingChangesComments />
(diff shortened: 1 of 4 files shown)


site/docs/about/breaking_changes.mdx
```diff
@@ -0,0 +1,16 @@
...
+menu: About
+---
+import BreakingChangesComments from '../../src/components/BreakingChangesComments';
+import LargeParagraph from "../../src/components/LargeParagraph";
+import Link from "../../src/components/Link";
+
+# Upcoming breaking changes
+
+<LargeParagraph>
+  Here you will find summarized notes of breaking changes of HDS. The source can be found here, <Link href="https://github.com/City-of-Helsinki/helsinki-design-system/issues/495" external>Github breaking changes issue</Link>.
```
laurakarhu, line 13: Tässä käyttäisin sitä "upcoming breaking changes", tarkoituksena on resolvata noita sitten kun muutokset on julkaistu.

laurakarhu (approved):
Thanks!

site/src/components/BreakingChangesComments.js
```diff
@@ -0,0 +1,38 @@
...
+
+const BreakingChangesComments = () => {
+  const [comments, setComments] = useState([]);
+
+  useEffect(() => {
+    fetch('https://api.github.com/repos/City-of-Helsinki/helsinki-design-system/issues/495/comments')
+      .then((response) => response.json())
+      .then((resultData) => {
+        setComments(resultData);
+      }); // set data for the number of stars
```
VilleMiekkoja, line 19: Number of stars?

site/src/components/BreakingChangesComments.js
```diff
@@ -0,0 +1,38 @@
...
+  <p style={{ fontStyle: 'italic' }}>
+    {label}: <time dateTime={date}>{format(parseJSON(date), 'd.M.yyyy')}</time>
+  </p>
+);
+
+const BreakingChangesComments = () => {
+  const [comments, setComments] = useState([]);
+
+  useEffect(() => {
+    fetch('https://api.github.com/repos/City-of-Helsinki/helsinki-design-system/issues/495/comments')
```
VilleMiekkoja, line 15: Is the goal to add upcoming breaking changes to that github issue, and then fetch them from there to our doc site? I fear this adds one more manual step to release process; after release (or before release), one must check the github issue and update it. Though we probably won't have breaking changes that often, so probably not going to be a big burden

VilleMiekkoja: Also Im not sure if it's a good idea to promise upcoming breaking changes. What if we end up not making those breaking changes after more close inspection? Hmm..

laurakarhu: Good discussion and relevant pondering on the topic :) Yes the goal is that to have one place to update the information about upcoming breaking changes. This brings the same information available on the docsite e.g. for POs and designers who are not that familiar with Github, which is more of a developer tool.

We have been requested to communicate breaking changes beforehand so that the projects will have the possibility to prepare mentally for those. 

Also, communications always require some manual work, this makes it happen in one place only. 

I wouldn't be too worried about not making the breaking changes in the end. It is entirely up to us to decide what and when we say something is going to have breaking changes. The current notes include navigation (which has an epic of 42 fixes waiting to be done) and typography update (which will affect every single component) so I would be surprised if nothing breaks and that would be a very positive surprise.

site/src/components/BreakingChangesComments.js
minevala: Will remove 😁

VilleMiekkoja (approved):
Good job 🎖️ 👍

site/src/components/BreakingChangesComments.js
minevala: - [x] comment removed

PR #500, opened 2021-08-05
Feature | Support clearing DateInput value programmatically

Nizar-Rahme:
## Description

- Fix the condition of updating the `DateInput` value to allow setting it as an empty string by:
  - Removing `if (providedValue)` to allow **all** passed values even empty strings
  - Adding `if (didMount)` to allow `defaultValue` to be set for uncontrolled components
- Write a [Story for the intended use case](http://localhost:6006/?path=/story/components-dateinput--with-external-clear-value-button)
- Remove a redundant wrapper in the `PasswordInput` story

## Related Issue

Closes #

## Motivation and Context

We are using the `DateInput` component in `YJDH - Helsinki-lisä` project, but we faced an issue where you cannot clear the input value programmatically (setting it to an empty string `''`).

## How Has This Been Tested?
This can be tested in the Storybook


packages/react/src/components/dateInput/DateInput.tsx
```diff
@@ -89,7 +90,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
      * Set the input value if value prop changes
      */
     useEffect(() => {
-      if (providedValue) {
+      if (didMount) {
         setInputValue(providedValue);
       }
     }, [providedValue]);
```
minevala, line 96: The prevent "exhaustive-deps"-warning, we need to provide didMount in the dependency array.

minevala (commented):
Good job 👍
I guess we could add one component test for clearing the value

minevala:
I also noticed it still shows the previous month in the datepicker after clearing the value (
when no value is provided, it shows the current month)

minevala:
I also noticed that defaultValue is not passed into textInput.

Nizar-Rahme:
> Good job 👍
> I guess we could add one component test for clearing the value/replacing the value with empty string etc.

Added a test

packages/react/src/components/dateInput/DateInput.tsx
Nizar-Rahme: Adding `didMount`, as it was, to the array caused another issue; however, I changed the value of `didMount`, and now the warnings are gone.

Nizar-Rahme:
> I also noticed it still shows the previous value month in the datepicker after clearing the value
> (when no value is provided, it should show ongoing month)
> EDIT: it seems that datepicker's current month is not updated when the value is cleared.

Fixed

Nizar-Rahme:
> I also noticed that defaultValue is not passed into textInput.

Yes, it is not. I tried to add it but got a fair warning about changing the type of the input element that reads:
> Warning: ForwardRef contains an input of type text with both value and defaultValue props. Input elements must be either controlled or uncontrolled

Which got me thinking, that with the current implementation `defaultValue` is redundant, and basically the `DateInput` is always a controlled input. But deleting the prop `defaultValue` (or supporting uncontrolled input types properly) is a breaking change and it might be beyond the scope of this PR.

packages/react/src/components/dateInput/DateInput.test.tsx
```diff
@@ -296,4 +296,18 @@ describe('<DateInput /> spec', () => {
     // No date should be selected in the calendar
     expect(container.querySelector('[aria-pressed="true"]')).toBeNull();
   });
+
+  it('should be able to clear the value', async () => {
+    const { rerender } = render(<DateInput id="date" label="Foo" value="10.02.2022" />);
+
+    // The initial value should be there
+    expect(screen.getByRole('textbox')).toHaveValue('10.02.2022');
```
minevala, line 304: The test could also check that DatePicker shows the value's month:
```js
expect(screen.getByLabelText('Month')).toHaveValue('1');
expect(screen.getByLabelText('Year')).toHaveValue('2022');
```
  
and then shows today's (mocked today is 17.1.2021 in this testset) month:
```js
expect(screen.getByLabelText('Month')).toHaveValue('0');
expect(screen.getByLabelText('Year')).toHaveValue('2021');
```

packages/react/src/components/dateInput/DateInput.tsx
minevala: I ended up testing something similar 😁

minevala (commented):
Nice improvements 👍 
I think it works now as expected. 
I added a tiny test comment

minevala:
> Which got me thinking, that with the current implementation, defaultValue is redundant (or not used as it should be), and basically the DateInput is always a controlled input. But deleting the prop defaultValue (or supporting uncontrolled input types properly) is a breaking change and it might be beyond the scope of this PR

Yes, it is a breaking change, and it is outside of the scope of the PR.
I'll add a separate a ticket of that to our backlog.

packages/react/src/components/dateInput/DateInput.test.tsx
Nizar-Rahme: Added

VilleMiekkoja (approved):
Well done 🏅

minevala (approved):
Nice job 👍

PR #507, opened 2021-08-10
[bug-fixing] Add lang-attributes to Navigation.LanguageSelector examples

minevala:
## Description
- Add lang-attributes to Navigation.LanguageSelector examples
- Log warn if LanguageSelector option is missing a lang-attribute.

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-753

## Motivation and Context
- The fix helps assistive technology to read the language options

## How Has This Been Tested?
- Locally on dev's machine

### DIFF
--- a/packages/react/src/components/navigation/navigationLanguageSelector/NavigationLanguageSelector.tsx
+++ b/packages/react/src/components/navigation/navigationLanguageSelector/NavigationLanguageSelector.tsx
@@ -7,6 +7,7 @@ export const NavigationLanguageSelector = ({ children, id = 'languageSelector', 
   React.useEffect(() => {
     React.Children.forEach(children, (child:React.ReactElement) => {
       if (!child.props.lang) {
+        // eslint-disable-next-line no-console
         console.warn(`NavigationLanguageSelector item "${child.props.label}" is missing a lang property.`);
       }
     });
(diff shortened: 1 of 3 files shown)


laurakarhu (approved):
:)

packages/react/src/components/navigation/navigationLanguageSelector/NavigationLanguageSelector.tsx
```diff
@@ -3,9 +3,20 @@ import React from 'react';
...
-  <MenuButton className={styles.languageDropdown} id={id} label={label} menuOffset={10} {...rest} closeOnItemClick>
-    {children}
-  </MenuButton>
-);
+export const NavigationLanguageSelector = ({ children, id = 'languageSelector', label, ...rest }: MenuButtonProps) => {
+  React.useEffect(() => {
+    React.Children.forEach(children, (child: React.ReactElement) => {
+      if (!child.props.lang) {
+        // eslint-disable-next-line no-console
+        console.warn(`NavigationLanguageSelector item "${child.props.label}" is missing a lang property.`);
```
VilleMiekkoja, line 11: Should we add the empty array in the dependencies of the useEffect, so that this warning won't be printed many times? Without the empty array useEffect is run after each render I think?

minevala: Good catch 👍 
I will add an empty array

minevala: I will probably need add children as dependency but I think it is ok

VilleMiekkoja (approved):
Good job! 🎉

PR #511, opened 2021-08-10
[bug fixing] Pass native event to onTitleClick

VilleMiekkoja:
## Description

Pass native event to onTitleClick. This is to prevent page reload which happens because of anchor tag

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-919

### DIFF
--- a/packages/react/src/components/navigation/Navigation.tsx
+++ b/packages/react/src/components/navigation/Navigation.tsx
@@ -54,7 +54,7 @@ export type NavigationProps = React.PropsWithChildren<{
   /**
    * Callback fired when the title or logo is clicked
    */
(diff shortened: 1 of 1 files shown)


minevala (commented):
The onTitleClick callback type is still `onTitleClick?: () => void;` which gives an error when trying to use the event argument in the callback function.

VilleMiekkoja:
> The onTitleClick callback type is still `onTitleClick?: () => void;` which gives an error when trying to use the event argument in the callback function.

This is fixed now.

minevala (approved):
Nice 👍

laurakarhu (approved):

PR #521, opened 2021-08-18
Feature/fileinput

minevala:
## Description
- Add FileInput component

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-702

## Depends on:
- https://github.com/City-of-Helsinki/helsinki-design-system/pull/522

## How Has This Been Tested?
- Manually on developers machine
- Automated unit tests
- Automated visual regression loki tests

[👉 Storybook Demo](https://city-of-helsinki.github.io/hds-demo/file-input/?path=/story/components-fileinput--single)

Näyttökuva 2021-9-17 kello 12 45 38
Näyttökuva 2021-9-17 kello 12 46 29
Näyttökuva 2021-9-17 kello 12 46 51


VilleMiekkoja:
This loki reference file seemed to be cut off, not sure if we could do something about it: chrome_iphone7_Components_FileInput_With_Drag_And_Drop.png

packages/react/src/components/fileInput/FileInput.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+        removeButtonAriaLabel={(name) => `Remove ${name} from the list`}
+        removeSuccessMessage="File removed."
+        onChange={onChangeTest}
+      />,
+    );
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
+  test('should list added files', async () => {
```
VilleMiekkoja, line 46: test -> it

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+  const onDragLeave = (event: React.DragEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    event.stopPropagation();
+    setIsDragOverDrop(false);
+  };
+
+  useEffect(() => {
+    if (onChange) {
+      onChange(selectedFiles);
```
VilleMiekkoja, line 221: I guess this means that the onChange is also run after the component is mounted, before the user has selected any files? Just wondering if this behaviour is expected. What if user only expects the onChange to be run when there's actual change in the input?

minevala: The onChange is also called when a file is removed from the list or dragged into the drag-area. 
The selectedFiles value might differ from the value of the actual input.
But it might not be necessary to run this effect when the component has mounted.

minevala:
> This loki reference file seemed to be cut off, not sure if we could do something about it: chrome_iphone7_Components_FileInput_With_Drag_And_Drop.png

minevala:
> This loki reference file seemed to be cut off, not sure if we could do something about it: chrome_iphone7_Components_FileInput_With_Drag_And_Drop.png

I tweaked the drag-and-drop area styles so the area should fit on the screen.

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: Good point! And the users of the onChange method should always check if there are any selected files. Though it might be little unexpected that the onChange will be called on first render even there's no change.

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,365 @@
...
+      inputRef.current.value = '';
+    }
+  };
+
+  const handleSingleFileChange = (files: File[]) => {
+    if (files.length > 0) {
+      setSelectedFiles(files);
+      setSuccessText(getAddSuccessMessage(language));
+    } else {
+      setSuccessText(undefined);
```
VilleMiekkoja, line 196: Should there be `setSelectedFiles([])` here? If user removes the file, would we want to set the selectedFiles empty, so that the onChange callback fires?

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] prevent initial onChange

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,365 @@
...
+   */
+  label: string;
+  /**
+   * The label for the file button. The button is not visible for assistive technology
+   */
+  buttonLabel: string;
+  /**
+   * The language of the component. It affects which language is used to present component-specific messages, labels, and aria-labels
+   *
+   * @default "en"
```
VilleMiekkoja, line 45: Should the default language be fi instead of en? 🤔

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,365 @@
...
+                  <IconUpload aria-hidden />
+                  <span className={styles.dragAndDropLabelText}>{dragAndDrop.label}</span>
+                </div>
+              </div>
+              <div className={styles.dragAndDropHelperText}>{dragAndDrop.helperText}</div>
+            </>
+          )}
+          <div className={styles.fileInputWrapper}>
+            <Button
+              aria-hidden
```
VilleMiekkoja, line 301: Is this aria-hidden necessary?

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,365 @@
...
+    helperText,
+    successText,
+    errorText,
+    id,
+    label,
+    required,
+    style,
+  };
+
+  const passButtonClickToInput = () => {
```
VilleMiekkoja, line 179: In single-file mode, should we warn the user that he is overwriting the previous file that he sent? That would probably require some kind of modal which might be tedious to implement here though 🤔

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] set "fi" as default lang

packages/react/src/components/fileInput/FileInput.tsx
minevala: When the user clicks "cancel" in the file window, it triggers onChange with an empty list. I'm not sure if we want to remove the current file then 🤔

packages/react/src/components/fileInput/FileInput.tsx
minevala: Yes, two inputs side-by-side complicate things with assistive technology.

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+   */
+  id: string;
+  /**
+   * The label for the input
+   */
+  label: string;
+  /**
+   * The label for the file button. The button is not visible for assistive technology
+   */
+  buttonLabel: string;
```
VilleMiekkoja, line 43: In some components we have these props listed in alphabetical order, while in others they're not ordered like that. If I would have to guess based on few samples that I checked, I would say it's more common to have them in alphabetical order. So maybe we should order them here as well?

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+  }[language];
+};
+
+const getAddSuccessMessage = (language: Language, numberOfAdded: number, numberOfTotal: number): string => {
+  const partOfTotalStr = numberOfAdded === numberOfTotal ? numberOfAdded : `${numberOfAdded}/${numberOfTotal}`;
+
+  return {
+    en: `${partOfTotalStr} file(s) added.`,
+    fi: `${partOfTotalStr} tiedosto(a) lisätty.`,
+    sv: `${partOfTotalStr} file(s) added.`,
```
VilleMiekkoja, line 163: Seems to be missing the swedish translation

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+                  const { dataTransfer }: { dataTransfer: DataTransfer } = event;
+                  onDragLeave(event);
+                  onFilesChange(Array.from(dataTransfer.files));
+                }}
+                className={classNames(styles.dragAndDrop, isDragOverDrop && styles.dragAndDropActive)}
+                ref={dropAreaRef}
+              >
+                <div className={styles.dragAndDropLabel}>
+                  <IconUpload aria-hidden />
+                  <span className={styles.dragAndDropLabelText}>{dragAndDrop.label}</span>
```
VilleMiekkoja, line 480: Noticed that if you click this drag and drop label, then it will open the choose file dialog. I wonder if that is what we want? It could be nice, but then we would probably want it to be open also if user click the drag and drop area?

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+  const ariaDescribedBy: string = [
+    composeAriaDescribedBy(id, helperTextToUse, errorTextToUse, successText, infoTextToUse),
+    hasFilesSelected && fileListId,
+  ]
+    .filter((text) => !!text)
+    .join(' ');
+
+  return (
+    <>
+      <InputWrapper {...wrapperProps}>
```
VilleMiekkoja, line 461: Is the `InputWrapper` necessary here? There's an input, but it's hidden from the screen. If I recall right, the InputWrapper will bring all sorts of styles related to text input? Though now when we don't have a text input here, those styles should not match? Could it complicate or interfere with styles of the component otherwise? It might bring some unnecessary styles that might have unknown side effects hmm.. Though it does bring the needed labels and info texts. I guess if we want to reuse those this is the way to go.

packages/react/src/components/fileInput/FileInput.stories.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [file, setFile] = React.useState<File[]>();
+  // eslint-disable-next-line no-console
+  console.log('selected file', file);
+
+  return (
+    <FileInput
+      {...args}
+      id="file-input"
+      label="Choose a file"
+      buttonLabel="Add file"
```
VilleMiekkoja, line 26: Not sure if this would grammatically require an article 'a'? Though not a big deal anyways

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+            tabIndex={-1}
+            ref={(el) => {
+              if (el && fileListRef.current && fileListFocusIndexRef.current === index) {
+                el.focus();
+              }
+            }}
+          >
+            {file.type.startsWith('image') ? <IconPhoto aria-hidden /> : <IconDocument aria-hidden />}
+            <div className={styles.fileListItemTitle}>
+              <span className={styles.fileListItemName}>{file.name}</span>
```
VilleMiekkoja, line 535: Noticed the font-size here had a small difference from design: 16px -> 14px.

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+            key={file.name}
+            className={styles.fileListItem}
+            tabIndex={-1}
+            ref={(el) => {
+              if (el && fileListRef.current && fileListFocusIndexRef.current === index) {
+                el.focus();
+              }
+            }}
+          >
+            {file.type.startsWith('image') ? <IconPhoto aria-hidden /> : <IconDocument aria-hidden />}
```
VilleMiekkoja, line 533: Not sure if it's easy/worth to do, but it might look better in some cases if the icon, file size and remove button would be vertically centered. If the filename is long it can span to multiple lines like this:

image

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+            <Button
+              aria-hidden
+              variant="secondary"
+              iconLeft={<IconPlus aria-hidden />}
+              onClick={() => passButtonClickToInput()}
+              disabled={disabled}
+            >
+              {buttonLabel}
+            </Button>
+            <input
```
VilleMiekkoja, line 496: We might want to set `tabIndex=-1` to this input, so that keyboard users cannot focus it. Screenreaders should still be able to reach it.

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: We should probably change it to the comment too. I saw you changed it already to code, nice 👍

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+
+const findDuplicateByNameAndType = (files: File[], fileToCompare: File): File | undefined =>
+  files.find((file: File) => isEqualFileBy(['name', 'type'], file, fileToCompare));
+
+type DragAndDropProps = {
+  label: string;
+  helperText: string;
+};
+
+type Language = 'en' | 'fi' | 'sv';
```
VilleMiekkoja, line 29: This seems to be a common pattern in other HDS components too, that we support 3 languages. We discussed before that it does have the bad side-effect of limiting the usage only to services that has these languages. I wonder if we should not limit to 3 languages, but let users give those messages through props? Though there's quite many texts there that the user should then give. When the component gives them based on the language that has been selected, it automates things a lot. I wonder if we should still enable users overwriting those texts some way 🤔

packages/react/src/components/fileInput/FileInput.tsx
```diff
@@ -0,0 +1,554 @@
...
+   * If `true`, the label is displayed as required and the `input` element will be required
+   */
+  required?: boolean;
+  /**
+   * Override or extend the styles applied to the component
+   */
+  style?: React.CSSProperties;
+};
+
+const formatBytes = (bytes: number): string => {
```
VilleMiekkoja, line 104: There might be value for making separate unit tests for this format function? There seems to be some arithmetic operations and logic inside this that could be tested separately.

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: Oh good point! We would somehow need to know if it was an actual removal, or just cancelled operation, in order to trigger onChange correctly

packages/react/src/components/fileInput/FileInput.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+        buttonLabel="Add file"
+        language="en"
+        accept=".png,.jpg"
+        onChange={onChangeTest}
+      />,
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should not have basic accessibility issues', async () => {
```
VilleMiekkoja, line 27: Would it be good idea to also test the accessibility in a state where there are multiple files added?

VilleMiekkoja:
Good job with this one! I added couple minor comments here and there, but looking good 👍

packages/react/src/components/fileInput/FileInput.tsx
minevala: Yes, true. We have to remember to fix these 👍

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] order props

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] add drag-and-drop area click handler

packages/react/src/components/fileInput/FileInput.tsx
minevala: I have to think about this 🤔  It would mean duplicate code

packages/react/src/components/fileInput/FileInput.stories.tsx
minevala: - [ ] Fix typo

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] check correct font size

packages/react/src/components/fileInput/FileInput.tsx
minevala: They should be "vertically centered", but I have to check this 🤔

minevala: - [ ] check the alignment of icon, size text and remove button

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] set tabIndex=-1 for input to prevent unnecessary focus

packages/react/src/components/fileInput/FileInput.tsx
minevala: I would restrict the languages because this is not the only component with internal texts. 
Another option is to make all ten labels configurable. I tested the API of component, and it starts to be a bit annoying to use 🤔

packages/react/src/components/fileInput/FileInput.tsx
minevala: Then it should be a separate utility function?

packages/react/src/components/fileInput/FileInput.test.tsx
minevala: It could. I need to figure out how I'm going to do it 👍

packages/react/src/components/fileInput/FileInput.tsx
minevala: But on the other hand, it could make texts more apparent for the library user 🤔

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] in documentation too

packages/react/src/components/fileInput/FileInput.test.tsx
minevala: - [ ] add accessibility test for state where files are selected files

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: Could we export it separately from this file, so that it could be tested?

packages/react/src/components/fileInput/FileInput.tsx
minevala: Should we have a message which tells when file or files are replaced?

packages/react/src/components/fileInput/FileInput.tsx
minevala: It might be that I misunderstood but they are already aligned 🤔

packages/react/src/components/fileInput/FileInput.tsx
minevala: I need to check this from SiteImprov since we are focusing on the input when files are removed 🤔

packages/react/src/components/fileInput/FileInput.tsx
minevala: Yes, it might work but where would I put the tests?

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: Hmm.. that might be useful. Normally there's info that says 'Files added'. Maybe it could say: 'File replaced'? 🤔

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: Maybe it could be tested in the same file where the FileInput component is tested? Like inside a different describe block

packages/react/src/components/fileInput/FileInput.tsx
VilleMiekkoja: Yeah I was thinking something like this:
131997229-31d0bb7b-ae52-4a4d-b82f-0c8a2dc17d4b

->

Vertically aligned

VilleMiekkoja: Though it might even look better without the vertical alignment. Specially if the filename spans to more lines than 2, it can look weird if the icon is centered

minevala: Yes, it was originally implementation but it got fixed when the names spanned to multiple lines

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] add test for formatBytes

ronijaakkola:
There seems to be a slight difference in the alignment of the uploaded file name+size+remove button div. The elements are not fully vertically centered on the container.

Implementation:
image

Design files:
image

minevala:
> There seems to be a slight difference in the alignment of the uploaded file name+size div. The elements are not fully vertically centered on the container.
> 
> Also, a slight difference can be seen on the dotted line (not a big deal though).
> 
> Implementation:
> image
> 
> Design files:
> image

The spacings should be easy to fix but the border-line might be trickier. There is only one kind of dotted line in CSS.

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] fix alignment

site/docs/components/file_input.mdx
```diff
@@ -0,0 +1,121 @@
...
+  - A FileInput allowing multiple files fits well for situations where all the files are related to each other (e.g. multiple photos from the same event).
+  - If the user has to upload multiple files that are not related to each other, prefer using multiple file uploads - one for each file. 
+- Dragging and dropping files can be allowed by using the `dragAndDrop` property.
+
+## Accessibility
+- While dragging and dropping files is not considered to be accessible, HDS will always include the default input below the drag and drop area. Clicking the drag and drop area also opens the device-specific file browser.
+- For selecting files, HDS utilises the file browser native to the user's device. While all devices' file browsers do not have the best accessibility features, they are often the most familiar way of selecting files for the user.
+
+## Security
+<p>As HDS password input provides only the front-end solution, remember that you need to take care of the security in your project.
```
laurakarhu, line 38: File input, not password

laurakarhu (approved):
One minor copy-paste error in documentation, otherwise looks good!

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] make button unfocusable

packages/react/src/components/index.ts
```diff
@@ -9,6 +9,7 @@ export * from './dialog';
 export * from './dropdown';
 export * from './errorSummary';
 export * from './fieldset';
+export * from './fileInput';
```
VilleMiekkoja, line 12: Not sure if we want to export the _ formatBytes_ function also from the file? Though not sure if it matters really

minevala: I only exported it because of the tests. I'm not sure if it has any usage outside the component.

packages/react/src/components/fileInput/index.ts
```diff
@@ -0,0 +1 @@
+export * from './FileInput';
```
VilleMiekkoja, line 1: Should we only export the component here, and leave the _formatBytes_ without export 🤔

site/docs/components/file_input.mdx
minevala: - [ ] fix copy-paste typo

VilleMiekkoja (approved):
Looking good! 👍

packages/react/src/components/fileInput/FileInput.tsx
minevala: - [ ] fix translations

ronijaakkola (approved):

laurakarhu (approved):
Looking good!

VilleMiekkoja (approved):
Well done! 🎉

PR #523, opened 2021-09-01
Fix/number input on change

VilleMiekkoja:
## Description

Fix a bug in number input where onChange was not triggered when user changes value through stepper buttons

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-947
https://github.com/facebook/react/issues/22212

## How Has This Been Tested?

- Through automated unit tests

### DIFF
--- a/packages/react/src/components/numberInput/NumberInput.test.tsx
+++ b/packages/react/src/components/numberInput/NumberInput.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
+import { act } from 'react-dom/test-utils';
 
 import { NumberInput } from './NumberInput';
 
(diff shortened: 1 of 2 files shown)


packages/react/src/components/numberInput/NumberInput.tsx
```diff
@@ -120,6 +120,13 @@ export const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
       }
     }, [inputRef, ref]);
 
+    const dispatchNativeOnChangeEvent = (): void => {
```
minevala, line 123: This is some higher-level DOM coding 😁 👍 
Do you know the reason why stepDown- and stepUp-method do not trigger native onChange?

packages/react/src/components/numberInput/NumberInput.test.tsx
```diff
@@ -38,4 +39,30 @@ describe('<NumberInput /> spec', () => {
     const numberInput = screen.getByLabelText('Test label number input', { selector: 'input' });
     expect(numberInput).toHaveValue(0);
   });
+  it('should call onChange when input value is changed directly from input field', async () => {
```
minevala, line 42: A linebreak between test cases would make it more readable, at least in my eyes 👀

minevala (commented):
Nice 👍 
I added a small comment and a question

packages/react/src/components/numberInput/NumberInput.tsx
VilleMiekkoja: No idea. I made an issue about it to facebook/react: https://github.com/facebook/react/issues/22212

packages/react/src/components/numberInput/NumberInput.test.tsx
VilleMiekkoja: - [ ] Add a linebreak between test cases

minevala (approved):
Good improvement + OS contribution 👍

PR #524, opened 2021-09-01
Feature/whatsapp icon

VilleMiekkoja:
## Description

Add whatsapp icon to react and core packages

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-960

image


packages/core/src/svg/some/whatsapp.svg
```diff
@@ -0,0 +1,5 @@
+<svg viewBox="0 0 24 24" version="1.1" role="img" xmlns="http://www.w3.org/2000/svg">
```
ronijaakkola, line 1: It probably makes zero difference but this specifies `version="1.1"` which is not currently included in other SVG icons.

minevala (commented):
Works on the site and packages 👍

packages/core/src/svg/some/whatsapp.svg
minevala: I think it can be removed since it is deprecated and it has no effect on rendering or processing: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version

laurakarhu (approved):
I second to Roni's and Mika's note on the version 1.1, otherwise looks good!

packages/core/src/svg/some/whatsapp.svg
VilleMiekkoja: - [ ] Remove `version=1.1`

minevala (approved):
Nice 👍

PR #525, opened 2021-09-03
Examples/update form examples

VilleMiekkoja:
## Description

Update form examples to use HDS DateInput and PhoneNumber components

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-761

## How Has This Been Tested?

image


packages/react/src/examples/forms/validationStatic.stories.tsx
```diff
@@ -221,6 +228,23 @@ export const Static = () => {
...
+          <div className="hds-example-form__grid-6-6">
+            <div className="hds-example-form__item">
+              <PhoneInput
+                id="phoneNumber"
+                name="phoneNumber"
+                label="Phone number"
+                onChange={formik.handleChange}
+                onBlur={formik.handleBlur}
+                pattern="[0-9]{3}-[0-9]{3}-[0-9]{4}"
+                helperText="Use format 123-456-7890"
```
minevala, line 240: Should we use some format that is more familiar for Finnish users?

packages/react/src/examples/forms/validationStatic.stories.tsx
```diff
@@ -304,15 +328,18 @@ export const Static = () => {
...
             {formik.values.parkingPeriod === 'temporary' && (
               <div className="hds-example-form__item">
-                <TextInput
+                <DateInput
                   id="permitEndDate"
                   name="permitEndDate"
                   label="Permit end date"
                   helperText="Use format DD.MM.YYYY"
-                  onChange={formik.handleChange}
+                  onChange={(value) => {
```
minevala, line 336: I noticed that the form validates the value when the date picker is opened. I think it should not show error messages when the user is about to select a date. I'm not sure if there is a way to prevent this.

minevala (commented):
Nice 👍 
Couple of minor comments

packages/react/src/examples/forms/validationStatic.stories.tsx
VilleMiekkoja: This seemed to happen if you first go to the input field, and click the date picker button after that. If you directly click date picker, then there's no error. I guess there's a simple validate on blur. And the onBlur gets called if you click the date picker button when input is focused.

minevala: Ok, so it might work correctly then

VilleMiekkoja: Hmm.. maybe the onBlur should not happen if we open the datepicker? Not sure if it's possible to do it so that it does not happen

minevala: But I guess it is DateInput issue? 🤔

packages/react/src/examples/forms/validationHybrid.stories.tsx
```diff
@@ -334,12 +359,14 @@ export const Hybrid = () => {
             </div>
             {formik.values.parkingPeriod === 'temporary' && (
               <div className="hds-example-form__item">
-                <TextInput
+                <DateInput
```
ronijaakkola, line 362: It probably would be a good idea to add some kind of check to these now that they use actual DateInput. For example, the permit end date should not be in the past. I do not know how hard implementing a such check would be but I think it could be a good example.

VilleMiekkoja: - [ ] Check that the permit end date should not be in the past

packages/react/src/examples/forms/validationStatic.stories.tsx
VilleMiekkoja: - [ ] Use format that is more familiar to Finnish users

packages/react/src/examples/forms/validationStatic.stories.tsx
VilleMiekkoja: Yes, it might be actually something about our TextInput component, that is applied by DateInput. It seems the onBlur is only passed to the input element, and not to the whole wrapper. That would explain why the onBlur is called, even the focus still stays for the whole wrapper. Not sure if there's any way fixing it without causing a breaking chance. Also as the TextInput is used in so many other components, it might be scary to change its behaviour. To fix the issue we would have to remove the onBlur from the input, and then set it to the input wrapper instead.

VilleMiekkoja: As suggested by a11y audit, this can be dealt with by calling the onBlur validation only if the field is dirty:

- [ ] Validate onBlur only if the field is dirty

laurakarhu (approved):
Looking good, thanks!

packages/react/src/examples/forms/validationHybrid.stories.tsx
minevala: Hmm, it seems it does not validate when I change the date by using keyboard 🤔

minevala (approved):
Nice 🏅

PR #526, opened 2021-09-06
[Change] Documentation example heading color

ronijaakkola:
## Description
Since Hel.fi renewal no longer uses brand colour in text headings, we will update HDS to follow that guideline. This PR updates HDS form and form validation example images to use color `--color-black-90` instead of `--color-bussi`. This includes both documentation site image examples and interactive examples in the React Storybook.

## Related Issue
https://helsinkisolutionoffice.atlassian.net/browse/HDS-840

## How Has This Been Tested?
Tested by running the documentation and React Storybook site locally.

image


packages/react/src/examples/forms/validation.scss
```diff
@@ -3,14 +3,14 @@
   max-width: 792px;
 
   &__main-title {
-    color: var(--color-bus);
+    color: var(--color-black-90);
```
minevala, line 6: It might be unnecessary to declare titles' colors anymore since they 
use the same color as the body text.

minevala (commented):
👍 
Small comment

packages/react/src/examples/forms/validation.scss
ronijaakkola: Ty, removed it.

minevala (approved):
Nice 👍

laurakarhu (approved):
Nice, thanks!

PR #530, opened 2021-09-08
[Change] Accordion improvements

VilleMiekkoja:
## Description

1. Make accordion open via header click. Previously it opened only from the accordion button (caret down icon)
2. Support accordion being open by default.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-642
https://helsinkisolutionoffice.atlassian.net/browse/HDS-966

## How Has This Been Tested?

- Automated unit tests

[👉 Storybook Demo](https://city-of-helsinki.github.io/hds-demo/accordion/?path=/story/components-accordion--card-accordion)

image


ronijaakkola:
This looks great! 

I will add a small note to the documentation about the usage of `initiallyOpen`.

laurakarhu (approved):
Looking good!

packages/react/src/components/accordion/Accordion.tsx
```diff
@@ -113,24 +132,30 @@ export const Accordion = ({
...
-      <div className={styles.accordionHeader}>
-        <div role="heading" aria-level={headingLevel} id={`${accordionId}-heading`}>
-          {heading}
-        </div>
-        <button
-          type="button"
-          {...buttonProps}
-          className={styles.accordionButton}
+      <div className={classNames(styles.accordionHeader)}>
+        <div
```
minevala, line 136: I'd guess this is not a button but div because of heading etc.?

packages/react/src/components/accordion/Accordion.test.tsx
```diff
@@ -14,4 +15,13 @@ describe('<Accordion /> spec', () => {
     const results = await axe(container);
     expect(results).toHaveNoViolations();
   });
+  it('should open the accordion when accordion header is clicked', async () => {
```
minevala, line 18: I would add a line break between test cases for readability

packages/react/src/components/accordion/Accordion.stories.tsx
```diff
@@ -23,6 +23,9 @@ export default {
```
minevala, line 23: Could you fix a typo in the story arguments: buttton -> button?
https://github.com/City-of-Helsinki/helsinki-design-system/blob/7bb2fec51ab86ad54aff123792b14ee21899e846/packages/react/src/components/accordion/Accordion.stories.tsx#L19

minevala (commented):
Nice implementation 👍 
I added a couple of tiny comments 🤏

ronijaakkola (approved):

packages/react/src/components/accordion/Accordion.test.tsx
VilleMiekkoja: - [ ] add a line break between test cases for readability

packages/react/src/components/accordion/Accordion.stories.tsx
VilleMiekkoja: - [ ] Fix typo in the story arguments: buttton -> button

packages/react/src/components/accordion/Accordion.tsx
VilleMiekkoja: Good catch. We could apply directly the button tag here. Though tested it and it would require some more work since user agents of the browsers will add some styles based on the button tag. Maybe we go with this for simplicity for now.

minevala (approved):
Good job 👍

PR #531, opened 2021-09-08
Replace post-css compose with common SASS file

minevala:
## Description
- Replace Postcss compose with common SASS file

## Related Issue
- https://helsinkicity.slack.com/archives/CHCV3KTHA/p1631016008034500

## Motivation and Context
- Postcss's compose might generate invalid CSS when including parts of SASS files.  CSS is valid only if it is built in a specific order. This problem occurs when projects are testing their products using the HDS library and Jest, for example.

## How Has This Been Tested?
- locally on dev machine


VilleMiekkoja (approved):
Great 👍

laurakarhu (approved):
Good catch!

PR #541, opened 2021-09-28
Feature/basic table

VilleMiekkoja:
## Description

Implements basic table styles to HDS core

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-815

## How Has This Been Tested?

👉 [Storybook demo](https://city-of-helsinki.github.io/hds-demo/table/?path=/story/components-table--dark)

image
image
image


packages/core/src/components/table/table.css
```diff
@@ -0,0 +1,109 @@
...
+  border-left: 1px solid var(--color-black-10);
+}
+
+.hds-table--with-vertical-lines.hds-table--with-vertical-header.hds-table--light .hds-table__header-row {
+  border-left: none;
+}
+
+.hds-table__content tr td {
+  font-weight: 400;
+  padding: 12px 24px;
```
minevala, line 73: Is it possible to replace these values with hds-variables?

packages/core/src/components/table/table.css
```diff
@@ -0,0 +1,109 @@
...
+  border-bottom: 1px solid var(--color-black-50);
+  font-size: 16px;
```
minevala, line 75: Is it possible to replace this value with hds-variables?

packages/core/src/components/table/table.css
```diff
@@ -0,0 +1,109 @@
...
+.hds-table--with-vertical-lines .hds-table__header-row {
+  border-right: 1px solid var(--color-bus);
+  border-left: 1px solid var(--color-bus);
+}
+
+.hds-table--with-vertical-lines.hds-table--with-vertical-header .hds-table__header-row {
+  border-left: none;
+}
+
+.hds-table--with-vertical-lines.hds-table--light .hds-table__header-row {
```
minevala, line 62: A suggestion: Would it make this selector more readable if it would start with the class that makes it different? In this case, it would be .hds-table--light.

packages/core/src/components/table/work-trial.json
```diff
@@ -0,0 +1,618 @@
+[
```
minevala, line 1: The name of the file could have some example or story-example prefix so we would know it has nothing to do with the implementation.

packages/core/src/components/table/table.stories.js
```diff
@@ -0,0 +1,351 @@
...
+          <td class="hds-table--text-align-right">${work['Postinumero']}</td>
+          <td class="hds-table--text-align-right">${work['Paikkoja']}</td>
+          <td class="hds-table--text-align-right">${work['Haastatteluun halutaan']}</td>
+        </tr>`
+  })
+  return content;
+}
+
+export const WideAndLongExample = () => `
+  <div class="hds-table-container" style="height: 600px;" tabindex="0">
```
minevala, line 333: I'm not sure should we even try to fix focus behavior in the core package because the table might fit into screen in some cases 🤔 
On the other hand, is it bad if a large table without scrollbars does have this behavior?

minevala (commented):
Good and clean PR 👍 
I had a few small comments and questions

packages/core/src/components/table/table.stories.js
VilleMiekkoja: Well, in this case the table won't fit into the container, as there's only 600px height for it. But this is a good point. Not sure if we should include `tabindex="0"` to every table, since in mobile view they won't fit to the container?

packages/core/src/components/table/table.css
```diff
@@ -0,0 +1,109 @@
...
+
+.hds-table {
+  border-collapse: collapse;
+  font-family: var(--font-default); /* Tartteeko laittaa? */
+  width: 100%;
+  line-height: 1.5;
+}
+
+.hds-table--dark {
+  --background-color: var(--color-bus);
```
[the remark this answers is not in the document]
VilleMiekkoja: Can you elaborate more of the abstraction you had in mind?

packages/core/src/components/table/table.css
VilleMiekkoja: - [ ] Start with the class that makes it different from previous

packages/core/src/components/table/work-trial.json
VilleMiekkoja: True, I'll prefix it:

- [ ] Prefix file to emphasise it's only example data

packages/core/src/components/table/table.css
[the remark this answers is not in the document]
minevala: For example this could be:
```css
.hds-table--dark {
  --background-color: var(--background-color);
  }
 ```
 Then we could declare somewhere higher in the scope:
 ```css
  --background-color: var(--color-bus); 
 ```
In this way, the background color is already customizable without theming via CSS variables. It would also be in line with older components if we ever decide to continue with themes.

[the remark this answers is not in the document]
Comment (minevala) [reply packages/core/src/components/table/table.css:13]: Plaah, I think I have linked into the wrong line 🙈 
I think some of the borders could be part of the css variables 🤔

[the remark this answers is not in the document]
Comment (minevala) [reply packages/core/src/components/table/table.css:13]: Or maybe not, I'll leave this for you to consider 😁

File: packages/core/src/components/table/table.css [hunk 7]
```diff
@@ -0,0 +1,109 @@
+.hds-table-container {
+  overflow-x: auto;
+}
+
+.hds-table {
+  border-collapse: collapse;
+  font-family: var(--font-default); /* Tartteeko laittaa? */
```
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Remove this for now

packages/core/src/components/table/table.stories.js
VilleMiekkoja: We discussed this on todays a11y meeting, and decided to set all hds tables focusable, to guarantee a11y. 

- [ ] Set all tables focusable

packages/core/src/components/table/table.css
VilleMiekkoja: - [ ] Replace with hds-variables

packages/core/src/components/table/table.css
VilleMiekkoja: - [ ] Replace with hds-variables

minevala (approved):
Looks good 👍

VilleMiekkoja:
Tasks remaining:

- [ ] Dense variant of the table (uses mobile styles)
- [ ] Table focus styles

packages/core/src/components/table/table.css
```diff
@@ -0,0 +1,122 @@
...
+}
+
+.hds-table th {
+  border-bottom: 1px solid var(--color-black-20);
+  white-space: nowrap;
+  text-align: left;
+  font-weight: 500;
+  font-size: var(--fontsize-heading-xxs);
+  padding: var(--spacing-xs) var(--spacing-m);
+}
```
minevala, line 30: Suggestion: would it be easier to maintain if default styles and variant styles would be in groups like in the button.css?

minevala (approved):
Now HDS will have nice tables. Looks good 👍

ronijaakkola:
@VilleMiekkoja Updated some of the Playground example data sets to better show when some of the variants should be used. Maybe we should update these to Storybook as well?

PR #543, opened 2021-09-29
Feature/new doc init

minevala:
## Description
- A frame for new documentation site
- based on Gatsby default starter: https://www.gatsbyjs.com/starters/gatsbyjs/gatsby-starter-default

## Motivation and Context
- This is base work for the new documentation site

## How Has This Been Tested?
- Locally on dev's machine

Screenshot 2021-09-29 at 17 23 25


new-site/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+yarn build
+```
+
+4. Start the development server. It will watch for file changes automatically.
+```
+cd new-site
+yarn start
+```
+
```
laurakarhu, line 42: If we want to improve the guidance, either remove the yarn start from just above or add a subtitle before the step, saying "If starting the new docsite fails, install missing local dependencies".

laurakarhu (approved):
It's alive :D

new-site/README.md
```diff
@@ -0,0 +1,57 @@
...
+5. Install local dependencies.
+```
+yarn
+```
+
+
+6. Start the development server. It will watch for file changes automatically.
```
VilleMiekkoja, line 49: This seems to be duplicate of 4.?

new-site/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "dependencies": {
+    "gatsby": "^3.13.1",
+    "gatsby-plugin-image": "^1.13.0",
+    "gatsby-plugin-manifest": "^3.13.0",
+    "gatsby-plugin-offline": "^4.13.0",
+    "gatsby-plugin-react-helmet": "^4.13.0",
+    "gatsby-plugin-sharp": "^3.13.0",
+    "gatsby-source-filesystem": "^3.13.0",
+    "gatsby-transformer-sharp": "^3.13.0",
+    "hds-core": "^1.5.0",
```
VilleMiekkoja, line 15: Would it be possible to use hds-core and hds-react directly from the repository instead of as a dependency? 

One possible benefit would be that then the site is always aligned with the newest versions of core and react. No need to update these version numbers. Developing the site would be faster, since you don't have to release hds-core and hds-react to get the new component version. But you can just do the modifications to the same pr. I know we could use `yarn link` to link local versions, but it has some drawbacks: kind of slow, and also you need to switch it off by running yarn unlink.

One possible drawback could be that then the way we use those components from core and react in this site would not be similar of how they are used by consumer projects of HDS. The bundled versions of HDS might introduce a bug that would not show itself if we just directly import the component. 🤔 

Slightly related, Im not sure which option would be better for the feature of supporting multiple versions of HDS documentation in the site.

new-site/.gitignore
```diff
@@ -0,0 +1,69 @@
...
+coverage
+
+# nyc test coverage
+.nyc_output
+
+# Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
+.grunt
+
+# Bower dependency directory (https://bower.io/)
+bower_components
```
VilleMiekkoja, line 27: It might be good idea to check if these all are relevant for us?

new-site/gatsby-node.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Implement Gatsby's Node APIs in this file.
+ *
+ * See: https://www.gatsbyjs.com/docs/node-apis/
+ */
+
+// You can delete this file if you're not using it
```
VilleMiekkoja, line 7: Maybe we delete this for now?

new-site/gatsby-ssr.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Implement Gatsby's SSR (Server Side Rendering) APIs in this file.
+ *
+ * See: https://www.gatsbyjs.com/docs/ssr-apis/
+ */
+
+// You can delete this file if you're not using it
```
VilleMiekkoja, line 7: We could add this also later if we apply gatsby ssr

new-site/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "hds-react": "^1.5.0",
+    "prop-types": "^15.7.2",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "react-helmet": "^6.1.0"
+  },
+  "devDependencies": {
+    "hds-design-tokens": "^1.5.0",
```
VilleMiekkoja, line 23: I wonder should this be in the dependencies section, as hds-core and hds-react are there too?

new-site/src/components/layout.js
```diff
@@ -0,0 +1,59 @@
...
+  const title = siteDate?.title || 'Title';
+  const contentId = 'content';
+
+  return (
+    <div className="page">
+      <Header contentId={contentId} className="pageHeader" siteTitle={title} />
+      <Container id={contentId} className="pageContent" style={{ margin: '0 auto' }}>
+        <main>{children}</main>
+      </Container>
+      <Footer
```
VilleMiekkoja, line 44: I like the abstraction, but it kind of makes it also so that the user who reads this root layout, does not directly notice that the components used here come from hds-react. It also gives an example that the way these components should be used is that you should make abstractions out of them. I wonder if that is what we want?

minevala: I can remove the abstraction

new-site/package.json
minevala: I'm not sure how does it work in build step?

new-site/gatsby-node.js
minevala: - [ ] delete unused file

new-site/gatsby-ssr.js
minevala: - [ ] delete unused file

new-site/src/components/layout.js
minevala: - [ ] remove unnecessary abstraction

new-site/package.json
VilleMiekkoja: Yeah it might be technically harder. We could think about it later 👍

VilleMiekkoja:
Excellent job with this one! I added few small comments here and there.

new-site/package.json
minevala: - [ ] fix dependencies

new-site/src/components/layout.js
minevala: > Though I guess in reality people end up making those abstractions, like we did in Hauki 😅

I think we used abstractions because we needed to override some styles in common components like buttons so they became hauki-buttons 🤔

new-site/package.json
ronijaakkola: > One possible benefit would be that then the site is always aligned with the newest versions of core and react. No need to update these version numbers. Developing the site would be faster, since you don't have to release hds-core and hds-react to get the new component version.
As a document writer, I would love this feature so I could easily create documentation in the feature branch.

PR #544, opened 2021-10-05
Fix/Allow reseting number input to an empty value

VilleMiekkoja:
## Description

NumberInput did not allow reseting the component to an empty value when used as a controlled component. This PR fixes it by allowing users reset to an empty value for controlled component.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1013

## How Has This Been Tested?

- Automated unit test

### DIFF
--- a/packages/react/src/components/numberInput/NumberInput.test.tsx
+++ b/packages/react/src/components/numberInput/NumberInput.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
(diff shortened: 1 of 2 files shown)


laurakarhu (approved):
Thanks, looks good to me :)

packages/react/src/components/numberInput/NumberInput.test.tsx
```diff
@@ -72,4 +72,30 @@ describe('<NumberInput /> spec', () => {
...
+        <div>
+          <button
+            type="button"
+            onClick={() => {
+              setValue('');
+            }}
+          >
+            Reset number input
+          </button>
+          <NumberInput step={10} value={value} {...numberInputProps} />
```
minevala, line 89: An observation:
I noticed that this would require the onChange handler to be functional.
I tested it in a story, and I needed to add: 
```js
 onChange={(e) => {
    const eventValue = e.target.value;
    const newValue = eventValue === '' ? eventValue : parseInt(eventValue, 10);
    setValue(newValue);
 }}
```

minevala (approved):
Works 👍 
I was thinking should we have a separate guide how to implement reset button 🤔 
I noticed that the one in tests does not function properly if user tries to edit the value.

PR #545, opened 2021-10-05
Call onChange inside update functions to ensure it is not called on r…

minevala:
## Description
-The FileInput component called onChange property callback on component re-render without any file changes. Now onChange callback is called inside update functions to prevent unnecessary calls.

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1012

## How Has This Been Tested?


laurakarhu (approved):
Thanks!

PR #556, opened 2021-10-18
Feature/new docs side navigation

minevala:
## Description
- Add SideNavigation to doc site
- Convert all pages to .mdx format
- Fix Seo
- Fix font styles
- Add ability to host site in subfolders (Github pages)

## Closes #
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-914

## How Has This Been Tested?
- Locally on dev machine

👉 [Site Demo](https://city-of-helsinki.github.io/hds-demo/docsite-sidenav/)

Screenshot 2021-10-18 at 13 22 50


new-site/src/components/seo.js
```diff
@@ -1,51 +1,33 @@
...
 
+function Seo({ lang, meta, title, description, pageTitle }) {
   return (
     <Helmet
       htmlAttributes={{
         lang,
       }}
       title={title}
-      titleTemplate={defaultTitle ? `%s | ${defaultTitle}` : null}
+      titleTemplate={title ? `${title} | ${pageTitle}  ` : null}
```
VilleMiekkoja, line 18: Do we need the titleTemplate? I tried to study what it means, and it appears to be used in a case where react-helmet's `Helmet` component is used twice; the first usage has the titleTemplate defined, and then the second usage provides only title, which is then appended to the titleTemplate. Like in this example: https://dev.to/guimg/dynamic-document-head-with-react-helmet-28o6

I could not find the second usage of the `Helmet` component in this repository. Am I missing something, or have I understood the usage wrong?

minevala: - [ ] simplify title fields

VilleMiekkoja (approved):
Nice job! 🎉

PR #570, opened 2021-11-02
Feature/new docs playground

minevala:
## Description
- Add custom playground component which is more accessible and easier to use for new documentation
- Show Syntaxblock for other code blocks (non-playground blocks)

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1021

## Motivation and Context
- HDS is using Docz's playground component to provide a live example of components. This PR uses the same component react-live underneath. But it is more customized for HDS needs, accessibility, more apparent code block, etc.

## How Has This Been Tested?
- Locally on dev's machine

Screenshot 2021-11-02 at 13 03 59
Screenshot 2021-11-02 at 13 03 39

👉 [Storybook Demo](https://city-of-helsinki.github.io/hds-demo/docsite-playground/elements/components/buttons/)

### DIFF
--- a/new-site/src/components/Playground.js
+++ b/new-site/src/components/Playground.js
@@ -128,14 +128,13 @@ const Editor = ({ onChange, initialCode, code, language }) => {
         </div>
       </div>
       <div className="playground-block-buttons">
-        <Button ref={copyButtonRef} variant="secondary" size="small" tabIndex={0} onClick={copy}>
+        <Button ref={copyButtonRef} variant="secondary" size="small" onClick={copy}>
           Copy code
         </Button>
         <Button
           variant="secondary"
           iconLeft={<IconArrowUndo aria-hidden />}
           size="small"
-          tabIndex={0}
           disabled={initialCode === code}
           onClick={reset}
         >
--- a/new-site/src/components/Playground.js
+++ b/new-site/src/components/Playground.js
@@ -119,15 +119,7 @@ const Editor = ({ onChange, initialCode, code, language }) => {
               <LiveEditor
                 key={resetCount}
                 onChange={onChange}
-                style={{
-                  background: 'none',
-                  padding: '0 0 var(--spacing-s)',
-                  overflow: 'visible',
-                  whiteSpace: 'nowrap',
-                  width: 'max-content',
-                  minWidth: '100%',
-                  fontSize: '16px'
-                }}
+                className="playground-block-editor-code-input"
                 textareaId={textAreaId}
                 theme={theme}
               />
--- a/new-site/src/components/Playground.js
+++ b/new-site/src/components/Playground.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { LiveProvider, LiveEditor, LiveError, LivePreview, withLive } from 'react-live';
 import * as Hds from 'hds-react';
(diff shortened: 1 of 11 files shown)


laurakarhu (approved):
So cool!

ronijaakkola:
To me it seems the the code font size is slightly too small. In designs it seems to match the font size of the help text but the implementation does not use the same code font so it might look a bit different.

The code font in the implementation is 13px size. I would try maybe 16px for the Monospace font. If that is too big, then maybe 14px.

image

minevala:
> To me it seems the the code font size is slightly too small. In designs it seems to match the font size of the help text but the implementation does not use the same code font so it might look a bit different.
> 
> The code font in the implementation is 13px size. I would try maybe 16px for the Monospace font. If that is too big, then maybe 14px.
> 
> image

I need to check this. It is using Courier 16px on my laptop 🤔

new-site/src/components/Playground.js
```diff
@@ -0,0 +1,226 @@
...
+              <span className="playground-block-editor-helper" id={helperTextId}>
+                Press Enter to start editing. Press Esc to stop editing.
+              </span>
+            </div>
+            <div className="playground-block-editor-scrollbox" tabIndex={-1}>
+              <LiveEditor
+                key={resetCount}
+                onChange={onChange}
+                style={{
+                  background: 'none',
```
VilleMiekkoja, line 123: Should we put these styles to css file, or is it better to inline them here?

new-site/src/components/Playground.js
```diff
@@ -0,0 +1,226 @@
...
+  };
+
+  useEffect(() => {
+    if (viewPortRef.current) {
+      const textArea = getTextArea(viewPortRef.current);
+      textArea.setAttribute('aria-describedby', helperTextId);
+      textArea.setAttribute('tabIndex', '-1');
+      textArea.addEventListener('blur', onTextAreaBlur);
+    }
+  }, [getTextArea, helperTextId, onTextAreaBlur]);
```
VilleMiekkoja, line 98: I guess technically the `onTextAreaBlur` can never change in the current setting, as it's wrapped inside a `useCallback` hook that has an empty list as a dependency? But React probably still requires it to be set as a dependency for the useEffect.

new-site/src/components/Playground.js
```diff
@@ -0,0 +1,226 @@
...
+                }}
+                textareaId={textAreaId}
+                theme={theme}
+              />
+            </div>
+          </div>
+        </div>
+      </div>
+      <div className="playground-block-buttons">
+        <Button ref={copyButtonRef} variant="secondary" size="small" tabIndex={0} onClick={copy}>
```
VilleMiekkoja, line 139: Do we need the `tabIndex={0}` here? Buttons should be focusable by default.

new-site/src/components/Playground.scss
```diff
@@ -0,0 +1,106 @@
...
+
+@import '~hds-design-tokens/lib/all.scss';
+
+.playground-block {
+  --focus-border-width: 2px;
+  margin: var(--spacing-l) 0;
+}
+
+.playground-block-tabs {
+  --tablist-border-color: transparent;
```
VilleMiekkoja, line 11: I guess this overrides the `--tablist-border-color` variable defined in Tabs.module.scss? Does it depend on the css order for which one of them takes precedence over another?

VilleMiekkoja:
Works and looks good! 👍  Added couple minor comments.

new-site/src/components/Playground.js
minevala: The LiveEditor component seems to take only style property. It passes those styles to underlying textarea and pre components. I guess it is possible to give the className to parent LiveProvider component and then override the child element styles in css file 🤔

new-site/src/components/Playground.js
minevala: - [ ] simplify onTextAreaBlur

new-site/src/components/Playground.js
minevala: - [ ] remove unnecessary tabIndexes

new-site/src/components/Playground.scss
minevala: I believe it overrides the root value when the property is given in the element scope 🤔

new-site/src/components/Playground.js
minevala: Hmm, it might actually take to className as well. I'll try to use that one

minevala: - [ ] replace live-editor inline styles with classname

ronijaakkola (approved):
This is beautiful.

VilleMiekkoja (approved):
Good job with this one! 🥇

PR #574, opened 2021-11-04
Fix/date input keyboard navigation

VilleMiekkoja:
## Description

Date picker doesn’t allow focusing all dates with keyboard if there is min date set. Dates above the min and below the max should be focusable with keyboard. This PR fixes the issue.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-861

## How Has This Been Tested?

- Automated unit tests

:mega: [Storybook demo](https://city-of-helsinki.github.io/hds-demo/datepicker-keyboard-navigation-fix/?path=/story/components-dateinput--with-min-date)


ronijaakkola (approved):
Cannot really comment the technical fix but I can confirm that the functionality is correct. 👍

laurakarhu (approved):
Seems to be working as expected :) There is one accessibility violation in Storybook, is this real and can we fix it?

packages/react/src/components/dateInput/DateInput.test.tsx
```diff
@@ -317,4 +317,36 @@ describe('<DateInput /> spec', () => {
...
+    });
+
+    await act(async () => {
+      userEvent.tab();
+      userEvent.tab();
+      userEvent.tab();
+      userEvent.tab();
+    });
+
+    const currentDateButton = container.querySelector('button[data-date="2021-01-17"]');
```
minevala, line 348: I was wondering if we could use a screen.getByRole - query etc. here. But then I noticed that the button has two spans. Maybe it is not worth it 🤔

minevala (approved):
Nice fix 🏅

VilleMiekkoja:
> Seems to be working as expected :) There is one accessibility violation in Storybook, is this real and can we fix it?

That is weird, for me it shows no errors 🤔 

image

PR #575, opened 2021-11-08
Fix/select null resetting

minevala:
## Description
- Accept null as single Select defaultValue and value

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1034

## Motivation and Context
- Resetting Select value with null (as it should be with controlled components) causes a Typescript error

## How Has This Been Tested?
- Locally on dev's machine

### DIFF
--- a/packages/react/src/components/dropdown/select/Select.stories.tsx
+++ b/packages/react/src/components/dropdown/select/Select.stories.tsx
@@ -96,7 +96,7 @@ export const Controlled = (args) => {
         style={{ marginTop: 'var(--spacing-s)' }}
       />
 
-      <Button onClick={() => setSelectedItems(null)} style={{ marginTop: 'var(--spacing-l)' }}>
+      <Button onClick={() => setSelectedItems([])} style={{ marginTop: 'var(--spacing-l)' }}>
         Reset
       </Button>
       <Button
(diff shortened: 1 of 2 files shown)


packages/react/src/components/dropdown/select/Select.stories.tsx
```diff
@@ -96,7 +96,7 @@ export const Controlled = (args) => {
         style={{ marginTop: 'var(--spacing-s)' }}
       />
 
-      <Button onClick={() => setSelectedItems(null)} style={{ marginTop: 'var(--spacing-l)' }}>
+      <Button onClick={() => setSelectedItems([])} style={{ marginTop: 'var(--spacing-l)' }}>
```
Nizar-Rahme, line 99: I got one question, why don't we keep this as `null` since the component will now accept `null` values?

minevala: At the moment, only the single Select accepts null as values. Should the multi Select accept nulls as well?
I can fix it but I was not sure about it at first

Nizar-Rahme: Ah, I didn't notice that this story is for the multi-select. Not sure if making it nullable will be needed in some scenarios 🤷    

But looks good to me :)

laurakarhu (approved):
Looks that null gets accepted :)

VilleMiekkoja (approved):
Nice! 👍

PR #579, opened 2021-11-10
Upgrade dependencies [Dependabot alerts]

minevala:
## Description
- Regenarate yarn.lock to update libraries
- Fix react-types conflict
- Update eslint-import
- Add ts-jest to fix nullish coalescing errors in tests
- Update jest-libraries and config to fix tests in node v12

React-types conflict: https://github.com/yarnpkg/yarn/issues/4489#issuecomment-360812460
Eslint no-import error fix: https://github.com/airbnb/javascript/issues/2331#issuecomment-724114465 
Ts-jest: https://github.com/kulshekhar/ts-jest

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1089

## Motivation and Context
- Fixes a bunch of dependabot warnings

## How Has This Been Tested?
- Locally on dev machine

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -56,14 +56,15 @@
     "@testing-library/react": "10.0.4",
     "@testing-library/user-event": "^12.8.1",
     "@types/cleave.js": "1.4.4",
-    "@types/jest": "25.2.3",
-    "@types/jest-axe": "3.5.1",
+    "@types/jest": "^26.0.1",
+    "@types/jest-axe": "^3.5.3",
     "@types/react": "16.9.50",
     "@types/react-dom": "16.9.8",
     "@types/yup": "^0.29.9",
     "@typescript-eslint/eslint-plugin": "2.34.0",
     "@typescript-eslint/parser": "2.34.0",
     "@wessberg/rollup-plugin-ts": "1.3.5",
+    "babel-jest": "^26.0.1",
     "babel-plugin-require-context-hook": "1.0.0",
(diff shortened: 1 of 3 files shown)


packages/react/package.json
```diff
@@ -78,9 +79,9 @@
     "formik": "^2.2.5",
     "identity-obj-proxy": "3.0.0",
     "inquirer": "7.1.0",
-    "jest": "26.0.1",
-    "jest-axe": "4.1.0",
-    "jest-environment-jsdom-sixteen": "^1.0.3",
+    "jest": "^26.0.1",
```
VilleMiekkoja, line 82: Wonder if we wish to include this '^' or let it be as it was before? There seems to be no common pattern for this. Though maybe defining explicit version is little more common?

VilleMiekkoja (approved):
Good job! 🎉 One minor comment.

packages/react/package.json
minevala: I can change that but it should be pretty safe to use the "hat". Then it would automatically use the latest minor or patch versions. But then again it might be risky if some library does not follow semver versioning strictly.

laurakarhu (approved):
Thanks!

packages/react/package.json
minevala: I guess we need to use the hat to get the library updates and fixes. I found at least five library updates between the oldest major version and the latest one.

PR #580, opened 2021-11-11
Feature/react table

VilleMiekkoja:
## Description

Implements Table component for hds-react

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-646

## How Has This Been Tested?

1. Locally on developers machine
2. Automated unit tests
3. Automated visual regression tests (loki tests)

:point_right: [Storybook demo](https://city-of-helsinki.github.io/hds-demo/react-table/?path=/story/components-table--dark)

image
image


packages/react/src/components/table/Table.stories.tsx
```diff
@@ -0,0 +1,934 @@
...
+    <div style={{ maxWidth: '640px' }}>
+      <DataTable caption={caption} cellConfig={cellConfig} rows={rows} />
+    </div>
+  );
+};
+
+// args is required for docs tab to show source code
+// eslint-disable-next-line no-unused-vars
+export const Dark = (args) => {
+  const cellConfig = {
```
VilleMiekkoja, line 63: Have to repeat these DataTable prop initialisations for them to show in the story tab of storybook.

packages/react/src/components/table/Table.stories.tsx
```diff
@@ -0,0 +1,934 @@
...
+
+  const [tableRows, setTableRows] = useState(rows);
+  const [selectedRows, setSelectedRows] = useState([]);
+
+  const myRef = useRef();
+
+  return (
+    <div style={{ maxWidth: '640px' }}>
+      <div style={{ marginBottom: '16px', display: 'flex', flexDirection: 'row', justifyContent: 'flex-end' }}>
+        <Button
```
VilleMiekkoja, line 737: Should these 'Select all rows', 'Deselect all rows' and 'Delete selected' buttons be a part of DataTable implementation, or should we leave them out as they currently are?

packages/react/src/components/table/Table.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  it('renders the component', () => {
+    const { asFragment } = render(<DataTable caption={caption} cellConfig={cellConfig} rows={rows} />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+  it('should not have basic accessibility issues', async () => {
+    expect(true).toBe(true);
+    const { container } = render(<DataTable caption={caption} cellConfig={cellConfig} rows={rows} />);
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
```
minevala, line 54: Should we have some basic functional sorting test: library user provides a custom function, then user clicks and the row order changes accordingly or something like that?

packages/react/src/components/table/DataTable.tsx
```diff
@@ -0,0 +1,266 @@
...
+    }
+    if (aValue > bValue) {
+      return order === 'asc' ? 1 : -1;
+    }
+
+    return 0;
+  });
+}
+
+export const DataTable = React.forwardRef(
```
VilleMiekkoja, line 93: Should we change this name to be just 'Table'? I named it to DataTable because there was already a Table component that it applies. If we wish to name this to just 'Table', we need to figure a new name to the Table component that it uses.

packages/react/src/components/table/Table.module.scss
```diff
@@ -0,0 +1,117 @@
...
+.headerRow {
+  composes: hds-table__header-row from 'hds-core/lib/components/table/table.css';
+}
+
+.content {
+  composes: hds-table__content from 'hds-core/lib/components/table/table.css';
+}
+
+.dark {
+  --background-color: var(--color-bus);
```
VilleMiekkoja, line 31: Why not just compose the hds-table--dark from core? The reason was that I could then not get the custom theme for background color for react version of the table to work. Maybe we should not use the `useTheme` at all here? User could just pass a className where he overrides the variable? Problem with that is, that it is harder to demonstrate in a story, as passing a className hides the content of the className, which would be the beef here. 🤔

packages/react/src/components/table/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Table';
+export * from './DataTable';
```
VilleMiekkoja, line 2: Not really sure how the exporting of multiple components work here. The idea was that people would be able to compose their own table from the components that the `Table` component provides (look at the end of Table component file for more info). But also we would expose the actual complete Table, named DataTable here. Im not sure if we need/want people to be able to compose their own table from sub-components of Table? Would DataTable be enough already?

packages/react/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "sourceMap": true,
     "declaration": true,
     "esModuleInterop": true,
+    "resolveJsonModule": true,
```
VilleMiekkoja, line 14: Reading a json file inside stories required this flag for hds-react.

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,53 @@
...
+        )}
+        {...rest}
+      >
+        {children}
+      </table>
+    </div>
+  );
+};
+
+Table.TableBody = TableBody;
```
VilleMiekkoja, line 50: Do we need/want these sub-components to be public? At some point we had the idea of people being able to compose their own table. But it was when we thought that more complex features like sorting and checkbox selection would not be reasonable to do otherwise than using composition and story examples. Now that we actually managed to implement them directly from DataTable api, Im not sure if we need/benefit that much from exposing these? It does increase the complexity and a bit and then makes it harder to change the sub-components later on without breaking changes 🤔

packages/react/src/components/table/Table.test.tsx
VilleMiekkoja: - [ ] Add unit test for sorting

packages/react/src/components/table/Table.tsx
minevala: It would help to decide between DataTable and/or Table composition if we knew how tables are used in Helsinki Kaupunki projects. At first, it would feel easier to start with a concise API, but I guess it could be verified. And I'm a bit worried that we can not make breaking changes to new components because then we have to get them right on the first try, and then they are locked until the next major release 🤔

packages/react/src/components/table/Table.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <span>
+        <b>Table 1</b>: Table description
+      </span>
+    );
+  });
+
+  it('renders the component', () => {
+    const { asFragment } = render(<DataTable caption={caption} cellConfig={cellConfig} rows={rows} />);
+    expect(asFragment()).toMatchSnapshot();
+  });
```
minevala, line 48: I have to fix the generating script at some point. It does not add a line break between test cases.

packages/react/src/components/table/DataTable.tsx
```diff
@@ -0,0 +1,266 @@
...
+  variant?: 'dark' | 'light';
+  dense?: boolean;
+  zebra?: boolean;
+  caption?: string | React.ReactNode;
+  verticalLines?: boolean;
+  textAlignContentRight?: boolean; // defaults to false -> text is aligned left
+  checkboxSelection?: boolean;
+  ariaLabelCheckboxSelection?: string;
+  setSelections?: Function; // Callback that gets called with all selected row id values
+  initiallySelectedRows?: any[]; // Initially selected rows. Apply corresponding indexKey values here.
```
minevala, line 44: Could the row type be an object with arbitrary key and with known value types: string, number, date, etc.?

packages/react/src/components/table/DataTable.tsx
```diff
@@ -0,0 +1,266 @@
...
+      zebra = false,
+      verticalLines = false,
+      verticalHeaders,
+      textAlignContentRight = false,
+      setSelections,
+      caption,
+      theme,
+      ...rest
+    }: DataTableProps,
+    ref?: any,
```
minevala, line 112: Is this ref for table element?

packages/react/src/components/table/Table.stories.tsx
minevala: Is there some way to read into variable?

packages/react/src/components/table/Table.stories.tsx
minevala: Could the selected be part of the row-data? 🤔 
Then it could be completely outside of the Table component. And initallySelectedRows would be unnecessary

packages/react/src/components/table/Table.stories.tsx
minevala: Or would a function help here? I guess not

VilleMiekkoja: I think we could read a constant into a variable, but then the story tab would only show the read operation like:
`const cellConfig = commonCellConfig`

packages/react/src/components/table/DataTable.tsx
VilleMiekkoja: In the current implementation it's ref for the DataTable component. Only used inside the imperativeHandle hook:

`    useImperativeHandle(ref, () => ({
      selectAllRows() {
        selectAllRows();
      },
      deSelectAllRows() {
        deSelectAllRows();
      },
    }));`

packages/react/src/components/table/Table.test.tsx
VilleMiekkoja: - [ ] Add empty row between unit tests to increase readability

packages/react/src/components/table/DataTable.tsx
minevala: I guess the type could then be HtmlTableRef or something?

VilleMiekkoja: True. I wonder if we should then also pass it to the html table element, as the name suggests?

packages/react/rollup.config.js
```diff
@@ -107,6 +107,7 @@ export default [
       'components/SelectionGroup/index': 'src/components/selectionGroup/index.ts',
       'components/SideNavigation/index': 'src/components/sideNavigation/index.ts',
       'components/StatusLabel/index': 'src/components/statusLabel/index.ts',
+      'components/Table/index': 'src/components/table/index.ts',
```
VilleMiekkoja, line 110: If we choose to have the separate DataTable component as well, then we probably need to add it separately to rollup as well.

packages/react/src/components/table/DataTable.tsx
minevala: Probably 👍

packages/react/rollup.config.js
minevala: I think it might be the easiest way to start with just one table component. We could add composition support later on if someone requests it.

packages/react/src/components/table/Table.tsx
laurakarhu: Nothing stops us from making every release a major release. Of course it's ideal if there isn't breaking changes all the time, but as long as the version numbers reflect the changes correctly.

packages/react/src/components/checkbox/Checkbox.tsx
```diff
@@ -81,7 +81,7 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
           aria-describedby={errorText ? `${id}-error` : null}
           {...rest}
         />
-        <label htmlFor={id} className={classNames(styles.label)}>
+        <label aria-hidden={!label} htmlFor={id} className={classNames(styles.label)}>
```
laurakarhu, line 84: This new property in Checkbox needs to be described in checkbox documentation, when it should be used etc.

packages/react/src/components/table/DataTable.tsx
```diff
@@ -0,0 +1,318 @@
+import React, { useEffect, useMemo, useState } from 'react';
+
+// import core base styles
+import 'hds-core';
+import styles from './Table.module.scss';
+import { Table } from './Table';
+import { Checkbox } from '../checkbox';
+import { useTheme } from '../../hooks/useTheme';
+import { Button } from '../button';
+import { IconTrash } from '../../icons';
```
laurakarhu, line 10: Only needed in custom actions example?

packages/react/src/components/table/index.ts
laurakarhu: Can Table be used ad standalone, or does it always require having DataTable around it? If it always requires DataTable anyway, it might be enough. If Table works alone, it should be available separately.

laurakarhu (approved):
Some remarks, and codespecific review needs a more talented developer, but works nicely :)

packages/react/src/components/table/DataTable.tsx
VilleMiekkoja: - [ ] Remove unused IconTrash from here

packages/react/src/components/checkbox/Checkbox.tsx
VilleMiekkoja: This does not change the direct react component api, so not sure if we need to add it to documentation. 🤔 

It does change the behaviour of the component little though; in situations where the user of the checkbox has not provided visible label, it will hide the (empty) label tag from screenreaders. This change basically makes it possible to not provide label for the checkbox, as earlier a11y tools would complain that empty labels are violating a11y. Why not just remove the whole label element then (when user does not give label)? Problem was that our Checkbox component was programmed in a way that it requires the label tag. If you remove the whole label tag, the checkbox disappears 😅. If we wish to support a checkbox without a label, we might have to make bigger changes to the existing Checkbox implementation.

packages/react/src/components/table/index.ts
VilleMiekkoja: Table works without DataTable, but then you need to add other table contents to it. Table basically just wraps native html table with hds styles to it. We discussed renaming it to TableContainer, so that we could then rename DataTable to just Table.

packages/react/rollup.config.js
VilleMiekkoja: I agree that would be the easiest way to start now. What do you think @laurakarhu ?

packages/react/src/components/checkbox/Checkbox.tsx
VilleMiekkoja: - [ ] Remove unnecessary changes to Checkbox component

packages/react/src/components/table/Table.stories.tsx
VilleMiekkoja: We discussed this and it turned out to be hard/problematic.

packages/react/rollup.config.js
laurakarhu: Sounds like a good plan :)

packages/react/src/components/checkbox/Checkbox.tsx
laurakarhu: I think the checkbox is defined to have a mandatory label and tables are the first exception to that. It was the biggest reason we had a separate ticket for making one without label to be used inside other elements.

packages/react/rollup.config.js
VilleMiekkoja: - [ ] Expose only Table component, rename DataTable -> Table

packages/react/src/components/table/DataTable.tsx
VilleMiekkoja: This has been renamed as 'Table'

packages/react/src/components/table/index.ts
VilleMiekkoja: We decided to publish only Table now (former DataTable), and then publish sub-components later on if they are needed / wanted.

packages/react/src/components/table/Table.tsx
VilleMiekkoja: We decided to publish only Table now (former DataTable), and then publish sub-components later on if they are needed / wanted.

packages/react/src/components/table/DataTable.tsx
VilleMiekkoja: The selectedRow type is actually different from the basic row type. This is only a list of index key values, like:
`[1000, 1001, 1002]`. Maybe we could restrict it string and number?

VilleMiekkoja: - [ ] Create and use SelectedRow typescript type

packages/react/src/components/table/Table.module.scss
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  &:focus {
+    outline: 3px solid var(--color-white);
+  }
+}
+
+/* MOBILE STYLES */
+
+@media (max-width: 767px) {
```
VilleMiekkoja, line 122: Check if can use existing breakpoints

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,345 @@
...
+};
+
+export interface TableCustomTheme {
+  '--background-color'?: string;
+}
+
+type SelectedRow = string | number;
+
+export type TableProps = React.ComponentPropsWithoutRef<'table'> & {
+  ariaLabelCheckboxSelection?: string;
```
VilleMiekkoja, line 29: We should probably add documentation for all of these props?

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,345 @@
...
+        {...rest}
+      >
+        {caption && <caption className={styles.caption}>{caption}</caption>}
+        {verticalHeaders && verticalHeaders.length && <VerticalHeaderColGroup />}
+        <thead>
+          <HeaderRow>
+            {verticalHeaders && verticalHeaders.length && <td role="presentation" />}
+            {checkboxSelection && <td className={styles.checkboxHeader} />}
+            {cols.map((column) => {
+              if (column.key === indexKey && !renderIndexCol) {
```
minevala, line 276: Would this condition be more apparent as a function or constant?
```const isHiddenIndexColumn = (column.key === indexKey && !renderIndexCol)```
Or would it be possible to remove it altogether from the columns?
```const visibleColumns =  renderIndexCol ? cols : cols.filter(column => (column.key === indexKeu))```

File: packages/react/src/components/table/components/SortingHeaderCell/SortingHeaderCell.tsx [hunk 18]
```diff
@@ -0,0 +1,93 @@
...
+  IconSortDescending,
+} from '../../../../icons';
+
+export type SortingHeaderCellProps = React.ComponentPropsWithoutRef<'th'> & {
+  ariaLabelSortButtonNeutral: string;
+  ariaLabelSortButtonAscending: string;
+  ariaLabelSortButtonDescending: string;
+  colKey: string;
+  setSortingAndOrder: Function;
+  order: 'neutral' | 'asc' | 'desc';
```
minevala, line 20: Another name for neutral could be 'unset'? Neutral is ok too

packages/react/src/components/table/Table.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    rows.forEach((row) => {
+      expect(container.querySelector(`[id="hds-table-checkbox-${row.id}"]`)).toBeChecked();
+    });
+
+    userEvent.click(container.querySelector('[data-testid="hds-table-deselect-all-button"]'));
+    rows.forEach((row) => {
+      expect(container.querySelector(`[id="hds-table-checkbox-${row.id}"]`)).not.toBeChecked();
+    });
+  });
+});
```
minevala, line 135: Could there be some test for conditional rendering, for renderIndexCol, for example?

minevala (commented):
Good job 👍 
I wrote some minor comments and suggestions
Works nicely 🏅

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,345 @@
...
+        <TableBody textAlignContentRight={textAlignContentRight}>
+          {processedRows.map((row, index) => (
+            <tr key={row[indexKey]}>
+              {verticalHeaders && verticalHeaders.length && <th scope="row">{verticalHeaders[index].headerName}</th>}
+              {checkboxSelection && (
+                <td className={styles.checkboxData}>
+                  <Checkbox
+                    checked={selectedRows.includes(row[indexKey])}
+                    id={`hds-table-checkbox-${row[indexKey]}`}
+                    label=""
```
minevala, line 311: There seems to be an empty label

VilleMiekkoja: - [ ] Remove empty label

packages/react/src/components/table/components/SortingHeaderCell/SortingHeaderCell.tsx
VilleMiekkoja: Unset feels better, let's change it.

- [ ] Change neutral to unset

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const firstRenderedColumnKey = cols.find((column) => {
+    if (!renderIndexCol) {
+      return column.key !== indexKey;
+    }
+    return true;
+  }).key;
+
+  return (
+    <>
+      {(checkboxSelection || heading || (customActionButtons && customActionButtons.length > 0)) && (
```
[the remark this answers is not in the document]
VilleMiekkoja: Good idea:

- [ ] Improve readability and decrease repetition by using a constant

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,345 @@
...
+  sortingEnabled?: boolean;
+  textAlignContentRight?: boolean; // defaults to false -> text is aligned left
+  theme?: TableCustomTheme; // Custom theme styles
+  variant?: 'dark' | 'light';
+  verticalHeaders?: Array<Header>;
+  verticalLines?: boolean;
+  zebra?: boolean;
+};
+
+function processRows(rows, order, sorting, sortingEnabled, cols) {
```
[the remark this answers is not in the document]
VilleMiekkoja: No, we should use only one. I guess function expressions is more common, so maybe we go with that?

packages/react/src/components/table/Table.stories.tsx
```diff
@@ -0,0 +1,919 @@
...
+      id: 1002,
+      firstName: 'Anneli',
+      surname: 'Routa',
+      age: 50,
+      city: 'Turku',
+      profession: 'Meteorologist',
+      experience: 25,
+    },
+    {
+      id: 1002,
```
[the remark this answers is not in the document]
VilleMiekkoja: Good catch!

- [ ] Remove duplicate id

packages/react/src/components/table/Table.tsx
VilleMiekkoja: Good idea:

- [ ] use visibleColumns to increase readability and decrease repetition

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+  useEffect(() => {
+    // This tackles the case where rows have been deleted; deleted row cannot be among selected
+    const newSelectedRows = selectedRows.filter((selectedRowId) => {
+      const selectedRowExistsInRows = rows.find((row) => row[indexKey] === selectedRowId);
+      return !!selectedRowExistsInRows;
+    });
+
+    setSelectedRows(newSelectedRows);
+  }, [rows]);
```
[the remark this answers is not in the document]
VilleMiekkoja: Hmm.. setting selectedRows as a dependency seems to be creating an infinite loop, as the selectedRows are changed by this useEffect 😅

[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Solved using React.ref

packages/react/src/components/table/Table.stories.tsx
```diff
@@ -0,0 +1,919 @@
...
+  return (
+    <div style={{ maxWidth: '640px' }}>
+      <Table cols={cols} rows={rows} caption={caption} indexKey="id" renderIndexCol={false} />
+    </div>
+  );
+};
+
+// args is required for docs tab to show source code
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
+export const Dark = (args) => {
```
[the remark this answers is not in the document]
VilleMiekkoja: Yes they could. Should we go with name 'Dark' or 'Default'? The default variant is the dark variant. In Core version there is no default at the moment. It is coded so that you need to tell which variant you want to use. It could have probably been coded differently there, but it felt more clear to give them separate classnames: https://hds.hel.fi/storybook/core/?path=/info/components-table--dark

packages/react/src/components/table/Table.tsx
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Reformat properties

packages/react/src/components/table/Table.tsx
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Solved by moving selected rows state outside of component

packages/react/src/components/table/Table.stories.tsx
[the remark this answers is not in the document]
minevala: Default is ok 👍

packages/react/src/components/table/Table.module.scss
minevala: I guess we can, IIRC breakpoint-m is pretty close to this.

site/docs/components/table.mdx
```diff
@@ -38,10 +41,18 @@ HDS Tables are currently only available for HDS Core and they include base style
 - Table container is allowed to be scrollable. This is common on smaller screens where it can be impossible to fit the full table on the device screen.
   - On smaller screens, it may be beneficial to consider other presentation formats for table data. For example, a table can be broken into cards.
 
+### Interactable tables (React)
+- HDS Tables support sorting. When a column is set to be sortable, a sorting button appears after its header.
```
VilleMiekkoja, line 45: Single column cannot be set sortable; it's all columns or no sorting at all at the moment

site/docs/components/table.mdx
```diff
@@ -449,7 +526,38 @@ Vertical lines can be used on tables that have a lot of columns or a lot of simi
...
+);
+
+return (
+  <div style={{ maxWidth: '640px' }}>
+    <Table verticalLines cols={cols} rows={rows} caption={caption} indexKey="id" renderIndexCol={false} />
+  </div>
+);
+```
 
 ### Dense tables
```
VilleMiekkoja, line 562: Do we want to replace the dense also with a react version, like we have done with the other variants?

site/docs/components/table.mdx
```diff
@@ -1130,7 +1238,456 @@ If your table has a large amount of data, you can use the `dense` property to fo
...
+const caption = (
+  <span>
+    <b>Table 6</b>: Job trials (source: Helsinki Employment Services)
+  </span>
+);
+
+return (
+  <div style={{ maxWidth: '1200px', height: '600px' }}>
+    <Table
+      sortingEnabled
```
VilleMiekkoja, line 1297: Not sure if we should show sorting here, as this is only about dense tables?

site/docs/components/table.mdx
```diff
@@ -1130,7 +1238,465 @@ If your table has a large amount of data, you can use the `dense` property to fo
...
+      cols={cols}
+      rows={rows}
+      caption={caption}
+    />
+  </div>
+);
+```
+
+### Selecting rows
+HDS Table rows can be set to be selectable. This makes the first column to be a selection column. This also adds two (2) action buttons above the table; one button for selecting all rows and one button to clear selection.
```
VilleMiekkoja, line 1416: clear selection -> clear selections?

packages/react/src/components/table/Table.stories.tsx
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Combine default and dark: remove dark

site/docs/components/table.mdx
ronijaakkola: Hmm. I do not know whether this is a good thing or not. In audit, we were told to think carefully when to use sorting. Now it is easily possible that some of the columns are not easily sortable when you mostly want to sort names/number etc.

site/docs/components/table.mdx
ronijaakkola: I thought about it but the React version in Storybook was not easily transferrable to Playground. I think it does not matter if this example is in Core for now.

site/docs/components/table.mdx
ronijaakkola: - [ ] Remove sorting from dense table doc exampl

site/docs/components/table.mdx
ronijaakkola: I guess both are ok. Selection can include multiple elements.

packages/react/src/components/table/Table.module.scss
VilleMiekkoja: - [ ] Apply existing breakpoint constants

packages/react/src/components/table/Table.test.tsx
VilleMiekkoja: - [ ] Add test for conditional rendering of index column

site/docs/components/table.mdx
VilleMiekkoja: We can leave it as it is 👍

site/docs/components/table.mdx
VilleMiekkoja: True, we can leave it as it is 👍

site/docs/components/table.mdx
ronijaakkola: - [ ] Clarified how table sorting works in the table docs

site/docs/components/table.mdx
```diff
@@ -1130,7 +1238,456 @@ If your table has a large amount of data, you can use the `dense` property to fo
...
+      indexKey="Paikka-ID"
+      caption={caption}
+      dense
+    />
+  </div>
+);
+```
+
+### Sorting
+Columns can be configured to be sortable which makes the sorting button appear after the column header. The sorting icon is different depending on the data type of the column (alphabetical or numeric/other).
```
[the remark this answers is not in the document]
ronijaakkola: - [ ] Clarified how table sorting works in the table docs

laurakarhu (approved):
Looking good!

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+import { VerticalHeaderColGroup } from './components/VerticalHeaderColGroup';
+import { SortingHeaderCell } from './components/SortingHeaderCell';
+import { TableBody } from './components/TableBody';
+import { Checkbox } from '../checkbox';
+import { useTheme } from '../../hooks/useTheme';
+import { Button } from '../button';
+
+type Header = {
+  /**
+   * Boolean indicating whether column is sortable
```
minevala, line 17: Boolean indicating whether **a** column is sortable

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+  /**
+   * Key of header. Maps with the corresponding row data keys.
+   */
+  key: string;
+  /**
+   * Visible header name that is rendered.
+   */
+  headerName: string;
+  /**
+   * Sort icon type to be used in sorting. Use type string if content is string, otherwise use type other.
```
minevala, line 29: Use type string if **the** content is **a** string, otherwise use type other

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * @default 'string'
+   */
+  sortIconType?: 'string' | 'other';
+  /**
+   * Transform function for the corresponding row data. Use this to render custom content inside table cell.
```
minevala, line 34: inside **the** table

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * @default 'Järjestetty nousevaan järjestykseen'
+   */
+  ariaLabelSortButtonAscending?: string;
+  /**
+   * Aria-label for sort button in descending state.
+   * @default 'Järjestetty laskevaan järjestykseen'
+   */
+  ariaLabelSortButtonDescending?: string;
+  /**
+   * Aria-label for sort button in not ordered state.
```
minevala, line 65: button in the unordered state

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * @default ''
+   */
+  ariaLabelSortButtonUnset?: string;
+  /**
+   * Caption of the table.
+   */
+  caption?: string | React.ReactNode;
+  /**
+   * Boolean indicating whether the table has checkbox selection column to select rows.
```
minevala, line 74: has **the** checkbox

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * @default false
+   */
+  checkboxSelection?: boolean;
+  /**
+   * Text for clear selected rows button.
+   * @default 'Tyhjennä valinnat'
+   */
+  clearSelectionsText?: string;
+  /**
+   * Colums of the table header row. Defines header name, optional sort icon type and optional cell row transform function.
```
minevala, line 84: Columns

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * Table heading aria level.
+   */
+  headingAriaLevel?: number;
+  /**
+   * Table heading id. Used to name table to assistive technologies. Only applicable when heading prop is used.
+   * @default 'hds-table'
+   */
+  headingId?: string;
+  /**
+   * Column key used as unique identifier for row
```
minevala, line 114: Column key used as **a** unique identifier for **a** row

packages/react/src/components/table/Table.tsx
```diff
@@ -0,0 +1,435 @@
...
+   * Sorting order applied for initial sorting.
+   */
+  initialSortingOrder?: 'asc' | 'desc';
+  /**
+   * Boolean indicating whether index column is rendered in the table.
+   * @default true
+   */
+  renderIndexCol?: boolean;
+  /**
+   * Table rows. Array of objects where keys map with the keys of col.
```
minevala, line 131: An array

site/docs/components/table.mdx
```diff
@@ -38,10 +41,18 @@ HDS Tables are currently only available for HDS Core and they include base style
 - Table container is allowed to be scrollable. This is common on smaller screens where it can be impossible to fit the full table on the device screen.
   - On smaller screens, it may be beneficial to consider other presentation formats for table data. For example, a table can be broken into cards.
 
+### Interactable tables (React)
+- HDS Tables support sorting. When a column is set to be sortable, a sorting button appears aftet its header.
```
minevala, line 45: aftet -> after

site/docs/components/table.mdx
```diff
@@ -38,10 +41,18 @@ HDS Tables are currently only available for HDS Core and they include base style
...
+  - The sorting icon type can be changed depending on the type of data in the column.
+- Only use sorting when it is necessary or otherwise beneficial to the user. 
+- HDS Table's rows can be made selectable. In this case, HDS automatically provides actions for selecting and deselecting all rows.
+- If your table includes actions that can affect multiple rows (e.g. deleting selected rows or exporting an CSV file), these actions must be placed above the table. For more information, see [Selecting rows](#selecting-rows) and [Custom actions](#custom-table-actions) variant examples below.
```
minevala, line 49: an CSV -> a CSV

minevala (commented):
Really nice component 👍 
I commented some small typos but otherwise, it looks good and is ready for release

minevala (approved):
🔷 👍

PR #584, opened 2021-11-17
Feature/cookie consent

NikoHelle:
Users must give consents to cookies and other means user could be tracked. These consents must be given by the user before a website can store data to cookies or, for example, send data to analytics.

The purpose of this PR is to
- provide a controller for reading / writing the cookie where user's consents are stored
- provide "names" for common consents, so websites in same domain can use consents given in another website for the same purpose
- demonstrate usage of the controller and context in React applications

### Consents

There are required consents that users must approve. The are also optional consents. All consents, required and optional, must be handled by the user. Handling means approving all consents, approving only required consents or approving all required and some optional.

Cookie consent dialog is shown if
- all required consents are not approved
- any listed consent is not found in the cookie. This happens when a new consent is added after user has already handled consents before.
- user has rejected an optional consent A in service X, but that same consent A is marked as required in service Z. The dialog is shown is Service Z.

Current common consents are: 'matomo', 'tunnistamo', 'language', 'marketing', 'preferences'.  These are just placeholders. Any service using cookieConsentController can use these common consents, make up its own names or combination on both. Common consent can be required or optional.

### Core files:
**packages/react/src/components/cookieConsent/cookieConsentController.ts**
Reads and parses the consents from a cookie and also updates the cookie on the user's device. Stores current state of consents before saving by providing getters and setters. Common consent names are exported here.
This file is not a React file and can be used in any javascript project.

**packages/react/src/components/cookieConsent/CookieConsentContext.tsx**
Uses cookieConsentController and provides its functions and consent data in a React Context. It also has two callbacks which expose consent changes outside of the context:
- "onAllConsentsGiven" is called when user has saved consents. Not called if user won't change consents
- "onConsentsParsed" is called once, when context is initialized and  stored cookies are read from the device.

**packages/react/src/components/cookieConsent/CookieConsent.tsx**
React component which animates and renders the given content for the user when dialog should be shown.

**packages/react/src/components/cookieConsent/__storybook__/* **
Demo files for Storybook and developers


packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+          //  start tracking
+        }
+      }}
+      onConsentsParsed={(consents) => {
+        if (consents.matomo) {
+          //  start tracking
+        }
+      }}
+    >
+      <Demo />
```
VilleMiekkoja, line 37: I like the abstraction, but storybook story tab unfortunately then hides the example code. I wonder if it would be usefull to show it in this case?

VilleMiekkoja:
> Cookie consent dialog is shown if

> all required consents are not approved

Was it so that we need to allow user to the site even if they do not approve all required consents? If true, then would it be annoying for the user to have to always disapprove the consent when they return back to the site? Alternative would be to remember that user has not approved the required cookies, and just do not show the cookie consent. Though there needs to be a way for the user to allow the cookies within the app/page manually.

minevala (commented):
👍 I added comments. I need to revisit this after the layout is decided

packages/react/src/components/cookieConsent/buttons/Buttons.tsx
```diff
@@ -0,0 +1,46 @@
...
+        Pakolliset evästeet
+      </Button>
+      <Button
+        variant="secondary"
+        onClick={() => {
+          onClick('showDetails');
+        }}
+        data-testid="cookie-consent-read-more-button"
+      >
+        Asetukset
```
minevala, line 40: Should the settings button be hidden when the user is on the cookie settings view?

packages/react/src/components/cookieConsent/CookieConsent.test.tsx
```diff
@@ -0,0 +1,209 @@
...
+    it('and child components are rendered when consents have not been handled', () => {
+      const result = renderCookieConsent(defaultConsentData);
+      verifyElementExistsByTestId(result, dataTestIds.container);
+      verifyElementExistsByTestId(result, dataTestIds.languageSwitcher);
+      verifyElementExistsByTestId(result, dataTestIds.informationComponent);
+      verifyElementExistsByTestId(result, dataTestIds.approveAllButton);
+      verifyElementExistsByTestId(result, dataTestIds.approveRequiredButton);
+      verifyElementExistsByTestId(result, dataTestIds.readMoreTextButton);
+    });
+    it('is rendered if a required consent has not been approved. It could have been optional before', () => {
```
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Add linebreaks between test cases

packages/react/src/components/cookieConsent/__storybook__/Application.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+function Application(): React.ReactElement {
+  const { willRenderCookieConsentDialog } = useContext(CookieConsentContext);
+  return (
+    <div
+      className={classNames(styles.application, willRenderCookieConsentDialog ? styles['no-scroll'] : '')}
+      aria-hidden={willRenderCookieConsentDialog ? 'true' : 'false'}
+    >
+      <h1>This is a dummy application</h1>
+      {willRenderCookieConsentDialog && (
```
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Apply ternary

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
VilleMiekkoja: - [ ] Unravel the abstraction

packages/react/src/components/cookieConsent/content/Content.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <div
+      className={styles.content}
+      role="dialog"
+      aria-labelledby="cookie-consent-active-heading"
+      aria-describedby="cookie-consent-description"
+      id="cookie-consent-content"
+      aria-live="assertive"
+    >
+      {!showMore && <Main onClick={onAction} />}
```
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Apply ternary

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
```diff
@@ -95,11 +98,22 @@ export const Example = () => {
       onAllConsentsGiven={(consents) => {
         if (consents.matomo) {
           //  start tracking
+          // window._paq.push(['setConsentGiven']);
+          // window._paq.push(['setCookieConsentGiven']);
         }
+        document.getElementById('focused-element-after-cookie-consent-closed').focus();
```
minevala, line 104: Could the id be in some constant?

packages/react/src/components/cookieConsent/CookieConsent.module.scss
```diff
@@ -0,0 +1,278 @@
...
+.text-content {
+  padding: 0;
+}
+
+.text-content > p {
+  padding-bottom: var(--spacing-2-xl);
+}
+
+.main-content {
+  padding-top: calc(var(--common-spacing) * 2);
```
VilleMiekkoja, line 79: It might be more simple to just use hds design token variable `--spacing-l` here.

packages/react/src/components/cookieConsent/CookieConsent.module.scss
```diff
@@ -0,0 +1,278 @@
...
+    position: absolute;
+    right: 0;
+    top: calc(var(--spacing-xl) - 1 * var(--spacing-s));
+    padding: var(--spacing-s);
+  }
+}
+
+.title-with-checkbox {
+  display: flex;
+  margin-right: calc(var(--spacing-s) * 3);
```
VilleMiekkoja, line 176: We could use `--spacing-2-xl` (48px) here?

packages/react/src/components/cookieConsent/consentsInModal/ConsentsInModal.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { Modal } from '../modal/Modal';
+import { Content, Provider as CookieContextProvider } from '../CookieConsentContext';
+
+export function ConsentsInModal(props: { content: Content; cookieDomain?: string }): React.ReactElement | null {
```
VilleMiekkoja, line 6: Since this is now the public api, it would be beneficial to type the props here, and give comments about them. Then they will show up in storybook controls tab description:

image

packages/react/src/components/cookieConsent/page/Page.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useState } from 'react';
+
+import { Content, Provider as CookieContextProvider } from '../CookieConsentContext';
+import Details from '../details/Details';
+import styles from '../CookieConsent.module.scss';
+import Buttons from '../buttons/Buttons';
+import { Notification } from '../../notification/index';
+
+export function Page(props: { content: Content; cookieDomain?: string }): React.ReactElement | null {
```
VilleMiekkoja, line 9: Since this is the public component, I think we need to give it a name that has the word cookie in it. Maybe CookiePage?

packages/react/src/components/cookieConsent/consentsInModal/ConsentsInModal.tsx
VilleMiekkoja, line 6: Maybe CookieModal could be a better name?

packages/react/src/components/cookieConsent/consentGroups/ConsentGroups.tsx
```diff
@@ -0,0 +1,68 @@
...
+          data-testid={checkboxId}
+          name={checkboxId}
+          label={title}
+          aria-describedby={getConsentGroupIdenfier('description')}
+          style={checkboxStyle}
+          {...checkboxProps}
+        />
+      </div>
+      <p aria-hidden>{text}</p>
+      <VisuallyHidden id={getConsentGroupIdenfier('description')} aria-hidden>
```
VilleMiekkoja, line 54: VisuallyHidden and aria-hidden?

packages/react/src/components/cookieConsent/consentGroups/ConsentGroups.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <div className={styles['consent-group-parent']}>
+      <div className={styles['title-with-checkbox']}>
+        <Checkbox
+          id={checkboxId}
+          data-testid={checkboxId}
+          name={checkboxId}
+          label={title}
+          aria-describedby={getConsentGroupIdenfier('description')}
+          style={checkboxStyle}
+          {...checkboxProps}
```
VilleMiekkoja, line 50: I wonder if we should inline these props as well? It feels more clear that all are inlined or none are inlined?

packages/react/src/components/cookieConsent/consentGroups/ConsentGroups.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { consentGroups, isRequired } = props;
+  const cookieConsentContext = useContext(CookieConsentContext);
+  const triggerAction = useCookieConsentActions();
+  if (!consentGroups) {
+    return null;
+  }
+  const selectPercentage = cookieConsentContext.countApprovedOptional();
+  const allApproved = isRequired || selectPercentage === 1;
+  const { title, text, groupList, checkboxAriaDescription } = consentGroups;
+  const groupId = isRequired ? 'required' : 'optional';
```
VilleMiekkoja, line 26: I would move this line 26 (const groupId = ...) to line 37 to improve readability, as the groupId is only used in line 38. It's easier to read when things that are related are more close to each other.

packages/react/src/components/cookieConsent/consentGroups/ConsentGroups.tsx
```diff
@@ -0,0 +1,68 @@
...
+    onChange: isRequired ? () => undefined : () => triggerAction(checked ? 'unapproveOptional' : 'approveOptional'),
+    disabled: isRequired,
+    checked,
+    indeterminate: isRequired ? false : !Number.isInteger(selectPercentage),
+  };
+  const checkboxStyle = {
+    '--label-font-size': 'var(--fontsize-heading-m)',
+  } as React.CSSProperties;
+
+  const getConsentGroupIdenfier = (suffix: string) => `${groupId}-consents-${suffix}`;
```
VilleMiekkoja, line 38: There's probably a typo here: getConsentGroupIdenfier -> getConsentGroupIdentifier

packages/react/src/components/cookieConsent/consentGroups/ConsentGroups.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {...checkboxProps}
+        />
+      </div>
+      <p aria-hidden>{text}</p>
+      <VisuallyHidden id={getConsentGroupIdenfier('description')} aria-hidden>
+        {checkboxAriaDescription || text}
+      </VisuallyHidden>
+      <ul className={styles.list}>
+        {groupList.map((group, index) => (
+          <li key={group.title}>
```
VilleMiekkoja, line 59: Should we document somewhere, that the group.title has to be unique? Though it might be obvious 🤔

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+      counter += 1;
+      if (isApproved) {
+        approved += 1;
+      }
+    });
+    return approved / counter;
+  };
+
+  const areGroupConsentsApproved: CookieConsentContextType['areGroupConsentsApproved'] = (consentData) => {
+    const optionalConsentList = consentController.getOptional();
```
VilleMiekkoja, line 191: Does this return all optional cookie ids where user has given consent? Or those cookie ids where user is about to give consents (checkbox has been checked)?

packages/react/src/components/cookieConsent/consentGroup/ConsentGroup.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+function ConsentGroup(props: { group: ConsentGroupType; isRequired: boolean; id: string }): React.ReactElement {
+  const { group, isRequired, id } = props;
+  const { isOpen, buttonProps, contentProps } = useAccordion({
+    initiallyOpen: false,
+  });
+  const groupConsents = group.consents;
+  const cookieConsentContext = useContext(CookieConsentContext);
+  const triggerAction = useCookieConsentActions();
+  const areAllApproved = isRequired || cookieConsentContext.areGroupConsentsApproved(groupConsents);
```
VilleMiekkoja, line 24: `cookieConsentContext.areGroupConsentsApproved(groupConsents);`

Does this mean that the consents are approved already, or that user is about to approve them (checkbox is checked)?

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,245 @@
+import React, { createContext, useContext, useMemo, useState } from 'react';
+
+import create, { ConsentController, ConsentList, ConsentObject } from './cookieConsentController';
```
VilleMiekkoja, line 3: It might improve readability to rename the create function here to the same name that it has in the file cookieConsentController (createConsentController). It's little easier to find and remember when it's the same imo.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+export const Provider = ({ cookieDomain, children, content }: CookieConsentContextProps): React.ReactElement => {
+  const requiredConsents = content.requiredConsents
+    ? getConsentsFromConsentGroup(content.requiredConsents.groupList)
+    : undefined;
+  const optionalConsents = content.optionalConsents
+    ? getConsentsFromConsentGroup(content.optionalConsents.groupList)
+    : undefined;
+  const consentController = useMemo(() => create({ requiredConsents, optionalConsents, cookieDomain }), []);
+  const hasUserHandledAllConsents = () =>
+    consentController.getRequiredWithoutConsent().length === 0 && consentController.getUnhandledConsents().length === 0;
```
VilleMiekkoja, line 112: Should we combine the functions `getRequiredWithoutConsent` and `getUnhandledConsents `? They seem to only be used in this line, and reading the code here is little misleading. As we have a function named `getUnhandledConsents`, one could expect that calling it would suffice to find out whether user has handled all consents or not? Maybe we combine them under the name `getUnhandledConsents`, what do you think?

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+      }
+    }
+    return savedData;
+  };
+
+  const update: CookieConsentContextType['update'] = (key, value) => {
+    consentController.update(key, value);
+  };
+  const approveSelectedAndRequired = () => {
+    consentController.approveRequired();
```
VilleMiekkoja, line 139: Would `consentController.approveSelectedAndRequired();` describe the meaning of the function better?

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+    } else if (action === 'changeConsentGroup') {
+      consents.forEach((consent) => {
+        update(consent, value);
+      });
+    } else if (action === 'approveOptional') {
+      setOptional(true);
+    } else if (action === 'unapproveOptional') {
+      setOptional(false);
+    }
+    reRender();
```
VilleMiekkoja, line 175: Is this `reRender` necessary? There seems to be another `reRender` that is called at the end of almost each of the if clauses. For example, `approveAll` will call function `save`, and if user has handled all consents, the `save` function will call `reRender`. So reRender gets called twice, I wonder if it's always necessary?

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+  });
+
+  const [, forceUpdate] = useState<number>(0);
+  const reRender = () => {
+    forceUpdate((p) => p + 1);
+  };
+
+  const save = () => {
+    const savedData = consentController.save();
+    if (hasUserHandledAllConsents()) {
```
VilleMiekkoja, line 126: Is this if necessary? Is there a situation where we save cookies but user has not handled all consents?

packages/react/src/components/cookieConsent/page/Page.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <CookieContextProvider content={content} cookieDomain={cookieDomain}>
+      <div className={styles.page} data-testid="cookie-consent">
+        <div className={styles.content} id="cookie-consent-content">
+          <div className={styles['main-content']} data-testid="cookie-consent-information">
+            <h1>{title}</h1>
+            <p>{text}</p>
+          </div>
+          <Details />
+          {showSaveNotification && (
```
VilleMiekkoja, line 26: Think we need to tweak the styles of the save notification:

image

Maybe we need to add some spacing between the buttons and the notification? What do you think @ronijaakkola @NikoHelle

packages/react/src/components/cookieConsent/cookieConsentController.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+
+function createConsentsString(consents: ConsentObject): string {
+  if (!_isObject(consents)) {
+    return '{}';
+  }
+  return JSON.stringify(consents);
+}
+
+function createCookieController(
```
VilleMiekkoja, line 76: Should we move this function inside the file cookieController.ts? It sounds like a function that user would expect to find there. If we move it there, we would not have to deal with problems that arise from duplicate name cookieController. Im referring to lines 96 and 99, where we have had to rename cookieController to `const get = (): string => cookieControllerModule.get(COOKIE_NAME) || '';`

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,46 @@
...
+import Content from '../content/Content';
+
+export function Modal(): React.ReactElement | null {
+  const cookieConsentContext = useContext(CookieConsentContext);
+  const shouldShowModal = !cookieConsentContext.hasUserHandledAllConsents();
+  const [isModalInitiallyShown] = useState<boolean>(shouldShowModal);
+  const [popupTimerComplete, setPopupTimerComplete] = useState<boolean>(false);
+  const popupDelayInMs = 500;
+  // if hasUserHandledAllConsents() was false at first and then later true, user must have saved consents.
+  const showScreenReaderSaveNotification = isModalInitiallyShown && !shouldShowModal;
```
VilleMiekkoja, line 16: Would another alternative be to show the screenreader save notification when user clicks the save buttons, like done in the page version of the cookie consent?

packages/react/src/components/cookieConsent/consentGroupDataTable/ConsentGroupDataTable.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }, [tableHeadings]);
+
+  const rows = useMemo(() => {
+    return consents.map((consent) => {
+      return consent;
+    });
+  }, [consents]);
+
+  const theme = {
+    '--header-background-color': 'var(--color-black)',
```
VilleMiekkoja, line 28: According to design this should be `'var(--color-black-90)',`

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { settingsSaved } = useCookieConsentUiTexts();
+  useEffect(() => {
```
VilleMiekkoja, line 18: This leads to Modal and everything it contains to be rendered twice. Do we need to render it twice? We probably need it for the animation? Maybe we could only render it partially based on the `popupTimerComplete` variable value? We could disable rendering of the `Details` component as it is not shown when the cookie modal opens?

VilleMiekkoja:
Good and comprehensive unit tests, nice! 👍

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
```diff
@@ -14,6 +16,33 @@ export default {
...
 export const Example = () => {
+  const Application = () => {
+    const { willRenderCookieConsentDialog } = useContext(CookieConsentContext);
+    return (
+      <div
+        className={classNames(styles.application, willRenderCookieConsentDialog ? styles['no-scroll'] : '')}
+        aria-hidden={willRenderCookieConsentDialog ? 'true' : 'false'}
+      >
+        <h1>This is a dummy application</h1>
+        {willRenderCookieConsentDialog && (
```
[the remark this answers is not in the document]
NikoHelle: Code is outdated for this.

NikoHelle:
> Was it so that we need to allow user to the site even if they do not approve all required consents? If true, then would it be annoying for the user to have to always disapprove the consent when they return back to the site? Alternative would be to remember that user has not approved the required cookies, and just do not show the cookie consent. Though there needs to be a way for the user to allow the cookies within the app/page manually.

You cannot disapprove required consents. To modal is shown until approved.

packages/react/src/components/cookieConsent/CookieConsent.module.scss
```diff
@@ -0,0 +1,128 @@
...
+
+.container.animate-in .aligner {
+  transform: translateY(0%);
+}
+.container.animate-in .overlay {
+  opacity: 0.7;
+}
+
+.buttons {
+  padding: 20px 0;
```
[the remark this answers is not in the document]
NikoHelle: Outdated.

packages/react/src/components/cookieConsent/__storybook__/Application.tsx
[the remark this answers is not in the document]
NikoHelle: Outdated.

packages/react/src/components/cookieConsent/buttons/Buttons.tsx
NikoHelle: Outdated.

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
NikoHelle: Done.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
NikoHelle: No, because is does not do that. consentController.approveRequired() sets all required consents "true".

Selected consents are set when clicked, but those are saved here  too.

The code has changes after this and naming has improved (IMHO)

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
NikoHelle: Code have been refactored. Save just saves.

packages/react/src/components/cookieConsent/cookieConsentController.ts
NikoHelle: cookieController originally replaced universal-cookie and had only its methods. 

Good idea to add the creation there.

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: This component has no access to the buttons. The listener should be passed from here to child and from the to the buttons. That seems excessive prop drilling.

packages/react/src/components/cookieConsent/details/Details.tsx
```diff
@@ -0,0 +1,46 @@
...
+        variant="secondary"
+        onClick={() => {
+          onClick('approveRequired');
+        }}
+        data-testid="cookie-consent-approve-selections-button"
+      >
+        Hyväksy valinnat ja välttämättömät evästeet
+      </Button>
+      <p>
+        <button type="button" className={styles['plain-text-button']} onClick={() => onClick('hideDetails')}>
```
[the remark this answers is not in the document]
NikoHelle: Links should not act like buttons.

packages/react/src/components/cookieConsent/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './modal/Modal';
```
VilleMiekkoja, line 1: Remember to add here all functions and components that we wish to expose to public.

VilleMiekkoja:
Tested and noticed couple visual bugs when cookie settings from the modal are opened:
1. The background texts are visible in the side
2. The focus of 'Piilota asetukset' button seems little off:

image

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
```diff
@@ -0,0 +1,459 @@
...
+};
+
+// args is required for docs tab to show source code
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
+export const ModalVersion = (args) => {
+  const [language, setLanguage] = useState<SupportedLanguage>('fi');
+  const onLanguageChange = (newLang) => setLanguage(newLang);
+  const focusedElementAfterCloseId = 'focused-element-after-cookie-consent-closed';
+  const contentSource: ContentSource = {
+    siteName: 'Sivusto XXX',
```
VilleMiekkoja, line 25: XXX-refers to old style adult pages, maybe we think some other name for the example? 😅

packages/react/src/components/cookieConsent/componentWrapper/ComponentWrapper.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { Provider as CookieContextProvider } from '../CookieConsentContext';
+import { ContentSource } from '../content.builder';
+
+export function ComponentWrapper(props: {
```
VilleMiekkoja, line 6: Not sure if we need this ComponentWrapper abstraction?

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
NikoHelle: :) Let's wait until we have real content

NikoHelle:
Background see-through is fixed.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const [, forceUpdate] = useState<number>(0);
+  const reRender = () => {
+    forceUpdate((p) => p + 1);
+  };
+
+  const save = () => {
+    consentController.save();
+  };
+  const notifyOnAllConsentsGiven = () => {
+    if (content.onAllConsentsGiven && hasUserHandledAllConsents()) {
```
VilleMiekkoja, line 140: Can we sometimes call this so that it does not go inside this if? And if we do, how do we handle that situation?

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,262 @@
...
+      content.onAllConsentsGiven(mergeConsents());
+    }
+  };
+  const update: ConsentController['update'] = (key, value) => {
+    consentController.update(key, value);
```
VilleMiekkoja, line 145: Should we inline this to reduce jumping when reading the code? As it's only one-line I think it could be inlined as well.

VilleMiekkoja:
The focus styles of group accordion button needs some improvements:

image

packages/react/src/components/cookieConsent/__snapshots__/CookieConsent.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<CookieConsent /> spec renders the component 1`] = `<DocumentFragment />`;
```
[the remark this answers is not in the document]
NikoHelle: Snapshots are auto-generated, so I have no idea what to do to change this.

packages/react/src/components/cookieConsent/consentsInModal/ConsentsInModal.tsx
NikoHelle: List **all** object properties here? About 20 of them.

packages/react/src/components/cookieConsent/consentGroups/ConsentGroups.tsx
NikoHelle: It is documented in the docs (confluence). Most like there won't be two groups with same name in same category

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
NikoHelle: Not at the moment. "onAllConsentsGiven"-callback should never be called without all consents and it is easier to check it here rather than everywhere the `notifyOnAllConsentsGiven` is called.

minevala:
Closed as duplicate to https://github.com/City-of-Helsinki/helsinki-design-system/pull/789

PR #593, opened 2021-11-26
Fix/Duplicate checkbox id bug

VilleMiekkoja:
## Description

Fixes the bug with duplicate ids with table checkboxes. Also small improvements to table id's default names

## How to reproduce the bug

Have multiple HDS React tables on the same page, which are created from the same data rows. When this condition is met, there will be identical ids for the table checkboxes. Having duplicate ids is a violation of html in some cases, and generally not recommended at all.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1123

## How Has This Been Tested?

### DIFF
--- a/packages/react/src/components/table/Table.tsx
+++ b/packages/react/src/components/table/Table.tsx
@@ -107,9 +107,14 @@ export type TableProps = React.ComponentPropsWithoutRef<'table'> & {
   headingAriaLevel?: number;
   /**
    * Table heading id. Used to name table to assistive technologies. Only applicable when heading prop is used.
-   * @default 'hds-table'
+   * @default 'hds-table-heading-id'
    */
   headingId?: string;
+  /**
+   * Id that is passed to the native html table element.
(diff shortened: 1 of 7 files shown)


packages/react/src/components/table/Table.tsx
```diff
@@ -232,7 +237,8 @@ export const Table = ({
   dense = false,
```
minevala, line 237: Not related to this exact fix but should we have a default value for dataTestId as well?

minevala (approved):
Great fix 🏅 
Tested, fixes the documentation.
I had one small suggestion, not related to the actual fix.

packages/react/src/components/table/Table.tsx
VilleMiekkoja: - [ ] Add default value for dataTestId

minevala (approved):
Nice fix 🏅

PR #598, opened 2021-11-30
[Fix] add support to remove selectedItem with an enter and space keypress

minevala:
## Description
- Add support for removing a selected item with an enter and space keypress from multiselect and combobox
- Use role=button in Combobox and Select tags

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1107

## Motivation and Context
- This feature was reported as a bug. The behaviour might have changed in some library update. I did not find this change from the [Downshift major release](https://github.com/downshift-js/downshift/releases/tag/v6.0.0).

## How Has This Been Tested?
- locally on dev machine
- in unit test

👉 [Storybook Demo](https://city-of-helsinki.github.io/hds-demo/combobox-enter-remove/?path=/story/components-dropdowns-combobox--multiselect)

### DIFF
--- a/packages/react/src/internal/selectedItems/SelectedItems.tsx
+++ b/packages/react/src/internal/selectedItems/SelectedItems.tsx
@@ -239,7 +239,8 @@ export const SelectedItems = <OptionType,>({
                 onKeyDown: (event) => {
                   // some browsers navigate back when Backspace is pressed
                   if (event.key === 'Backspace') event.preventDefault();
-                  else if (event.key === 'Enter') {
+                  // Add support to remove an item with an enter or a space keypress
+                  else if (event.key === 'Enter' || event.key === ' ') {
                     event.preventDefault();
                     onRemove(_selectedItem);
                   }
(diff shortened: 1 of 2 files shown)


packages/react/src/internal/selectedItems/SelectedItems.tsx
```diff
@@ -239,6 +239,10 @@ export const SelectedItems = <OptionType,>({
                 onKeyDown: (event) => {
                   // some browsers navigate back when Backspace is pressed
                   if (event.key === 'Backspace') event.preventDefault();
```
VilleMiekkoja, line 241: Should we fire onRemove also when user removes selected item with backspace keypress?

ronijaakkola:
How hard would it be to make the item removable also on `Space` keypress? After discussing this with the accessibility people, the `Space` probably should be supported too.

packages/react/src/internal/selectedItems/SelectedItems.tsx
minevala: IIRC, Downshift does it by default, so we do not need to handle it.

minevala:
> How hard would it be to make the item removable also on Space keypress? After discussing this with the accessibility people, the Space probably should be supported too.
We need to add it here and ensure that space still works in the search and select.
It might need some additional changes 🤔

minevala:
> How hard would it be to make the item removable also on `Space` keypress? After discussing this with the accessibility people, the `Space` probably should be supported too.

We need to add it here and ensure that space still works in the search and select.
It might need some additional changes 🤔

VilleMiekkoja (approved):
Thanks, good fix! 👍

Issue #370, opened 2021-02-03
Tabs don't allow changing colors


laurakarhu:
Current implementation of tabs doesn't support changing colors.

PR #602, opened 2021-12-02
Feature/Enable customisation of Tabs component through theme prop

VilleMiekkoja:
## Description

Enable customisation of Tabs component through theme prop

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-614

## Closes 

https://github.com/City-of-Helsinki/helsinki-design-system/issues/370

## How Has This Been Tested?

- Automated visual loki regression tests

:point_right: [Storybook demo](https://city-of-helsinki.github.io/hds-demo/tabs-custom-theme/?path=/story/components-tabs--with-custom-theme)

image


ronijaakkola:
I think I will add an example of this to the documentation site as well. There we can explain what colours are fine to use and when the text colour should be something else than a brand colour.

packages/react/src/components/tabs/Tabs.stories.tsx
```diff
@@ -74,3 +74,38 @@ export const Small = () => (
     </TabPanel>
   </Tabs>
 );
+
+export const WithCustomTheme = () => {
+  const theme = {
+    '--tab-color': 'var(--color-brick)',
```
minevala, line 80: We could use some less bright color in the example 😎

packages/react/src/components/tabs/Tabs.tsx
```diff
@@ -6,18 +6,39 @@ import styles from './Tabs.module.scss';
...
+import { useTheme } from '../../hooks/useTheme';
+
+export interface TabsCustomTheme {
+  '--tablist-border-color'?: string;
+  '--tablist-border-size'?: string;
+  '--tab-color'?: string;
+  '--tab-font-size'?: string;
+  '--tab-height'?: string;
+  '--tab-min-width'?: string;
+  '--tab-spacing'?: string;
```
minevala, line 18: Do we need to open these properties for customization? As they might affect accessibility somehow for example.

minevala (commented):
Looks good 👍 
I had a couple of questions or comments.

packages/react/src/components/tabs/Tabs.stories.tsx
ronijaakkola: What do you mean? There are not many brand colours you can actually use in that example. There are not many brand colours that meet the `3:1` contrast requirement on UI elements.

packages/react/src/components/tabs/Tabs.tsx
ronijaakkola: One use case I can think of is that if someone uses tabs in some custom component that requires more changes to tab visuals. In this case, I would prefer that they use the HDS component even though visually modified one. These properties would allow the customisation if needed.

ronijaakkola: Yes, I think you can create some unaccessible modifications with these properties. But then again, you can do that with the colour properties alone so I don't think that is a big issue.

ronijaakkola (approved):

packages/react/src/components/tabs/Tabs.stories.tsx
minevala: I meant the red. It looks like there is an error in tabs

minevala (approved):
More flexible tabs 🏅

packages/react/src/components/tabs/Tabs.stories.tsx
VilleMiekkoja: - [ ] Improve custom theme colors; align colors with doc site

packages/react/src/components/tabs/Tabs.tsx
VilleMiekkoja: We decided to leave them as they are now

PR #605, opened 2021-12-03
Fix date input focus disappear bug

VilleMiekkoja:
## Description

Fix date input focus disappearing bug.
Bug could be reproduced by: Set max date to next month, and click next month button.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-862

## How Has This Been Tested?

- Locally on devs machine

👉 [Storybook Demo](https://city-of-helsinki.github.io/hds-demo/date-input-focus-bug-fix/iframe.html?id=components-dateinput--with-min-and-max-date&args=&viewMode=story)

image

### DIFF
--- a/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
+++ b/packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
@@ -187,6 +187,11 @@
     justify-content: flex-end;
     transform: translateY(-3px);
 
+    [aria-disabled="true"] {
+      color: rgba(16, 16, 16, 0.3);
+      cursor: not-allowed;
+    }
+
     button {
       cursor: pointer;
       padding: 0.15rem;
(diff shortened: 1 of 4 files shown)


packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
```diff
@@ -187,6 +187,11 @@
     justify-content: flex-end;
     transform: translateY(-3px);
 
+    [aria-disabled="true"] {
+      color: rgba(16, 16, 16, 0.3);
```
minevala, line 191: Suggestion: would it be possible to use some lighter variant of black?

minevala (approved):
A nice and clean fix 🏅

packages/react/src/components/dateInput/components/datePicker/DatePicker.module.scss
VilleMiekkoja: - [ ] Apply HDS black variant variable

PR #606, opened 2021-12-08
Fix/Doc site crashing locally

VilleMiekkoja:
## Description

Fix doc site crashing locally by reverting some of the previous dependency updates.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1135

## How Has This Been Tested?

- Locally on developers machine.

The change requires users to delete all previous node_modules (root, site, packages/core, packages/design-tokens, packages/react), and installing dependencies again by running command `yarn` in the root of the project

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,22 +10,22 @@
     tslib "~2.0.1"
 
 "@babel/cli@^7.5.5":
-  version "7.16.0"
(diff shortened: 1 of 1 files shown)


minevala (approved):
Tested. Now the doc site won't crash anymore after changing the source code 👍 🏅

PR #612, opened 2021-12-08
Fix/Add active state to HDS buttons to add focus outline to safari on click

VilleMiekkoja:
## Description

Safari does not focus button on press. Therefore the focus outline that is shown in other browser did not appear on Safari. This PR adds the same focus outline for css active state, which is triggered also in Safari when button is pressed.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1126

## How Has This Been Tested?

:point_right: [Storybook Demo](https://city-of-helsinki.github.io/hds-demo/button-active-styles/iframe.html?id=components-button--primary&args=&viewMode=story)

image

### DIFF
--- a/packages/core/src/components/button/button.css
+++ b/packages/core/src/components/button/button.css
@@ -50,7 +50,17 @@
   outline: none;
 }
 
+.hds-button:active {
+  background-color: var(--background-color-focus, transparent);
+  color: var(--color-focus);
+  outline: none;
+}
+
 .hds-button:focus:hover {
+  background-color: var(--background-color-hover-focus, transparent);
+}
+
+.hds-button:active:hover {
   background-color: var(--background-color-hover-focus, transparent);
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/components/button/button.css
```diff
@@ -50,10 +50,20 @@
...
+  color: var(--color-focus);
+  outline: none;
+}
+
 .hds-button:focus:hover {
   background-color: var(--background-color-hover-focus, transparent);
 }
 
+.hds-button:active:hover {
+  background-color: var(--background-color-hover-focus, transparent);
```
minevala, line 64: Could the active state be different from the hover state? Then the user could see the click or touch

minevala (commented):
👍 I added a little styling question

minevala (approved):
A solid fix 🏅

PR #614, opened 2021-12-09
Replace custom lfs cache with OS action

minevala:
## Description
- Replace custom lfs-cache implementation with nschloe/action-cached-lfs-checkout


VilleMiekkoja (approved):
Nice and clean, let's hope it works 🤞

VilleMiekkoja (approved):
👍

PR #615, opened 2021-12-09
Feature/Add placeholder support for search input

VilleMiekkoja:
## Description

Add placeholder support for search input

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1110

## How Has This Been Tested?

- Locally on devs machine
- Visual regression loki tests

👉 [Storybook demo](https://city-of-helsinki.github.io/hds-demo/search-input-placeholder/?path=/story/components-searchinput--with-placeholder)

image


packages/react/src/components/searchInput/SearchInput.stories.tsx
```diff
@@ -214,3 +214,15 @@ WithSuggestionsSpinner.args = {
   label: 'Search for a fruit',
   helperText: 'Assistive text',
 };
+
+export const WithPlaceholder = (args) => {
```
ronijaakkola, line 218: Instead of adding a new story to demo this, I think we should add the placeholder text to other stories. E.g. see how this is demoed for the TextInput for example.

ronijaakkola (changes requested):

ronijaakkola:
Note that the SearchInput is also available in the Core package: https://github.com/City-of-Helsinki/helsinki-design-system/tree/feature/search-input-placeholder/packages/core/src/components/search-input

We probably should add the example of this to the Core Storybook too.

ronijaakkola:
Added the placeholder text to the Core Storybook example too. Also added these to the documentation Playground examples.

packages/react/src/components/searchInput/SearchInput.stories.tsx
VilleMiekkoja: Good idea! 👍

packages/core/src/components/search-input/search-input.stories.js
```diff
@@ -14,6 +14,7 @@ export const Default = () => `
         id="example"
         class="hds-search-input__input"
         type="text"
+        placeholder="Placeholder text"
```
ronijaakkola, line 17: > "Placeholder text"

To be in line with other stories, this should only read "Placeholder".

ronijaakkola (changes requested):

packages/core/src/components/search-input/search-input.stories.js
VilleMiekkoja: - [ ] Placeholder text -> Placeholder

minevala (approved):
Good improvement 🏅

PR #616, opened 2021-12-09
Restrict node version because of node-sass

minevala:
## Description
- Add supported node range (v12 - v14) into package.json

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1058

## Motivation and Context
- HDS uses version 4.14 of node-sass, and it only works in node versions from v12 to v14 
- [Node-sass's Node version support policy](https://www.npmjs.com/package/node-sass)


laurakarhu (approved):
Thanks!

VilleMiekkoja (approved):
Nice 👍

PR #618, opened 2021-12-15
Feature/koros rotate

minevala:
## Description
- Add ability to rotate Koros

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1131

## Motivation and Context
- A rotated Koros can be used in Hero or similar components/elements

## How Has This Been Tested?
- Locally on dev machine
- Loki tests

👉  [Core Storybook Demo](https://city-of-helsinki.github.io/hds-demo/koros-rotate-core/?path=/story/components-koros--rotated-basic-45-deg)

👉 [React Storybook Demo](https://city-of-helsinki.github.io/hds-demo/koros-rotate/?path=/story/components-koros--rotated)

Screenshot 2021-12-15 at 12 24 29


ronijaakkola:
Looking good! This requires a new section to documentation as well. I will add that soon.

ronijaakkola:
Documentation added:
- A new variation "Angled" to demo angled koros
- Added Core code examples to the documentation page (they do not fully match the React counterpart but they are good enough)

packages/react/src/components/koros/Koros.stories.tsx
```diff
@@ -34,24 +34,73 @@ export const Storm = () => <Koros type="storm" />;
...
-    <Koros type="storm" flipHorizontal />
+    <Koros type="storm" flipHorizontal={args.flipHorizontal} />
   </>
 );
 
+Flipped.args = {
+  flipHorizontal: true,
+};
+
+export const Rotated = (args) => <Koros type={args.type} flipHorizontal={args.flipHorizontal} rotate={args.rotate} />;
```
ronijaakkola, line 59: I would probably either remove, rename or change either the Rotated or the RotatedKorosExample.They are very similar and it can be confusing why there are two examples of the same thing. 

RotatedKorosExample looks very good and it is a practical example of using Koros in a good way. Too bad it does not demo the possibility to rotate the koro shape with the property. 

Maybe renaming here would be the best solution? For example, Rotated and Rotated in a container or something?

ronijaakkola (changes requested):

packages/react/src/components/koros/Koros.stories.tsx
minevala: - [x] rename koros story

VilleMiekkoja (approved):
Looking good 👍

laurakarhu (approved):
Thanks, keep spinning the koros :)

ronijaakkola (approved):
Looks good!

PR #624, opened 2022-01-07
[hotfix] Remove confusing inline transform rotate from example

minevala:
## Description
- Remove confusing inline rotate style from docs example

## Motivation and Context
- Koros should be rotated only with the rotate property

## How Has This Been Tested?
- locally on dev machine


VilleMiekkoja (approved):
Nice 👍

PR #627, opened 2022-01-12
Add support to use Tabs component with namespace convention

minevala:
## Description
- Add support for component namespace into Tabs component

## Motivation and Context
- Problem: Tabs components do not render on MDX pages
- At the moment, Tabs-component's subsequent components are independent. But they can not be used without any of the parent components separately. The parent components modify the properties of child components (add indexes, etc.) to render them correctly. So the Tabs component group should be considered as a single component.

- MDX engine processes independent components into MDX components separately. This prevents parent components from finding the child components as the parents are using componentName to find the correct component. The componentName is lost in the MDX process. Now when the whole group is using the same namespace, the MDX considers the group as a single component and leaves child components intact. 

## How Has This Been Tested?
- Locally on dev machine

### DIFF
--- a/packages/react/src/components/tabs/Tabs.tsx
+++ b/packages/react/src/components/tabs/Tabs.tsx
@@ -75,6 +75,8 @@ export const Tabs = ({ children, small = false, theme }: TabsProps) => {
   );
 };
 
+// Using the Tabs component and its child components in the same namespace ensures that the group is considered as a single component.
+// Tabs components do some child tab component processing and this might break in some preprocessed environments if the components were separate components, for example in mdx environment.
 Tabs.TabList = TabList;
 Tabs.TabPanel = TabPanel;
 Tabs.Tab = Tab;
--- a/packages/react/src/components/tabs/Tabs.tsx
+++ b/packages/react/src/components/tabs/Tabs.tsx
@@ -7,6 +7,9 @@ import classNames from '../../utils/classNames';
 import { TabsContext } from './TabsContext';
 import { FCWithName } from '../../common/types';
 import { useTheme } from '../../hooks/useTheme';
+import { TabList } from './TabList';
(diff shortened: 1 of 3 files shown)


packages/react/src/components/tabs/Tabs.tsx
```diff
@@ -71,3 +74,7 @@ export const Tabs = ({ children, small = false, theme }: TabsProps) => {
     </TabsContext.Provider>
   );
 };
+
+Tabs.TabList = TabList;
```
ronijaakkola, line 78: I guess the burning question is that will is break anything in projects that currently use the Tabs component?

minevala: No, I did not change the exports, they can still use them as separate components. It is not the preferred way though. I'll fix the examples as well

minevala: I'll fix the examples

ronijaakkola: Maybe add a comment here to explain why this was done? Although might be hard if it is not completely certain why this fixes the issue.

minevala: I just updated the description. Comment does it make sense 😁

ronijaakkola: I mean adding a comment to before this actual code. I don't know how easy it would be to find this PR afterwards when someone starts wondering why this was done.

minevala: Yes. I could add that.

minevala: - [x] add code comment

VilleMiekkoja (approved):
Good work! 🎉

PR #629, opened 2022-01-14
Feature/tabs new docs

minevala:
## Description
- Split buttons.mdx into smaller page sections
- Use tabs to combine page sections on index page 

Depends on: #627 

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-915

## Motivation and Context
- By using Tabs the page size stays smaller and does not feel that overwhelming for the user.

## How Has This Been Tested?
- Locally on dev machine

Screenshot 2022-01-18 at 17 44 11

[👉 Storybook Demo](https://city-of-helsinki.github.io/hds-demo/docsite-playground/elements/components/buttons/)

### DIFF
--- a/new-site/src/components/LeadParagraph.js
+++ b/new-site/src/components/LeadParagraph.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const LeadParagraph = ({
+  color = 'var(--color-black-90)',
+  size = 'var(--fontsize-body-xl)',
+  style = {},
+  children,
+}) => (
+  <p
+    style={{
+      fontSize: size,
+      color,
+      maxWidth: 600,
+      ...style,
+    }}
+  >
+    {children}
+  </p>
+);
+
+LeadParagraph.propTypes = {
+  color: PropTypes.string,
+  size: PropTypes.string,
+};
+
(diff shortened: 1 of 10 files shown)


ronijaakkola (approved):

VilleMiekkoja:
Not sure if related to this PR, but the Acccessibility tab seems to contain more guidance than just a11y? I wonder if there would be a better name for it to cover all what it contains 🤔

VilleMiekkoja:
Would be possible/worth it to set height of Tabs always according to its most highest TabPanel? With the current implementation the content of the page jumps/scrolls when tab is switched due to different heights:

￼

minevala:
> Would it be possible/worth it to set height of Tabs always according to its most highest TabPanel? With the current implementation the content of the page jumps/scrolls when tab is switched due to different heights:
> 
>  1ECA45CE-B94D-4BBF-9086-AAE35DEF6D21.mov

It feels that it should be fixed in Tabs component 🤔

minevala:
> Not sure if related to this PR, but the Acccessibility tab seems to contain more guidance than just a11y? I wonder if there would be a better name for it to cover all what it contains 🤔

The content is more of placeholder things. I copied it from current documentation

PR #652, opened 2022-02-17
Feature/clearable in single select

### LINKED ISSUE #358: `clearable` prop nonfunctional in default Select component
## Describe the bug
For Select component (https://hds.hel.fi/storybook/react/?path=/story/components-dropdowns-select--default), the `clearable` option seems to apply only for multiselect. Changing the cleared prop for a standard select component does not have any effect.

## How to reproduce
1. Go to https://hds.hel.fi/storybook/react/?path=/story/components-dropdowns-select--default
2. Click on 'Controls' > 'Clearable' > 'true'
3. Select an option in the demo select component

## Device information
**Desktop (please complete the following information)**:
>  - OS: macOs
> - Browser Chrome
>  - Version 87

VilleMiekkoja:
## Description

Adds clear button to Select and Combobox in single select mode.

Bonus: Focus Select/Combobox after clear button is pressed

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-827
https://github.com/City-of-Helsinki/helsinki-design-system/issues/358

Closes #358 

## How Has This Been Tested?

image

## Demo:

https://city-of-helsinki.github.io/hds-demo/clear-button-for-single-select/?path=/story/components-dropdowns-combobox--with-clear-button

### DIFF
--- a/packages/react/src/components/dropdown/combobox/Combobox.tsx
+++ b/packages/react/src/components/dropdown/combobox/Combobox.tsx
@@ -9,11 +9,10 @@ import { useVirtual } from 'react-virtual';
 import 'hds-core';
 
 import styles from './Combobox.module.scss';
-import selectStyles from '../select/Select.module.scss';
 import { FieldLabel } from '../../../internal/field-label/FieldLabel';
 import classNames from '../../../utils/classNames';
-import { IconAlertCircleFill, IconAngleDown, IconCrossCircle } from '../../../icons';
-import { SelectedItems } from '../../../internal/selectedItems/SelectedItems';
+import { IconAlertCircleFill, IconAngleDown } from '../../../icons';
+import { ClearButton, SelectedItems } from '../../../internal/selectedItems/SelectedItems';
 import { multiSelectReducer, onMultiSelectStateChange, SelectCustomTheme, SelectProps } from '../select';
 import { DROPDOWN_MENU_ITEM_HEIGHT, getIsInSelectedOptions } from '../dropdownUtils';
 import { DropdownMenu } from '../../../internal/dropdownMenu/DropdownMenu';
@@ -510,17 +509,14 @@ export const Combobox = <OptionType,>(props: ComboboxProps<OptionType>) => {
           <IconAngleDown className={styles.angleIcon} aria-hidden />
         </button>
         {showClearButtonForSingleSelect && (
-          <button
-            type="button"
-            className={classNames(selectStyles.clearButton, !showToggleButton && selectStyles.noToggle)}
-            onClick={() => {
+          <ClearButton
+            toggleButtonHidden={!showToggleButton}
+            onClear={() => {
               resetCombobox();
               toggleButtonRef.current.focus();
             }}
-            aria-label={props.clearButtonAriaLabel}
-          >
-            <IconCrossCircle />
-          </button>
+            clearButtonAriaLabel={props.clearButtonAriaLabel}
+          />
         )}
         {/* MENU */}
         <DropdownMenu<OptionType>
--- a/packages/react/src/components/dropdown/combobox/Combobox.tsx
+++ b/packages/react/src/components/dropdown/combobox/Combobox.tsx
@@ -9,9 +9,10 @@ import { useVirtual } from 'react-virtual';
 import 'hds-core';
 
 import styles from './Combobox.module.scss';
+import selectStyles from '../select/Select.module.scss';
 import { FieldLabel } from '../../../internal/field-label/FieldLabel';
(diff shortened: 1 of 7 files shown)


packages/react/src/components/dropdown/combobox/Combobox.tsx
```diff
@@ -500,6 +505,16 @@ export const Combobox = <OptionType,>(props: ComboboxProps<OptionType>) => {
         >
           <IconAngleDown className={styles.angleIcon} aria-hidden />
         </button>
+        {showClearButtonForSingleSelect && (
+          <button
+            type="button"
+            className={classNames(selectStyles.clearButton, !showToggleButton && selectStyles.noToggle)}
+            onClick={() => resetCombobox()}
```
ronijaakkola, line 512: I'm wondering where the focus should move after the clear button is pressed. Now the focus disappears. Maybe need to ask this from the accessibility team?

VilleMiekkoja: In multiselect the focus disappeared, so I went with the existing behaviour.

packages/react/src/components/dropdown/combobox/Combobox.tsx
```diff
@@ -376,6 +380,7 @@ export const Combobox = <OptionType,>(props: ComboboxProps<OptionType>) => {
         !showToggleButton && styles.noToggle,
         hasFocus && selectedItems.length > 0 && styles.adjustSpacing,
         props.icon && props.multiselect && styles.inputWithIcon,
+        clearable && !props.multiselect && selectedItem && styles.withClearButton,
```
minevala, line 383: The three first statements are the same as showClearButtonForSingleSelect

packages/react/src/components/dropdown/combobox/Combobox.tsx
```diff
@@ -83,7 +84,7 @@ export const Combobox = <OptionType,>(props: ComboboxProps<OptionType>) => {
     catchEscapeKey,
     circularNavigation = false,
     className,
-    clearable = true,
+    clearable = props.multiselect,
```
minevala, line 87: Does this mean that the library user can not hide the clear button in the multiselect mode anymore?

packages/react/src/components/dropdown/select/Select.tsx
```diff
@@ -526,6 +544,16 @@ export const Select = <OptionType,>(props: SelectProps<OptionType>) => {
           {getButtonLabel()}
           <IconAngleDown className={styles.angleIcon} aria-hidden />
         </button>
+        {showClearButtonForSingleSelect && (
+          <button
```
minevala, line 548: Could we use the SelectedItems' clear button instead of duplicate button?

packages/react/src/components/dropdown/combobox/Combobox.tsx
VilleMiekkoja: No, this is the default value. If user does not provide clearable prop, then it will use the default value.

packages/react/src/components/dropdown/combobox/Combobox.tsx
VilleMiekkoja: - [ ] Focus Combobox/Select after clear button press

packages/react/src/components/dropdown/combobox/Combobox.tsx
VilleMiekkoja: - [ ] Use showClearButtonForSingleSelect

packages/react/src/components/dropdown/select/Select.tsx
VilleMiekkoja: It seemed to differ a bit. Maybe we could use the same css class atleast?

packages/react/src/components/dropdown/combobox/Combobox.tsx
minevala: Ok 👍  
It might difficult to hide clearbutton in multiselect because of some linter rules 🤔

packages/react/src/components/dropdown/select/Select.tsx
minevala: It is pretty close at least. It would be easier to maintain if we did not have two different buttons. But if it is difficult, we could have one button for both cases.

VilleMiekkoja: - [ ] Refactor to use only clear button from SelectedItems

minevala (approved):
Nice improvement 👍

PR #653, opened 2022-02-18
Add dense variant for Koros

VilleMiekkoja:
## Description

Adds dense variant of Koros

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1143

## Motivation and Context

In the new guidelines, there is a recommendation for larger and smaller koro variants.

## How Has This Been Tested?

image

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/koros-dense-variant/iframe.html?id=components-koros--dense&args=&viewMode=story)


packages/core/src/components/koros/koros.stories.js
```diff
@@ -20,6 +20,22 @@ export const Basic = () => `
   </div>
 `;
 
+export const Dense = () => `
```
minevala, line 23: Should this be dense basic? Technically, we should have all the dense type examples, but I think we can go with only this for now.

packages/react/src/components/koros/Koros.stories.tsx
```diff
@@ -24,6 +24,8 @@ export default {
 
 export const Basic = () => <Koros />;
 
+export const Dense = () => <Koros dense />;
+
 export const Beat = () => <Koros type="beat" />;
 
 export const Pulse = () => <Koros type="pulse" />;
```
minevala, line 31: I think dense property should be a property in the Playground story. It would help library users.

ronijaakkola:
Is there a way to test this with all Koro shapes in Storybook? Currently, there is only a way to test it with the basic shape. Maybe add all koro shapes to the Dense story?

ronijaakkola:
Also, this requires an addition to the documentation. I will add it.

packages/core/src/components/koros/koros.stories.js
ronijaakkola: But aren't the Core Koros purely just SVGs? So the developer is supposed to just copy these? We need to provide all examples if this is the case.

minevala: Yes. Then we probably should have all the dense variants.

VilleMiekkoja: - [ ] Add dense examples of all types

packages/react/src/components/koros/Koros.stories.tsx
VilleMiekkoja: Hmm.. It will probably conflict with the library update pull request, so maybe we do it later?

minevala: Could we first merge the library update?

minevala: Or we can merge this as well, and do it afterwards. It is only couple of lines 👍

minevala (approved):
Nice new Koros.
Hopefully we remember to add dense arg into Playground afterwards

PR #656, opened 2022-02-22
Revert url-parse bump because of conflicts

VilleMiekkoja:
## Description

Revert url-parse bump because of conflicts


minevala (approved):
Nice revert

PR #662, opened 2022-02-24
Fix core storybook docs tab rendering problems for icons

VilleMiekkoja:
## Description

Fix core storybook docs tab rendering problems for icons

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1204

### DIFF
--- a/packages/core/src/icons/icons.stories.js
+++ b/packages/core/src/icons/icons.stories.js
@@ -16,7 +16,7 @@ req.keys().forEach((fileName) => {
       return acc;
     }, '');
 
-    return `<span><style>.hds-icon {margin-right: var(--spacing-m);}</style>${all}</span>`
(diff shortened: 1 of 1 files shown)


packages/core/src/icons/icons.stories.js
```diff
@@ -2,20 +2,22 @@ import path from 'path';
 
 import './all.css';
 import './icon.css';
+import { storiesOf } from "@storybook/html";
```
minevala, line 5: Suggestion: should this be before all the other imports?

minevala (approved):
Nice fix 👍

packages/core/src/icons/icons.stories.js
VilleMiekkoja: - [ ] Move import to before other imports

PR #667, opened 2022-02-25
Fix/imports and markdown

minevala:
## Description
- Fix component pages examples markdown and imports

## Motivation and Context
- Get rid of console.errors 
- Fix broken examples 
- Add examples with component states

## How Has This Been Tested?
- Locally on dev machine

[Demo](https://city-of-helsinki.github.io/hds-demo/docsite-fixes/elements/components/accordion)

### DIFF
--- a/new-site/src/docs/elements/components/card/code.mdx
+++ b/new-site/src/docs/elements/components/card/code.mdx
@@ -3,18 +3,15 @@ slug: '/elements/components/card/code'
 title: 'Card - Code'
 ---
 
-import { Link } from 'hds-react';
-
-import { Card, Button } from 'hds-react';
-import { IconAngleDown, IconAngleUp, IconCheckCircleFill, IconCrossCircle } from 'hds-react';
-
+import { Button, Card, IconAngleDown, IconAngleUp, IconCheckCircleFill, IconCrossCircle, Link } from 'hds-react';
 import Playground from '../../../../components/Playground';
 
 ## Code
 
-### Code examples 
+### Code examples
 
 #### Empty
+
 <Playground>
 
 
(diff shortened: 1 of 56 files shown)


new-site/src/docs/elements/components/checkbox/code.mdx
```diff
@@ -3,21 +3,17 @@ slug: '/elements/components/checkbox/code'
...
-import { Link } from 'hds-react';
-
-import { Card, SelectionGroup } from 'hds-react';
-import { IconAngleDown, IconAngleUp, IconCheckCircleFill, IconCrossCircle } from 'hds-react';
+import { IconAngleDown, IconAngleUp, IconCheckCircleFill, IconCrossCircle, Link, SelectionGroup } from 'hds-react';
 
 import Playground from '../../../../components/Playground';
 
-## Code
+## Codes
```
VilleMiekkoja, line 10: This seems to be just Code in other component pages. Do we want to just repeat the tab name here?

new-site/src/docs/elements/components/date-input/customisation.mdx
```diff
@@ -19,13 +19,18 @@ You can use the `style` property to customise the component. See all available t
...
 
-
 ```jsx
-
-```
-
-```html
-
+<div style={{maxWidth: '400px'}}>
+  <DateInput
```
VilleMiekkoja, line 24: This seems be the same as in Code tab? We probably need to explain better above how date input can be customised, and then use those props here.

minevala: Yes. I tried to fix the customisation examples and add some working placeholders. But I guess some components might not need customisation tab at all.

new-site/src/docs/elements/components/checkbox/code.mdx
minevala: 👍  I'll fix it back
- [x] revert heading change

new-site/src/docs/elements/components/file-input/usage.mdx
```diff
@@ -8,36 +8,49 @@ import { FileInput } from 'hds-react';
...
-<FileInput
-  multiple
-  dragAndDrop
-  id="file-input"
-  label="Drag and drop files here"
-  language="en"
-  accept=".png,.jpg"
-/>
+
+export const FileInputExample = () => {
```
VilleMiekkoja, line 12: Do we need the export?

minevala: Yes, if we want to use the same block in the markdown.
It seems to be the only way to use stateful components inline in markdown.

new-site/src/docs/elements/components/card/code.mdx
```diff
@@ -26,29 +23,39 @@ import Playground from '../../../../components/Playground';
 ```
 
 ```html
-<div>
-  <div class="hds-card"></div>
-  <div class="hds-card hds-card--border"></div>
-</div>
+  <div class="hds-card">
```
VilleMiekkoja, line 26: There's some problems with rendering of these cards:

image

new-site/src/docs/elements/components/card/code.mdx
```diff
@@ -57,40 +64,52 @@ import Playground from '../../../../components/Playground';
   <div class="hds-card">
     <div class="hds-card__body">
       <div class="hds-card__heading" role="heading" aria-level="2">Card</div>
-      <div class="hds-card__text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</div>
+      <div class="hds-card__text">
```
VilleMiekkoja, line 67: The core rendering differs from React. Maybe some core styles are missing? Rebasing to master might help.

image

new-site/src/docs/elements/components/card/code.mdx
VilleMiekkoja, line 26: Also we should consider rendering the empty variant on different background color than white, so it would show something other than just white everywhere.

new-site/src/docs/elements/components/card/code.mdx
```diff
@@ -99,7 +118,10 @@ import Playground from '../../../../components/Playground';
   <div class="hds-card">
```
VilleMiekkoja, line 118: This has also the same rendering issues mentioned before.

new-site/src/docs/elements/components/card/code.mdx
minevala: - [ ] add wrapper around elements

minevala: I think we have similar examples in our current documentation

minevala: I think we could go through the content separately.

new-site/src/docs/elements/components/card/code.mdx
minevala: - [ ] import core component styles

new-site/src/docs/elements/components/card/code.mdx
minevala: It seems to work now 🤔

PR #674, opened 2022-03-07
Feature/docs component preview

minevala:
## Description
- Add a component to frame component examples

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1195

## Motivation and Context
- A bordered element gives more structure on the page. Previously it was difficult to find where the example started and ended
 
## How Has This Been Tested?
- Locally on dev machine

👉 [Accordion Demo](https://city-of-helsinki.github.io/hds-demo/docsite-component-preview/elements/components/accordion)
👉 [Button Demo](https://city-of-helsinki.github.io/hds-demo/docsite-component-preview/elements/components/buttons)

### DIFF
--- a/new-site/src/components/Playground.js
+++ b/new-site/src/components/Playground.js
@@ -22,6 +22,14 @@ Playground.propTypes = {
 
 export default Playground;
 
+const playgroundPreviewClassName = 'playground-block-preview';
+
+export const PlaygroundPreview = ({ children, ...props }) => (
+  <div {...props} className={playgroundPreviewClassName}>
+    {children}
+  </div>
+);
+
 const clearSelection = () => {
   if (window.getSelection) {
     window.getSelection().removeAllRanges();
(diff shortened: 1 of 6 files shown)


VilleMiekkoja:
Do you think it would be better if the border color of the playground preview component was grayed out? Like in the old documentation site there was these kind of examples:

image

The current border is visually pretty strong. Though maybe it is better in terms of a11y? I like the gray one more, as the border does not take too much focus

minevala:
> Do you think it would be better if the border color of the playground preview component was grayed out? Like in the old documentation site there was these kind of examples:

- [ ] tweak colors a bit

> Could this have a bit less padding on smaller screens? Now it has 32px (--spacing-l) across the board so maybe use 24px (--spacing-m) on tables sizes and 16px (--spacing-s) on the smallest screens.

- [ ] tweak paddings on smaller spaces

VilleMiekkoja (approved):

ronijaakkola (approved):

PR #677, opened 2022-03-08
Use correct syntax for providing fallback colors for table colors

minevala:
## Description
- Fix CSS syntax for fallback color
- https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#custom_property_fallback_values

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1214

## Motivation and Context
- Core tables do not have background color if it is not set. The header cells are unreadable

## How Has This Been Tested?
- Locally on dev machine

### DIFF
--- a/packages/core/src/components/table/table.css
+++ b/packages/core/src/components/table/table.css
@@ -46,7 +46,7 @@
 
 .hds-table--dark {
   /* TODO: --background-color is deprecated. Remove when possible. */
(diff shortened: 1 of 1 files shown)


VilleMiekkoja (approved):
Good catch and fix! 👍

PR #690, opened 2022-03-24
Fix/components paths

minevala:
## Description
- Change component index-path from /component-name/usage to /component-name

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1194

## Motivation and Context
- the site is easier to navigate when there are clear index paths to component pages

## How Has This Been Tested?
- locally on dev machine 

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/docsite-all-component-pages/elements/components/accordion)


ronijaakkola (approved):
Works well! Hopefully, this does not break too bad when we add the new HDS structure. Components will end up on their own page so the path will most likely be something like "/components/button" instead of "/elements/components/button". 🤞

PR #700, opened 2022-04-05
Add initial scripts to run analytics

minevala:
## Description
- Add script to collect hds-react usage statistics

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1228

## Motivation and Context
- Statistics are used in planning the future of the HDS

## How Has This Been Tested?
- locally on dev machine

### DIFF
--- a/analytics/README.md
+++ b/analytics/README.md
@@ -1,0 +1,26 @@
+# Helsinki Design System Analytics script
+
+A node script to collect the HDS-React package's usage statistics from City of Helsinki Github repositories.
+
+The script does the next steps:
+1. Searches all City of Helsinki TypeScript Github repositories. It filters out the Helsinki Design System repository
+2. Clones all the found repositories into the local folder
+3. Runs React-scanner for repositories to generate reports of HDS-React usage
+4. Removes the local temporary repository folder
+
+***Notice! Be careful with the script because it can make lots of Git clones to repositories. 
+It will strain the network and might exceed some Github rate limits. Running script requires some disk space and memory as well.***
+
+## Getting started
+
+### Prerequisites
+- [Node](https://nodejs.org/en/)
+- [Github cli](https://cli.github.com/)
+- [React-scanner](https://github.com/moroshko/react-scanner) as globally or locally installed
+
+### Running the script
+To collect analytics from a subset of found repositories (for testing purposes), run:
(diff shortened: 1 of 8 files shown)


analytics/README.md
```diff
@@ -0,0 +1,26 @@
+# Helsinki Design System Analytics script
+
+A node script to collect the HDS-React package's usage statistics from City of Helsinki Github repositories.
+
+The script does the next steps:
+1. Searches all City of Helsinki TypeScript Github repositories. It filters out the Helsinki Design System repository
```
VilleMiekkoja, line 6: Should we also search for pure JS repositories, or are they so rare that it's not worth it?

analytics/helpers/clone-repos.js
```diff
@@ -0,0 +1,28 @@
...
+        console.log(`Repo cloned, name: ${name}`);
+        return name;
+      })
+      .catch(() => {
+        console.log(`Repo clone failed, name: ${name}`);
+      }),
+  );
+
+  try {
+    return Promise.all(clonePromises);
```
VilleMiekkoja, line 22: I wonder if we should throttle/debounce these promises so that they are not all send at the same time? It might create too many concurrent network requests which could lead to problems? There's couple ways throttling could be done:

https://www.npmjs.com/package/promise-throttle
https://lodash.com/docs/4.17.15#debounce
https://stackoverflow.com/a/67271450/4662347

analytics/README.md
minevala: I did not want to include those because the previous statistics did not

analytics/helpers/clone-repos.js
minevala: Hopefully we have some library already installed which can do this. I would not want to add a separate package json for this yet 😁

VilleMiekkoja: With lodash you can import the specific library for debounce, it should not be very big I think. Or then we could use this solution which does not use external library: https://stackoverflow.com/a/67271450/4662347

minevala: - [x] add throttling to repo cloning

VilleMiekkoja (approved):
Well done! It's great to have these analytic scripts in version control

PR #707, opened 2022-04-11
Change/typography

VilleMiekkoja:
## Description

Typography updates to HDS headings. 

Breaking changes in the following components:

- Accordion
- Card
- Dialog
- ErrorSummary
- Footer
- Linkbox
- Navigation
- Notification
- Stepper
- Table
- Tooltip

Breaking changes in the following design-tokens:

- fontsize-heading-xl
- fontsize-heading-l

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-617

Updates HDS heading to match with hel.fi headings

## How Has This Been Tested?

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/typography-update/?path=/story/components-accordion--default)


packages/core/src/components/table/table.css
```diff
@@ -17,8 +17,10 @@
   border-bottom: 1px solid var(--color-black-20);
   white-space: nowrap;
   text-align: left;
-  font-weight: 500;
+  font-weight: 700;
```
minevala, line 20: The font feels almost too bold. In other places the font got thinner but in here it is bolder?

packages/core/src/components/card/card.css
```diff
@@ -34,8 +34,9 @@
 /* HEADING */
 .hds-card__heading {
   font-size: var(--fontsize-heading-m);
-  font-weight: bold;
-  line-height: var(--lineheight-m);
+  letter-spacing: -0.2px;
```
minevala, line 37: This seems same as the heading-m helper class. Is it possible to combine these?

packages/react/src/components/linkbox/Linkbox.module.scss
```diff
@@ -77,16 +77,25 @@
...
   font-size: var(--fontsize-heading-s);
+  font-weight: 700;
+  letter-spacing: 0.2px;
+  line-height: 1.4;
 }
 
 .headingMedium {
   padding-bottom: 3px;
   font-size: var(--fontsize-heading-m);
+  font-weight: 500;
```
minevala, line 88: This seems same as the heading-m helper class. Is it possible to combine these?

packages/react/src/examples/forms/validation.scss
```diff
@@ -4,13 +4,17 @@
 
   &__main-title {
     font-size: var(--fontsize-heading-xl);
-    line-height: var(--lineheight-m);
+    letter-spacing: -1.2px;
```
minevala, line 7: It might be better to use helper-classes here as an example

minevala (commented):
I think the change improves the readability a bit 👍 
A few comments

packages/core/src/components/table/table.css
VilleMiekkoja: Yeah, the S, XS and XXS got bolder (700) in this update: https://helsinkisolutionoffice.atlassian.net/browse/HDS-809. Since the heading in table th was fontsize-heading-xxs, it was set to 700. Same goes for the table caption. I wonder if we should revert these changes?

VilleMiekkoja: Asked Roni and the table caption and th should not use heading styles, but body styles. So I'll change these back to body styles:

- [ ] Revert table caption and th changes

ronijaakkola (commented):
Looks like the step heading has not been updated.

image

VilleMiekkoja:
> Looks like the step heading has not been updated.
> 
> image

- [ ] Update step heading

packages/react/src/components/linkbox/Linkbox.module.scss
VilleMiekkoja: - [ ] Use helper classes in linkbox

packages/react/src/examples/forms/validation.scss
VilleMiekkoja: - [ ] Use helper class

packages/react/src/components/linkbox/Linkbox.module.scss
```diff
@@ -41,52 +39,72 @@
...
+    padding-right: 22px;
+    padding-bottom: 70px;
+  }
+
+  .paddingWithoutImageAndWithBackground {
+    padding-top: 22px;
+  }
+}
+
+.contentLarge .text {
```
minevala, line 83: Would it be possible to use the spacing tokens where they would make sense?
I know it is not possible always but for example where we are adding space between the border and the content they would make sense

minevala (commented):
Nice contemporary look 👍 
One minor suggestion

packages/react/src/components/linkbox/Linkbox.module.scss
VilleMiekkoja: - [ ] Use spacing tokens

packages/core/src/components/card/card.css
```diff
@@ -34,8 +34,9 @@
 /* HEADING */
 .hds-card__heading {
```
minevala, line 35: Is this CSS class used anymore?

minevala (approved):
Nice style update 👍 
Can you check the if the CSS class hds-card__heading is unnecessary?
I think it is good to go then

ronijaakkola:
@VilleMiekkoja Would it be possible to rebase this to the current master? It would be easier to update the new documentation when it is up to date with the current version.

ronijaakkola:
@VilleMiekkoja It would be good to see these updates in the documentation site too. It would require some trickery with these dependencies: https://github.com/City-of-Helsinki/helsinki-design-system/blob/db792349c078643c882d114f3814090175838323/new-site/package.json#L31

packages/core/src/utils/helpers.css
```diff
@@ -55,3 +55,61 @@
...
+  font-size: var(--fontsize-heading-m);
+  font-weight: 500;
+  line-height: 1.33;
+  letter-spacing: -0.2px;
+}
+
+.heading-s {
+  font-size: var(--fontsize-heading-s);
+  font-weight: 700;
+  line-height: 1.4;
```
ronijaakkola, line 99: We are now partly using line height tokens here. Mostly the `lineheight-s` token is used and other than that we use custom values. I'm wondering what is the point of having the line height tokens if we cannot use them.

Possibly remove line height tokens or add these other values as tokens? Or maybe leave 1 or 2 line height tokens; such as 1.0 and 1.5?

ronijaakkola (changes requested):

VilleMiekkoja:
> @VilleMiekkoja Would it be possible to rebase this to the current master? It would be easier to update the new documentation when it is up to date with the current version.

- [ ] Rebase to master

VilleMiekkoja:
> @VilleMiekkoja It would be good to see these updates in the documentation site too. It would require some trickery with these dependencies:
> 
> https://github.com/City-of-Helsinki/helsinki-design-system/blob/db792349c078643c882d114f3814090175838323/new-site/package.json#L31

There's a script for showing local react in new documentation site. Im not sure if the build will also include token changes, but it's a worth a try. With these steps it should be possible:

1. Build hds-design-tokens, hds-core and hds-react
2. Run `yarn link-react` in new-site folder

VilleMiekkoja:
> ErrorSummary has an incorrect heading line height:
> 
> image

The error summary styles comes from notification component label. So probably the notification label has wrong styles also? I'll fix them both:

- [ ] Fix notification and error summary label styles

ronijaakkola:
Fieldset has an incorrect font size.

image

https://share.goabstract.com/d37cf290-66cb-44c8-92ee-e89c763d512b?mode=build&sha=70f80f964ac91696bf3355168c405e6f385a075c

ronijaakkola:
Same applies for the Notification (but I guess this was fixed with the ErrorSummary.

image

ronijaakkola:
The Section component (not much used I guess) still has the old heading style.

image

ronijaakkola (changes requested):

VilleMiekkoja:
> The Section component (not much used I guess) still has the old heading style.
> 
> image

- [ ] Update section headings

packages/core/src/components/card/card.css
VilleMiekkoja: - [ ] Remove unused class hds-card__heading

packages/core/src/components/card/card.css
```diff
@@ -34,8 +34,9 @@
...
   font-size: var(--fontsize-heading-m);
```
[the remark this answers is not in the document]
VilleMiekkoja: - [ ] Improve line-height accuracy

packages/react/.storybook/index.css
```diff
@@ -1,6 +1,6 @@
 @font-face {
   font-family: 'HelsinkiGrotesk';
-  src: url('https://makasiini.hel.ninja/delivery/HelsinkiGrotesk/565d73a693abe0776c801607ac28f0bf.woff') format('woff');
+  src: url('http://makasiini.hel.ninja/delivery/HelsinkiGrotesk/565d73a693abe0776c801607ac28f0bf.woff') format('woff');
```
minevala, line 3: A reminder to fix back if the cert will get renewed 😁

packages/core/src/components/notification/notification.css
```diff
@@ -46,6 +48,13 @@
   margin-right: var(--spacing-2-xs);
 }
 
+/* BODY */
+
+.hds-notification__body {
+  font-size: var(--fontsize-body-s);
```
ronijaakkola, line 54: This does not match the design in Abstract but then again, neither does the current implementation. I will fix designs to match these.

ronijaakkola:
Something is off on many components on the demo site. It seems that either Helsinki Grostesk is not working for many components or the font style has changed. These typography changes should not affect most component since only heading styles changed.

Demo:
image

Live:
image

packages/core/src/components/notification/notification.css
ronijaakkola: Apparently, this has already been fixed to the implementation - the demo just was not up to date. So disregard this.

VilleMiekkoja:
> Demo

Updated the demo, the issues should now be fixed

VilleMiekkoja:
> Something is off on many components on the demo site. It seems that either Helsinki Grostesk is not working for many components or the font style has changed. These typography changes should not affect most components since only heading styles changed.
> 
> Demo: image
> 
> Live: image

Updated the demo. The issues should now be fixed

ronijaakkola (approved):

PR #709, opened 2022-04-14
Change default link size from S to M

VilleMiekkoja:
## Description

Change default link size from S to M. M is more common, so we want it to be the default. This is a breaking change.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1246

## How Has This Been Tested?

image


minevala (commented):
Good fix for default link size 👍 
I noticed that only the icon size had changed in the Link_that_opens_in_a_new_tab reference image. And only the reserved space in Link_default reference image? 🤔

VilleMiekkoja:
> Good fix for default link size 👍 I noticed that only the icon size had changed in the Link_that_opens_in_a_new_tab reference image. And only the reserved space in Link_default reference image? 🤔

According to github the text should have also changed in both:

image

minevala (approved):
Good job 🏅 
Only thing that bothers me is the alignment of external icon in the medium size linkö.
But it has been like that from the start 🤷

PR #714, opened 2022-05-05
Add support for Link component ref

minevala:
## Description
- Pass ref to Link components nativa anchor element

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1242

## Motivation and Context
- React frameworks' (for example as Next and Gatsby) link implementations need to pass ref to be able to optimize page loads etc. https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-functional-component

## How Has This Been Tested?
- Locally on dev machine


VilleMiekkoja (approved):
Good job! 🏅

Issue #493, opened 2021-07-06
Add useState call to useMobile hook


valtterival:
## To which feature the request relates to

This request is related to the useMobile hook and the Navigation component in the 'react' package.

## What

I propose adding a `updateState();` call under `window.addEventListener('resize', updateState);` inside useMobile.tsx .

## Why

This would ensure that the `isMobile` state is up to date, when a web page using it is loaded for the first time on mobile.
Navigation item styles under navigation user would be correct.

## Related material
> Optionally, you may attach screenshots, Sketch files etc. to make your request more understandable. Also, you may provide links to similar content in other design systems.

useMobile.tsx

```js
import { useEffect, useState } from 'react';

const MOBILE_BREAKPOINT = 768;
const checkIfMobile = (): boolean => (typeof window !== 'undefined' && window.innerWidth < MOBILE_BREAKPOINT) || false;

export const useMobile = (): boolean => {
  const [isMobile, setIsMobile] = useState(false);

  useEffect(() => {
    const updateState = () => setIsMobile(checkIfMobile());

    window.addEventListener('resize', updateState);
    updateState();

    return () => window.removeEventListener('resize', updateState);
  }, []);

  return isMobile;
};
```

https://user-images.githubusercontent.com/85603770/124589736-fe53d400-de62-11eb-8e67-60d859a1d8f7.mp4

minevala:
Fixed in https://github.com/City-of-Helsinki/helsinki-design-system/releases/tag/v1.15.0

PR #717, opened 2022-05-11
Call updateState to ensure useMobile hook to initialize state

minevala:
## Description
- Call useMobile updateState on useEffect

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-889
- https://github.com/City-of-Helsinki/helsinki-design-system/issues/493

## Motivation and Context
- The state of useMobile hook is false when the page is loaded. This causes the search input to be hidden and variants do not have correct styles in mobile Navigation. It also causes that the SideNavigation is missing the aria-hidden attribute.

## How Has This Been Tested?
- locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/use-mobile-demo/iframe.html?id=components-navigation--default&args=&viewMode=story)
☝️ Test in mobile screen size and open navigation to see the missing search input

### DIFF
--- a/packages/react/src/hooks/useMobile.tsx
+++ b/packages/react/src/hooks/useMobile.tsx
@@ -10,6 +10,7 @@ export const useMobile = (): boolean => {
     const updateState = () => setIsMobile(checkIfMobile());
 
     window.addEventListener('resize', updateState);
+    updateState();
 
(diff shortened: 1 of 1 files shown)


VilleMiekkoja (approved):
Good job! 🎉

PR #718, opened 2022-05-11
Fix/sidenavigation mobile keyboard navigation

minevala:
## Description
- render Sidenavigation's menu only when the screen is larger than the mobile or when the menu is open in mobile screen sizes
- Depends on #717

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1241

## Motivation and Context
- When the user navigates Sidenavigation with a keyboard on a small screen (for example with close zoom), the focus moves from the open-menu button to the first hidden SideNavigation menu button. This will confuse the user if the purpose was to navigate next to the page content.

## How Has This Been Tested?
- locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/sidenavigation-mobile-fix/iframe.html?id=components-sidenavigation--default&args=&viewMode=story)


VilleMiekkoja (approved):
Tested and seems to be working, well done 👍

PR #719, opened 2022-05-12
Fix/tabs keyboard onclick

minevala:
## Description
- handle Tab onClick also on enter and space keypresses

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1239

## Motivation and Context
- Tab (without onClick) behaves similarly if the user clicks the mouse or presses enter or space key on Tab. But now only a mouse click triggers a Tab's custom onClick handler if it exists. This causes problems when onClick is used for custom navigation between tabs.

## How Has This Been Tested?
- Locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/tab-onclick-enter-space/iframe.html?id=components-tabs--with-custom-on-click-action&args=&viewMode=story)

### DIFF
--- a/packages/react/src/components/tabs/Tabs.stories.tsx
+++ b/packages/react/src/components/tabs/Tabs.stories.tsx
@@ -137,7 +137,7 @@ export const WithCustomOnClickAction = () => {
     setIsLoading(true);
     setTimeout(() => {
       setIsLoading(false);
-    }, 1000);
+    }, 500);
   };
 
   const onTabClick = () => {
--- a/packages/react/src/components/tabs/Tabs.stories.tsx
+++ b/packages/react/src/components/tabs/Tabs.stories.tsx
@@ -128,7 +128,7 @@ export const WithCustomOnClickAction = () => {
     }, 200);
   };
 
-  const onTabClick = (activeIndex) => {
+  const onTabClick = () => {
     mockLoading();
   };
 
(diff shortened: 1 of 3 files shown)


packages/react/src/components/tabs/Tabs.test.tsx
```diff
@@ -35,4 +37,36 @@ describe('<Tabs /> spec', () => {
     const results = await axe(container);
     expect(results).toHaveNoViolations();
   });
+
+  it('should call onClick when user clicks tab', () => {
```
VilleMiekkoja, line 41: Nice unit tests 👍

packages/react/src/components/tabs/Tabs.stories.tsx
```diff
@@ -112,3 +112,46 @@ export const WithCustomTheme = () => {
...
+
+  const mockLoading = () => {
+    setIsLoading(true);
+    setTimeout(() => {
+      setIsLoading(false);
+    }, 200);
+  };
+
+  const onTabClick = (activeIndex) => {
+    setActiveTabIndex(activeIndex);
```
VilleMiekkoja, line 133: Hmm.. I think there's internal logic in the Tab component where it sets the active index on tab click. So Im not sure if this is needed?

minevala: -[x] remove unnecessary state update from story onClick

VilleMiekkoja (approved):
Nice fix 🏅

PR #722, opened 2022-05-17
Fix/new doc site in safari

minevala:
## Description
- Disable the Playground component's Core code validation when regexp lookbehind is not supported

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1240

## Motivation and Context
- Documentation site's Playground component validates Core code with html-validate library. The validation broke the client-side javascript in the Safari browser. Safari does not support regexp lookbehind which is used in html-validate implementation.
- https://gitlab.com/html-validate/html-validate/-/issues/147

## How Has This Been Tested?
- locally on the dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/docsite-safari-fix/)


VilleMiekkoja (approved):
Good job on finding the core issue here and providing a nice solution! 👍

PR #727, opened 2022-05-23
Navigation shelf [Feature]

Arskiainen:
## Description

Adds a navigation shelf to the mobile navigation.
New NavigationRow variant "inlineShelf" added, which modifies mainly the mobile styles for the row and row items to keep it as a row.
Added new input parameter "shelved" to Navigation component, which defines items to be moved to the shelf in mobile.
Modified rearrangeChildrenForMobile to facilitate shelving.
Added inline function "moveComponentToShelf" to facilitate moving the items to shelf.
New story Shelf added to the preview the component variant.  
Small style changes to help shelf look ok.

This is a PoC and needs more work.

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1318

## Motivation and Context

Employment services designs, which is based on some combination of HDS and the new Drupal frontends and the previous employment services site, define the header region with some unique solutions that don't work out-of-the box with the current HDS navigation components. This should rectify that, without breaking existing uses.

## How Has This Been Tested?

Storybook story

Screenshot 2022-05-23 at 14 40 02

[👉 Demo](https://city-of-helsinki.github.io/hds-demo/navigation-shelf/?path=/story/components-navigation--shelf)

### DIFF
--- a/packages/react/src/components/navigation/Navigation.interface.ts
+++ b/packages/react/src/components/navigation/Navigation.interface.ts
@@ -6,7 +6,7 @@
  *
  * `inline` - items will be displayed in the header
  */
-export type NavigationVariant = 'default' | 'inline';
+export type NavigationVariant = 'default' | 'inline' | 'inlineShelf';
 
 export type NavigationTheme = 'light' | 'dark' | NavigationCustomTheme;
 
--- a/packages/react/src/components/navigation/Navigation.module.scss
+++ b/packages/react/src/components/navigation/Navigation.module.scss
@@ -174,6 +174,15 @@
   }
 }
 
+// mobile shelf
+.mobileShelf {
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+  border-bottom: 1px solid var(--header-divider-color);
+  font-size: var(--fontsize-body-s);
+}
+
 /* MEDIA */
 
 @media x-small-only {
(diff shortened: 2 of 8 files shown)


ronijaakkola:
Thanks for the PR and the suggestion @Arskiainen!

This is an essential discussion because HDS will adapt to the new Hel.fi navigation design during the coming months. This means that we will probably not touch the current navigation implementation anymore but rather make sure that the new navigation will support these needs. 

If you are interested what the new Hel.fi navigation looks like, it can be found on some of the soft-launched sites such as [the Health Stations page](https://www.hel.fi/en/health-and-social-services/health-care/health-stations). Note, that Hel.fi project is still revising the navigation and the one you see on that site probably will not be the final version. So I would not try to follow that navigation, for now, the HDS navigation is the current source of truth.

Here are a couple of questions we would be interested in hearing your comments:
- What kind of items custom items do you need to put to the navigation? I assume that "Oma-asiointi" is some kind of an external link. Is it the same as a log-in?
- Apparently, these custom items need to be always visible on the mobile size navigation as well? Am I correct? Would it be ok to hide them when the user starts scrolling (e.g. see how the mobile language selector works on [this site](https://www.hel.fi/en/health-and-social-services/health-care/health-stations))?
- Is the custom element required to be a button? Could it be a custom nav element with the same style as the "default" ones (see the image below)?
image

ronijaakkola:
After looking at this from the design point of view, it may require some work.

Having the language select and the custom action button on the same row does not look good. It probably will get some comments from the accessibility audit as well. To be safe, "shelved" elements should probably take their own row. It is also worth considering not making the language select to take its own row. The HDS nav already supports showing the language selection in the mobile view - yes, it looks different from the Hel.fi nav but it works. So this change would only be done to make it look like the Hel.fi nav which seems pointless if we are redoing the navigation later after the summer anyway.

Having custom actions visible in the mobile navigation is a good addition though and I think this would be quite a trivial change.

minevala:
Closed for now. Let's if we get back to this later.

PR #737, opened 2022-06-03
[Fix] Update grayscale hex values

ronijaakkola:
## Description
Some of the grayscale token hex values do not match the exact percentage values. Since tokens are named with percentages, hex values should also match the percentage. The change is so small that we might as well make the change now. Technically this can be considered a breaking change since some contrast values may depend on these colours in some projects.

Updated the following grayscale tokens:
- color-black-70 from `#4c4c4c` to `#4d4d4d`
- color-black-40 from `#999898` to `#999999`
- color-black-30 from `#b2b2b2` to `#b3b3b3`
- color-black-10 from `#e5e5e5` to `#e6e6e6`
- color-black-5 from `#f1f1f1` to `#f2f2f2`

## Related Issue
https://helsinkisolutionoffice.atlassian.net/browse/HDS-1309

## How Has This Been Tested?
Tested by running the documentation site locally. The documentation site uses grayscale tokens directly.


laitine (commented):
Changes checked, all hex codes fixed. Ok to merge, if no accessibility issues come from changed contrasts.

ronijaakkola:
Here are the changes to WCAG contrast values. For shades up from 50% the background is white and below 50% the background is black.

- color-black-70 from 8.58:1 to 8.45:1 (decreased ↘️)
- color-black-40 from 7.29:1 to 7.37:1 (increased ↗️)
- color-black-30 from 9.9:1 to 10.01:1 (increased ↗️)
- color-black-10 from 16.67:1 to 16.82:1 (increased ↗️)
- color-black-5 from 18.59:1 to 18.75:1 (increased ↗️)

Since changes are so small and ratios are not even close to the threshold, there should not be any accessibility issues caused by this change in the HDS libraries. However, we cannot know how projects have used this color and on what background. So this should still be marked as a breaking change.

ronijaakkola:
The changes now also pass the visual regression tests.

ronijaakkola:
@laitine Contrasts are now checked and reference images updated. If all looks ok, mark this as accepted so it is ready for the release.

laitine (approved):
Looks like we got lucky with greyscale hex contrasts so this now is ready for release.

PR #739, opened 2022-06-03
[Fix] Make DialogContent scrollable only if scrollable prop is true

vvornanen:
## Description

Enable scrolling on `DialogContent` only if `scrollable` prop is defined.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1241

## Motivation and Context

Using a `DateInput` inside `DialogContent` opens the date picker inside the dialog content, making it scrollable whether the dialog is defined as scrollable or not. After this fix, the date picker opens over the dialog as expected.

## How Has This Been Tested?

Tested locally in Storybook with non-scrollable and scrollable content.

## Screenshots

With `scrollable` set to true and false:

https://user-images.githubusercontent.com/8843354/171940992-f5334877-4530-42b6-b152-9cc1b51ec121.mov

### DIFF
--- a/packages/react/src/components/dialog/Dialog.stories.tsx
+++ b/packages/react/src/components/dialog/Dialog.stories.tsx
@@ -7,6 +7,14 @@ import { Dialog } from './Dialog';
 import { IconAlertCircle, IconInfoCircle, IconPlusCircle, IconTrash } from '../../icons';
 import { DateInput } from '../dateInput';
 
+const argTypes = {
+  variant: {
+    options: ['primary', 'danger'],
+    control: { type: 'radio' },
+    defaultValue: 'primary',
+  },
+};
+
 export default {
   component: Dialog,
   title: 'Components/Dialog',
@@ -16,7 +24,13 @@ export default {
   },
   args: {
     id: 'example-dialog',
+    scrollable: false,
+    boxShadow: false,
+    theme: {},
+    style: {},
+    closeButtonLabelText: 'Close',
   },
+  argTypes,
 };
 
 export const Default = (args) => {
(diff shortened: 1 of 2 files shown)


minevala (commented):
Looks good. Seems to fix the issue 👍 
I was thinking that maybe we should fix the story args as well. At least for the default Dialog so library users can try different properties. 
I was thinking this kind of changes in the default Story:
```js
const argTypes = {
  variant: {
    options: ['primary', 'danger'],
    control: { type: 'radio' },
    defaultValue: 'primary',
  },
};

export default {
  component: Dialog,
  title: 'Components/Dialog',
  parameters: {
    controls: { expanded: true },
    loki: { skip: true },
  },
  args: {
    id: 'example-dialog',
    scrollable: false,
    boxShadow: false,
    theme: {},
    style: {},
    closeButtonLabelText: 'Close',
  },
  argTypes,
};

export const Default = (args) => {
  const openButtonRef = useRef(null);
  const [open, setOpen] = useState<boolean>(false);
  const close = () => setOpen(false);
  const titleId = 'custom-dialog-title';
  const descriptionId = 'custom-dialog-content';

  return (
    <>
      <Button ref={openButtonRef} onClick={() => setOpen(true)}>
        Open Dialog
      </Button>
      <Dialog
        id={args.id}
        scrollable={args.scrollable}
        boxShadow={args.boxShadow}
        theme={args.theme}
        style={args.style}
        variant={args.variant}
        closeButtonLabelText={args.closeButtonLabelText}
        aria-labelledby={titleId}
        aria-describedby={descriptionId}
        isOpen={open}
        focusAfterCloseRef={openButtonRef}
        close={close}
      >
```
I can fix these too

PR #743, opened 2022-06-06
Feature/dropdownlink

Arskiainen:
## Description

Adds a NavigationDropdownLink component that works on desktop via hover allowing the label to be a link. The new component is directly based on NavigationDropdown.
Adds a hoverAndClick boolean to MenuButton to position label outside the button element and adds hoverProps object and useHoverProps state to open the Menu on hover when useMobile returns false.

## Related Issue

- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1317

## Motivation and Context

This enables us to use the planned IA for employee services with the HDS navigation and allows the HDS navigation to work similar to the navigation in the newer hel.fi Drupal frontend without breaking existing uses of HDS navigation.

## How Has This Been Tested?

Storybook story.

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/navigation-dropdownlink/iframe.html?id=components-navigation--dropdown-links&args=&viewMode=story)


packages/react/src/internal/menuButton/MenuButton.tsx
```diff
@@ -40,24 +45,31 @@ export type MenuButtonProps = React.PropsWithChildren<{
    * Spacing between the toggle button and the menu
    */
   menuOffset?: number;
+  /**
+   * Render label outside button and open menu via hover.
+   */
+  hoverAndClick?: boolean;
```
minevala, line 51: Rename suggestion: withHoverOpen or something else?
Not important though, but we have boolean props named like withBorder etc

packages/react/src/components/navigation/navigationDropdownLink/NavigationDropdownLink.module.scss
```diff
@@ -0,0 +1,112 @@
...
+    > * {
+      font-weight: normal;
+
+      // dropdown menu item icon
+      span:first-of-type:not(:only-of-type) {
+        align-self: flex-start;
+        margin: 0 var(--spacing-2-xs) 0 calc(var(--spacing-3-xs) * -1);
+      }
+    }
+  }
```
minevala, line 26: Could we use the _.navigation_common.scss to get rid of duplicate CSS with NavigationDropdown.scss?
Maybe this could be a %dropDownItemBase and it could be used like:
```css
> div {
  @extend  %dropDownItemBase;
}
```

packages/react/src/components/navigation/navigationDropdownLink/NavigationDropdownLink.module.scss
```diff
@@ -0,0 +1,112 @@
...
+      top: 0 !important;
+
+      > * {
+        padding-left: var(--spacing-2-xs);
+
+        &:focus {
+          box-shadow: 0 0 0 var(--header-focus-outline-width) var(--header-focus-outline-color);
+        }
+      }
+    }
```
minevala, line 104: Same duplication suggestion as before

packages/react/src/components/navigation/navigationDropdownLink/NavigationDropdownLink.module.scss
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+@media medium-up {
+  .dropdownLinkItem.dropdownLinkItem > button {
+      align-self: stretch;
```
minevala, line 110: I did not see any effect with this. Maybe I am missing something 🤔

minevala (commented):
Nice work. I had a few suggestions but it looks good as it is already

packages/react/src/components/navigation/navigationDropdownLink/NavigationDropdownLink.module.scss
Arskiainen: The difference isn't much, but i believe it's meaningful. This is the same for the dropdownItem, except it includes the label area also.
Screenshot 2022-06-29 at 15 49 54
Screenshot 2022-06-29 at 15 49 48

ronijaakkola:
It would be slightly better if the DropdownLink focus style would visually match the style of the standard NavLink. The height of the focus container should be 100%.

It is up to you if this should be fixed. Depends how hard it would be to fix.

**Link**:
image

**DropdownLink**:
image

minevala:
> It is up to you if this should be fixed. Depends how hard it would be to fix.

Could it be fixed by setting the link to block or flex and adding height: 100% ?
I'm a bit worried of the side-effects it might have, Navigation is not the most tested component.

Arskiainen:
I thought the last style tweaks actually fixed that as well. I guess the
demo was not updated after those or I goofed something.

minevala:
Demo is now up-to-date

minevala (approved):
Nice new feature 🏅

PR #744, opened 2022-06-06
Content/deprecation warning to old docsite

VilleMiekkoja:
## Description

Adds a deprecation warning to old docsite for each page.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1307

## Motivation and Context

It tells users that they are viewing the old docs, and guides users to the new docsite.

## How Has This Been Tested?

- Locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/docsite-deprecated-warning/)

image


site/.env.production
```diff
@@ -0,0 +1 @@
+GATSBY_SHOW_DEPRECATED_WARNING=FALSE
```
VilleMiekkoja, line 1: We need to change this to TRUE when we release 2.0.0. Or to be more specific, when we build the old site and serve it under different folder.

minevala (approved):
Nice work 🏅

PR #755, opened 2022-06-13
Content/Fix broken site title link

VilleMiekkoja:
## Description

Fix broken navigation title link.

The solution has a minor setback. It works perfectly on local and production environments. But in demo environment, the link works nicely when clicked, but the href is incorrect (it points to production env). The reason the link still works is because it uses internal routing by javascript.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1022

How is the issue related? Well, I run lighthouse SEO tests and according to it broken links have negative impact on SEO, so decided to fix it.

## How Has This Been Tested?

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/docsite-fix-title-link/)

image

### DIFF
--- a/new-site/.env.development
+++ b/new-site/.env.development
@@ -1,0 +1,1 @@
+SITE_URL=http://localhost:8000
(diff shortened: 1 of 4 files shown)


minevala (approved):
Nice fix 👍

PR #761, opened 2022-06-15
Fix html-sanitizing on server side (fixes build)

minevala:
## Description
- Replace dompurify library with sanitize-html

## Motivation and Context
- Pagination pages are the first pages using only HTML in the Playground component. When Gatsby tried to create the initial pages on the node server, the domPurify.sanitize did not exist and the build failed.
This is because domPurify has a bit different API on the server and it needs to be initialized. 
Sanitize-html library has the same API on the client and the server.

## How Has This Been Tested?
- Locally on dev machine

### DIFF
--- a/new-site/src/components/Playground.js
+++ b/new-site/src/components/Playground.js
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useMDXScope } from 'gatsby-plugin-mdx/context';
 import { LiveProvider, LiveEditor, LiveError, LivePreview, withLive } from 'react-live';
-import DOMPurify from 'dompurify';
+import sanitizeHtml from 'sanitize-html';
 import { Tabs, TabList, TabPanel, Tab, Button, IconArrowUndo } from 'hds-react';
 import theme from 'prism-react-renderer/themes/github';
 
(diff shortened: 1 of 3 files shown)


VilleMiekkoja (approved):
Nice fix 👍

new-site/src/components/Playground.js
```diff
@@ -9,6 +9,58 @@ import theme from 'prism-react-renderer/themes/github';
 import './Playground.scss';
 import LiveErrorCore from './LiveErrorCore';
 
+const sanitizeConfig = {
+  allowedTags: false,
+  disallowedTagsMode: 'discard',
+  allowedAttributes: {
```
VilleMiekkoja, line 15: How do we know which attributes should be allowed and which not? This seems little odd to do it this way. Would be maybe easier if we could list attributes we do not want 🤔

new-site/src/components/Playground.js
```diff
@@ -9,6 +9,58 @@ import theme from 'prism-react-renderer/themes/github';
...
+    '*': [
+      'aria-describedby',
+      'aria-hidden',
+      'aria-label',
+      'aria-labelledby',
+      'aria-level',
+      'class',
+      'checked',
+      'd',
```
VilleMiekkoja, line 24: The pagination component has data-next attribute, we probably need to allow it?

new-site/src/components/Playground.js
minevala: I'm not sure is it possible: https://github.com/apostrophecms/sanitize-html#readme

new-site/src/components/Playground.js
minevala: - [ ] add missing attribute

PR #762, opened 2022-06-16
Content/remove contributing blockquotes

minevala:
## Description
- Remove unnecessary blockquotes

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1280

## Motivation and Context
- Previously the blockquotes created a background for an are but we need to revisit blockquotes once we implement them into HDS. So let's remove the unstyled blocks for now.

## How Has This Been Tested?
- Locally on dev machine

Before:
Screenshot 2022-06-16 at 17 52 52

After:
Screenshot 2022-06-16 at 17 53 25


ronijaakkola (approved):
I guess those blockquotes were used to visually links the image and the text together. So this breaks the structure of the file a bit. But it is just one file so we can worry about this later if needed.

VilleMiekkoja (approved):
👍

PR #765, opened 2022-06-20
Fix/manifest icon path

VilleMiekkoja:
## Description

Fix manifest icon path

## Motivation and Context

Android-chrome icons were not found due to wrong path.

## How Has This Been Tested?

- Locally on dev machine.

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/docsite-manifest-icon-path-fix/)


minevala (approved):
Yes, static can be omitted now 👍

PR #789, opened 2022-07-07
Feature / Cookie Consent component

minevala:
## Description
- Add CookieConsent component (modal and page version)

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-801
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-868

## Motivation and Context
- This is a common Cookie Consent component that can be used across city services. It makes it easier for the user to manage cookies because of unified UX and shared cookies

## How Has This Been Tested?
- Locally on dev machine
- Loki and functional tests

[Storybook Demo](https://city-of-helsinki.github.io/hds-demo/cookie-consent/?path=/story/components-cookieconsent--modal-version)

[Docs Demo](https://city-of-helsinki.github.io/hds-demo/cookie-consent-docs/components/cookie-consent)

## Texts from the previous fork PR:

Users must give consents to cookies and other means user could be tracked. These consents must be given by the user before a website can store data to cookies or, for example, send data to analytics.

The purpose of this PR is to
- provide a controller for reading / writing the cookie where user's consents are stored
- provide "names" for common consents, so websites in same domain can use consents given in another website for the same purpose
- demonstrate usage of the controller and context in React applications

### Consents

There are required consents that users must approve. The are also optional consents. All consents, required and optional, must be handled by the user. Handling means approving all consents, approving only required consents or approving all required and some optional.

Cookie consent dialog is shown if
- all required consents are not approved
- any listed consent is not found in the cookie. This happens when a new consent is added after user has already handled consents before.
- user has rejected an optional consent A in service X, but that same consent A is marked as required in service Z. The dialog is shown is Service Z.

Current common consents are: 'matomo', 'tunnistamo', 'language', 'marketing', 'preferences'.  These are just placeholders. Any service using cookieConsentController can use these common consents, make up its own names or combination on both. Common consent can be required or optional.

### Core files:
**packages/react/src/components/cookieConsent/cookieConsentController.ts**
Reads and parses the consents from a cookie and also updates the cookie on the user's device. Stores current state of consents before saving by providing getters and setters. Common consent names are exported here.
This file is not a React file and can be used in any javascript project.

**packages/react/src/components/cookieConsent/CookieConsentContext.tsx**
Uses cookieConsentController and provides its functions and consent data in a React Context. It also has two callbacks which expose consent changes outside of the context:
- "onAllConsentsGiven" is called when user has saved consents. Not called if user won't change consents
- "onConsentsParsed" is called once, when context is initialized and  stored cookies are read from the device.

**packages/react/src/components/cookieConsent/CookieConsent.tsx**
React component which animates and renders the given content for the user when dialog should be shown.

**packages/react/src/components/cookieConsent/__storybook__/* **
Demo files for Storybook and developers


packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const getContainerElement = (): HTMLElement | null => document.getElementById(containerId);
+  const [isDomReady, setIsDomReady] = useState<boolean>(false);
+
+  useEffect(() => {
+    setTimeout(() => setPopupTimerComplete(true), popupDelayInMs);
+  }, []);
+
+  useEffect(() => {
+    const containerElement = getContainerElement();
+    if (shouldShowModal && !isDomReady) {
```
minevala, line 32: Maybe some combined state would be better than resolve these from two different states

site/src/docs/patterns/cookies/basics.mdx
```diff
@@ -0,0 +1,114 @@
...
+| Only optional cookies               | Yes                            | Show the cookie consent banner and ask for consent for each optional cookie category. The list of cookies has to be available on a separate cookies page.                                                              
+| Both necessary and optional cookies | Yes, for all optional cookies  | Show the cookie consent banner and ask for consent for each optional cookie category. The list of cookies has to be available on a separate cookies page.                  
+| No cookies at all                   | No                             | No actions needed.                                                                                                  
+[Table 1:Cases when the consent is needed and when it is not]|
+
+Refer to the next chapter to learn how to categorise your cookies and decide which cookies should be necessary and which should not.
+
+## Cookie information and categorisation
+
+The cookie consent needs to be asked when your service becomes public (even during alpha/beta periods). [Service under Hel.fi? / Service on its own?]
```
VilleMiekkoja, line 45: I think this is for all services, not just under hel.fi

site/src/docs/components/cookie-consent/index.mdx
```diff
@@ -0,0 +1,39 @@
...
+---
+
+import { Link } from 'hds-react';
+import TabsLayout from './tabs.mdx';
+
+export default ({ children, pageContext }) => <TabsLayout pageContext={pageContext}>{children}</TabsLayout>;
+
+## Usage
+
+This documentation page is about HDS CookieCompliance component. If you are looking for documentation about cookies in general, please refer to <a href="/patterns/cookies">HDS cookie pattern documentation page</a>.
```
VilleMiekkoja, line 14: This link is not working on the demo site. It needs to use the gatsby `withPrefix` function. And also, we should probably use the HDS Link component in here.

site/src/docs/components/cookie-consent/index.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+### Example
+
+<Link href="/storybook/react/?path=/story/components-cookieconsent--modal-version">View modal example in Storybook</Link>
```
VilleMiekkoja, line 18: Same problem here, the Link is not working on the demo site. We need to use the gatsby `withPrefix`. Also because the link is to a storybook, that is outside the current docsite app, we should probably add the prop `openInNewTab` to open the link in to a new tab

site/src/docs/components/cookie-consent/index.mdx
```diff
@@ -0,0 +1,39 @@
...
+- You must offer the cookie consent modal and page in all languages that your service supports.
+  - By default, the cookie consent banner should use the same language as the service is using.
+  - The component has its language switcher so the user can easily switch the language of the dialog even if they opened the service with a language they do not understand.
+
+### Variations
+
+### Modal
+The default banner variant opens at the bottom of the viewport. It is opened if the consent for cookies has not been given or if the cookies have changed. The banner can be expanded to view and change cookie settings.
+
+<Link href="/storybook/react/?path=/story/components-cookieconsent--modal-version">View modal example in Storybook</Link>
```
VilleMiekkoja, line 34: Add `withPrefix` function to href, and `openInNewTab` prop to the link.

site/src/docs/components/cookie-consent/index.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+### Rendered on a page
+The cookie consent expanded settings page can be rendered as a separate page. This is used to create a cookie page that the user can access via a link in the [Footer](/components/footer) element. This allows the user to view or edit cookie settings after the initial consent has been given.
+
+<Link href="/storybook/react/?path=/story/components-cookieconsent--page-version">View page example in Storybook</Link>
```
VilleMiekkoja, line 39: Add `withPrefix` function to href, and `openInNewTab` prop to the link.

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <>
+      <CookieModal contentSource={contentSource} />
+      <Application />
+    </>
+  );
+};
+```
+
+<Link href="/storybook/react/?path=/story/components-cookieconsent--modal-version">View working modal demo in Storybook</Link>
```
VilleMiekkoja, line 67: The link is broken in demo environment. Add  gatsby`withPrefix` function to href, and `openInNewTab` prop for the Link component. As the storybook goes out of the docsite app, I think we should open it into a new tab so user can more easily get back to docsite.

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+  return (
+    <main>
+      <CookiePage contentSource={contentSource} />
+    </main>
+  );
+};
+```
+
+<Link href="/storybook/react/?path=/story/components-cookieconsent--page-version">View working page example in Storybook</Link>
```
VilleMiekkoja, line 103: The link is broken in demo environment. Add  gatsby`withPrefix` function to href, and `openInNewTab` prop for the Link component. As the storybook goes out of the docsite app, I think we should open it into a new tab so user can more easily get back to docsite.

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+### Packages
+| Package           | Included                      | Storybook link                                                                                                                      | Source link   |
+| ----------------- | ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- | ------------- |
+| **HDS React**     | <IconCheckCircleFill /> Yes   | <Link size="M" external href="/storybook/react/?path=/story/components-cookieconsent--modal-version">View in Storybook</Link>  | <Link size="M" external href="https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/components/cookieconsent">View source</Link> |
```
VilleMiekkoja, line 108: The storybook link is broken in demo environment. Add  gatsby`withPrefix` function to href, and `openInNewTab` prop for the Link component. As the storybook goes out of the docsite app, I think we should open it into a new tab so user can more easily get back to docsite.

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -0,0 +1,119 @@
...
+| **HDS Core**      | <IconCrossCircle /> No        | - | - |
+
+### Properties
+Note! You can find the full list of properties in the <Link size="M" href="/storybook/react/?path=/story/components-cookieconsent--default" external>React Storybook.</Link>
```
VilleMiekkoja, line 112: The storybook link is broken in demo environment. Add  gatsby`withPrefix` function to href, and `openInNewTab` prop for the Link component. As the storybook goes out of the docsite app, I think we should open it into a new tab so user can more easily get back to docsite. Also not sure if we should use external here or not? It goes to a different website, but stays on the same domain.

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+import TabsLayout from './tabs.mdx';
+
+export default ({ children, pageContext }) => <TabsLayout pageContext={pageContext}>{children}</TabsLayout>;
+
+## API
+
+### Component properties
+
+Common texts, groups and cookie data can be found in <Link href="">HDS-provided content file</Link>. The file contains all user interface strings, language options, descriptions for common cookie groups, and details about common Helsinki cookies. By default, this content file is used. You can also override parts of it and add your project-specific groups and cookies.
```
VilleMiekkoja, line 14: Missing Link href. Do we wish that user could download the HDS provided content file?

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `focusTargetSelector`                                                                                                            | No       | When the modal is closed or the ESC key is pressed when it is open, the focus must be shifted outside the modal. This defines the element to set focus to with this selector (string). Not needed when modal is not used. Throws an error is not set for modal.                                                                |
+[Table 1:ContentSource object properties]|
+
+##### ContentSource examples
+You can find examples of ContentSource property contents in <Link href="https://github.com/NikoHelle/helsinki-design-system/blob/feature/cookie-consent/packages/react/src/components/cookieConsent/CookieConsent.stories.tsx" external>the HDS CookieConsent Storybook examples (FIX LINK)</Link>.
```
VilleMiekkoja, line 36: Link href needs to be fixed

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+#### CookieDomain 
+Rarely used. Cookie consents are stored to a cookie whose domain is automatically set to accept subdomains: *.hel.fi or *.hel.ninja or *.domain.com. Some sites are included in the public suffix list and cookies for those sites cannot accept subdomain cookies. Github.io is one of those sites and HDS examples do not work there unless the cookie domain can be overridden. Therefore `cookieDomain` was added.
```
VilleMiekkoja, line 39: > Github.io is one of those sites and HDS examples do not work there unless the cookie domain can be overridden. Therefore `cookieDomain` was added.

I think this is good information to HDS developers internally, but maybe unnecessary and little misleading info to others. The cookieDomain is a good addition regardless of public suffix list, as not all services necessary wish to expose their cookies to sub domains. I suggest we replace this with:

"If your site domain is included in the public suffix list (Link to the public suffix list), or you do not wish to expose your cookies to subdomains, or you want to set the domain manually for some reason, then use the `cookieDomain` property.

VilleMiekkoja, line 39: > Rarely used.

Maybe replace this to >Rarely needed

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+#### Detailed info about properties
+
+##### Required and optional cookie categories
```
VilleMiekkoja, line 43: Reading this it's little unclear that which properties these are? The main api seems to have only two properties: `contentSource` and `cookieDomain`. Are these part of the `contentSource` property? Maybe we could change the header to something like:

#### contentSource.requiredConsents / contentSource.optionalConsents

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `optionalCookies`                                                                                                                | No       | Optional cookies can have partial data or all data described in the table below. Missing texts are auto-filled. See examples below this table.                                                                                                                                                                          |
+| `requiredConsents.groups`, `optionalConsents.groups`                                                                             | No       | List of cookie groups. All cookies must belong to a group. Groups with property `commonGroup` will be auto-filled from `content.json`.                                                                                                                                                                                  |
```
VilleMiekkoja, line 25: Should this be `requiredCookies.groups`? I was unable to find the property `requiredConsents` from the `contentSource` property.

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `requiredConsents.groups[].cookies`, `requiredConsents.cookies`, `optionalConsents.groups[].cookies`, `optionalConsents.cookies` | No       | List of individual cookies. All cookies must belong to a group or they must have `propertycommonGroup` or `groupId`. The `groupId` must match `group.id` of a custom group and the `commonGroup` must match groups in `content.json`. Do not give both.                                                                |
```
VilleMiekkoja, line 26: Same here, probably we mean `requiredCookies.groups[].cookies` instead?

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `currentLanguage`                                                                                                                | **Yes**  | Currently selected language. Used for picking translations from `content.json`. Is also set to `content.language.current`. Supported values are `fi`, `sv`, and `en`.                                                                                                                                                   |
+| `requiredCookies`                                                                                                                | No       | Required cookies can have partial data or all data described in the table below. Missing texts are auto-filled. See examples below this table.                                                                                                                                                                         |
```
VilleMiekkoja, line 23: Would it be a good idea to add link to the table (to the nearest header) where the content of `requiredCookies` is described in more detail? It could help understanding the complex structure of the content source.

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+##### Required and optional cookie categories
+
+| Property                    | Description                                               |
+| --------------------------- | --------------------------------------------------------- |
+| `requiredConsents`          | List of required consents and their descriptions.         |
+| `optionalConsents`          | List of optional consents and their descriptions.         |
+| `*.title`                   | Title for the consent group.                              |
+| `*.text`                    | Description for the consent group.                        |
+| `*.checkboxAriaDescription` | Aria description of the checkbox shown with the category. |
```
VilleMiekkoja, line 51: These seemed to be also optional.

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `*.groups`                  | Optional list of groups. More info is below.                 |
```
VilleMiekkoja, line 52: I suggest we replace the 

> More info is below.

WIth a link to the Cookie groups header.

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `*.cookies`                 | Optional list of cookies. More info is below.                |
```
VilleMiekkoja, line 53: I suggest we replace the

> More info is below.

WIth a link to the Cookie information header.

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `cookie[].hostName`                         | Hostname where it is used. Only for custom cookies.  Cannot be set for a common cookie.                                                                                                                                                |
+| `cookie[].description`                      | Cookie usage description. Only for custom cookies.  Cannot be set for a common cookie.                                                                                                                                                  |
+| `cookie[].expiration`                       | When does the cookie expire. Only for custom cookies. Cannot be set for a common cookie.                                                                                                                                                |
+[Table 4:Cookie information properties]|
+
+##### Callbacks
+The `content` object passed to the component can also include callback functions:
```
VilleMiekkoja, line 82: `content` -> `contentSource`

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+| Property             | Usage                                                                                                 | Arguments                                                                                                                                                                                     |
```
VilleMiekkoja, line 84: Should we add column `usage`, that would describe at least one scenario where the user would wish to use the callback?

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| -------------------- | ----------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `onAllConsentsGiven` | Called when the user has given all consents. Not called if consents are not asked for.                    | Object with cookie.id as the key and true/false as value `{matomo: true, marketing: false}`                                                                                                   |
+| `onConsentsParsed`   | Called when cookie consents are read from the browser. Called always even if consent has been given. | Object with cookie.id as the key and true/false as value `{matomo: true, marketing: false}` and boolean indicating are all consents handled. If true, there is no reason to ask for consent. |
```
VilleMiekkoja, line 87: We could maybe add this: 

> Called when cookie consents are read from the browser, **before the cookie modal will possibly be shown**

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `ui.hideSettings`                       | Text on the show/hide settings -button when settings are shown.                                                       |
+| `ui.approveAllConsents`                 | Button text for giving consent to all cookies. Shown when settings are closed and the user cannot give optional consents. |
+| `ui.approveRequiredAndSelectedConsents` | Button text for giving consent to required cookies and selected cookies. Shown when settings are open.                |
+| `ui.approveOnlyRequiredConsents`        | Button text for giving consent to only to required cookies. All optional cookie consents are revoked.                |
+| `ui.readMore`                           | Text shown on a button when modal loses focus and is shrunk.                                                          |
+[Table 7:User interface string overrides]|
+
+| Property                    | Description                                                 |
+| --------------------------- | ----------------------------------------------------------- |
+| `tableHeadings.name`        | Title of the “hostname” column in the cookie description.  |
```
VilleMiekkoja, line 113: `hostname` -> `name`

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `tableHeadings.hostName`    | Title of the “path” column in the cookie description.       |
```
VilleMiekkoja, line 114: `path` -> `hostName`. Interestingly, there is no path column described. Was it so that we decided to not include it or?

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `tableHeadings.description` | Title of the “expiration” column in the cookie description. |
```
VilleMiekkoja, line 115: `expiration` -> `description`

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `tableHeadings.expiration`  | Title of the “name” column in the cookie description.       |
```
VilleMiekkoja, line 116: `name` -> `expiration`

site/src/docs/components/cookie-consent/api.mdx
```diff
@@ -0,0 +1,135 @@
...
+| `language.onLanguageChange`          | Callback called when new language is selected. Called with new language.code as the argument. |
+[Table 9:Language properties]|
+
+### Saving consents
+
+Consents that the user has given are saved to a `city-of-helsinki-cookie-consents` cookie.
+
+Cookies are always strings and consents are stored as a JSON-formatted string such as `{"cookie-id":"true","another-cookie-id":"false"}`.
+
+The ID of the cookie is stored to `cookie.id`. `True` means the user has given consent for that cookie.  All required cookies are always stored `true` But other sites may list a common cookie optional, so the stored consent is not `true` by default.
```
VilleMiekkoja, line 135: But -> , but

VilleMiekkoja, line 135: Do not understand this sentence:

> All required cookies are always stored `true` But other sites may list a common cookie optional, so the stored consent is not `true` by default.

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja, line 108: Need to fix the vertical alignment of the icon and text

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -0,0 +1,119 @@
...
+};
+```
+
+<Link href="/storybook/react/?path=/story/components-cookieconsent--page-version">View working page example in Storybook</Link>
+
+### Packages
+| Package           | Included                      | Storybook link                                                                                                                      | Source link   |
+| ----------------- | ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- | ------------- |
+| **HDS React**     | <IconCheckCircleFill /> Yes   | <Link size="M" external href="/storybook/react/?path=/story/components-cookieconsent--modal-version">View in Storybook</Link>  | <Link size="M" external href="https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/components/cookieconsent">View source</Link> |
+| **HDS Core**      | <IconCrossCircle /> No        | - | - |
```
VilleMiekkoja, line 109: Need to fix the vertical alignment of the icon and text

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+### Properties
+Note! You can find the full list of properties in the <Link size="M" href="/storybook/react/?path=/story/components-cookieconsent--default" external>React Storybook.</Link>
+
+Also, note that this component is an input. All features supported by the HDS TextInput are also supported by this component. See [TextInput documentation page](/components/text-input) for more information.
```
VilleMiekkoja, line 114: This refers to TextInput component. It has probably been added as a base here, and we just have not set the actual properties here yet. Should we list the properties here, or add a link to the API tab?

site/src/docs/patterns/cookies/basics.mdx
```diff
@@ -0,0 +1,114 @@
...
+| **Cookie set by**      | What domain is the provider of this cookie?                                                       | "hel.fi"
+| **Purpose of use**     | Why the cookie is needed? What purpose does it serve?                                             | "Used to store information about whether visitors have given or declined the use of cookie categories used on this service."
+| **Period of validity** | How long the cookie will be valid? This means time to expire from the creation of the cookie.     | "100 days"
+[Table 2:Information that needs to be shown for each cookie]|
+
+#### What if the user has given consent to my cookies in some other Helsinki service?
+
+Using common Helsinki cookie consent brings the benefit of sharing consents. If user gives consents at site A, same consents does not have to be asked again at site B. Sites sharing cookies must have the same domain. e.g. *.hel.fi. To learn more about common Helsinki service cookies, please refer to the <Link href="/patterns/cookies/common-helsinki-cookies">Common Helsinki cookies page</Link>.
```
VilleMiekkoja, line 70: The link is broken in demo environment. Use this syntax instead to get the path prefix added:

`[Common Helsinki cookies page](/patterns/cookies/common-helsinki-cookies)`

site/src/docs/patterns/cookies/basics.mdx
```diff
@@ -0,0 +1,114 @@
...
+| **Device information**        | Device information cookies contain information collected from the user’s device that ensures the functionality of the service. | Never necessary.                                                                                                                                                                                                                                                                             | Device information, OS, and browser information                                              |
+| **Chat services**             | Chat service cookies enable communication between the user and the service.                                                    | Can only be considered necessary if real-time communication is the main purpose of the service. When the user opens the chat window, cookies related to it become necessary. Note that they are not necessary until the user opens the chat.                                                 | Chat provider cookies                                                                        |
```
VilleMiekkoja, line 97: These two sentences seems to be contradicting each other:

> Can only be considered necessary if real-time communication is the main purpose of the service. When the user opens the chat window, cookies related to it become necessary.

If a service has a chat window, it rarely is the main purpose of the service. So according to the first sentence, the chat cookies should then not be considered necessary. But according to the second sentence, they still have the option to become necessary. And we only ask cookies when the page loads. So I don't think we can change cookie from optional to necessary during the web page visit.

site/src/docs/patterns/cookies/basics.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+#### Can I use a custom category?
+
+If your cookies do not fit into any of the pre-defined categories, you are allowed to use custom categories.
+
+#### We only have a handful of cookies. Do we still have to categorize them?
+
+Yes, cookies always have to be put into a category. If you have a small number of cookies you can consider grouping them into a single category. Also, if most of the cookies are related to a single third-party service, you can group them per service (E.g. "Service: YouTube").
+
+### Prepare your CookieConsent component
```
VilleMiekkoja, line 112: We refer to React's CookieConsent component here? Technically, we currently do not have a component named CookieConsent, which might be little misleading. Instead, we currently expose these two components:

1. `CookieModal`
2. `CookiePage`

Im not sure if we need to be so specific on things. But usually when we refer to a component in the documentation, that same name can be found in the React code also.

site/src/docs/patterns/cookies/basics.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+Now that you have gathered and categorised a list of cookies for your service, it is a time to learn how the HDS CookieConsent is used. To read more about the component, please refer to the <Link href="/patterns/cookies/using-the-cookieconsent">Using the CookieConsent page</Link>.
```
VilleMiekkoja, line 114: The link does not work on demo environment. Use this syntax instead:

`[Using the CookieConsent page](/patterns/cookies/using-the-cookieconsent)`

site/src/docs/patterns/cookies/common-helsinki-cookies.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Key principles of using common cookies
+- You can use common Helsinki cookies if your service is using the *.hel.fi domain.
+- When common cookies are available, use them instead of creating custom cookies. This way the user only has to give their consent once between multiple Helsinki services.
+- When a common cookie is used, it must be used **exactly** as described on this page. If you modify the cookie in any way, it is no longer a common cookie.
+- If your service is using custom cookies alongside common cookies, you still need to ask for consent for the custom cookies (if they are considered optional).
+- You can either use all or a part of the common cookies.
+  - Note, that some of the services (e.g. Matomo) may require multiple common cookies for it to work.
+
+## Common cookies between *.hel.fi sites
```
VilleMiekkoja, line 27: Are there only 3 common cookies? I was kind of expecting more 🤔.

site/src/docs/patterns/cookies/common-helsinki-cookies.mdx
```diff
@@ -0,0 +1,45 @@
...
+The following is a list of approved common cookies between services using *.hel.fi domain. You can find this list as a JSON object in the <Link href="" external>Helsinki Design System GitHub repository</Link>.
```
VilleMiekkoja, line 28: Link missing the correct href

site/src/docs/patterns/cookies/using-the-cookieconsent.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+<StatusLabel type="alert">WIP</StatusLabel>
+
+<LeadParagraph>
+  HDS offers a consistent way of informing the user about cookies and, if needed, asking for consent. The CookieConsent component is used for this. This page describes how the component is used.
+</LeadParagraph>
+
+If you are looking for technical documentation about the CookieConsent component, please refer to <Link href="/components/cookie-consent">the component page</Link>.
+
+If you are looking for an explanation about gathering and categorising cookies, please refer to <Link href="/patterns/cookies/basics">the Cookie basics page</Link>.
```
VilleMiekkoja, line 21: These links are broken in the demo environment

site/src/docs/patterns/cookies/using-the-cookieconsent.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+## How does the cookie consent banner affect other content of the service?
+There may be elements or content on your service that do not work properly until the user has given consent to optional cookies. It is recommended to either hide this content or show clear error messages at the position of these elements that indicate why they are not available.
+
+## What happens after the user has accepted or denied the optional cookies?
+The user must be able to view and edit the cookie consent after they have initially submitted their choices. You must offer a separate cookies page that allows changing viewing the cookies and changing consent. The link to this page should be placed on the footer element of the page. Note that the HDS CookieConsent component offers functionality to render it as a page. This allows you to use the same component and data to construct a separate cookie page.
+
+If new cookies are added or current ones changed, consent should be asked from the user again the next time the service is used.
```
VilleMiekkoja, line 63: We should probably provide an example of how to technically change cookie in a way that it would trigger the cookie consent modal and show the cookie as not accepted. Because currently it is id-based, so if the id stays the same, the component won't notice that the cookie has changed.

packages/react/src/components/cookieConsent/CookieConsent.module.scss
```diff
@@ -0,0 +1,350 @@
...
+
+.aligner {
+  position: absolute;
+  bottom: 0;
+  z-index: 2;
+  width: 100%;
+  overflow-y: auto;
+  max-height: 80vh;
+  border-top: 8px solid var(--color-bus);
+  background: #ffffff;
```
VilleMiekkoja, line 20: We should probaby use hds design token variable for this color

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { VisuallyHidden } from '@react-aria/visually-hidden';
+
+import classNames from '../../../utils/classNames';
+import styles from '../CookieConsent.module.scss';
+import { CookieConsentContext, useCookieConsentUiTexts, useFocusShift } from '../CookieConsentContext';
+import { Content } from '../content/Content';
+import { useEscKey } from '../useEscKey';
+
+export function Modal(): React.ReactElement | null {
+  const cookieConsentContext = useContext(CookieConsentContext);
```
VilleMiekkoja, line 12: This might cause lot of unnecessary re-renders of the whole modal component. If you check any checkbox in the cookie modal, then the whole modal will be re-rendered again when I tested. This will cause lot of unnecessary JS to be run. For example, the ReactDOM.createPortal will be run again, and all the initial JS that are inside this component will be run again.

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const renderModal = (): JSX.Element => (
+    <div
+      className={classNames(styles.container, styles.modal, popupTimerComplete && styles.animateIn)}
+      data-testid="cookie-consent"
+    >
+      <div className={styles.aligner}>{popupTimerComplete && <Content />}</div>
+    </div>
+  );
+
+  return ReactDOM.createPortal(renderModal(), getContainerElement());
```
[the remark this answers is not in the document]
VilleMiekkoja: There might be some optimisations we could make to the rendering of this modal. When the modal opens initially, the ReactDOM.createPortal method gets called twice.

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const { settingsSaved } = useCookieConsentUiTexts();
+  const containerId = 'HdsCookieConsentContainer';
+  const getContainerElement = (): HTMLElement | null => document.getElementById(containerId);
+  const [isDomReady, setIsDomReady] = useState<boolean>(false);
+
+  useEffect(() => {
+    setTimeout(() => setPopupTimerComplete(true), popupDelayInMs);
+  }, []);
+
+  useEffect(() => {
```
VilleMiekkoja, line 30: When user accepts cookies, this useEffect function gets called twice. I wonder if we could somehow reduce the amount 🤔

packages/react/src/components/cookieConsent/modal/Modal.tsx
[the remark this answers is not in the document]
VilleMiekkoja: Also, I wonder if the whole Modal component should be memoized in the CookieModal.tsx? Because if the cookie modal is a child component to app, then every render or stage change in the parent (for example using the site while the cookie modal is visible), will cause a complete re-render of this `Modal` component. Which would also mean that the `React.DOM.createPortal` would be run again. This might lead into performance problems. Though Im not sure if you can memoize a component that will use react.context inside it? Obviously, the component should re-render again if the context changes, regardless of the memoize 🤔

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+      setIsDomReady(true);
+    } else if (!shouldShowModal && isDomReady) {
+      if (containerElement) {
+        document.body.removeChild(containerElement);
+      }
+      setIsDomReady(false);
+    }
+  }, [shouldShowModal, isDomReady, setIsDomReady]);
+
+  useEscKey(useFocusShift());
```
VilleMiekkoja, line 48: There's quite much happening here.

1. `useEscKey` is a useEffect hook, which will be run after every render. Is it necessary to run it after every render? What the hook does is that it adds an event listener for the 'esc' key. When esc is pressed, then the hook will run the callback function that has been given to it in arguments. The hook has also a return value, which will remove the event listener after the component unmounts, and also before each render. So always when the component renders or re-renders, the hook will first remove the previous event listener, and add a new event listener. That is quite much for every render of this modal. Would it be possible/reasonable to add the event listener only once when the component mounts, and then remove it when it unmounts?

2. The callback for the `useEscKey` is created by calling another hook function, called `useFocusShift`. The `useFocusShift` will return a function where focus will be forced on the cookie consent `focusTargetSelector`. So all in all, when you press an 'esc' key, the element `focusTargetSelector` will be focused. There's quite many react hooks stacked here. The `useFocusShift` uses yet another react hook, called `useCookieConsentContent`, and then returns a function, in which function `forceFocusToElement` will be called.  The `useCookieConsentContent` uses yet another react hook called `useContext(CookieConsentContext)`, to get content, which has attribute `focusTargetSelector`. Im not sure if there's too many hooks stacked, which makes following and understanding the code little harder. Though then again, some of those hooks are used in multiple places, which will reduce code repetition. The `useFocusShift` hook seem to be used only inside the Modal.tsx component, so would it be good idea to move it there? It could be there inlined or renamed with a more specific meaning.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,283 @@
...
+      });
+    } else if (action === 'approveOptional') {
+      selectOptional();
+    } else if (action === 'unapproveOptional') {
+      deselectOptional();
+    }
+    reRender();
+  };
+
+  const countApprovedOptional: CookieConsentContextType['countApprovedOptional'] = () => {
```
VilleMiekkoja, line 205: Would `getApproveRatioForOptional` be more descriptive? Or `getApprovalRatingForOptional`?

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: 1. Fixed resetting the document.addListener in each render. The passed callback is now a a cached with useCallback. 

But useAccordion returns new callbacks on each render, so the callback used with useEsc also changes everytime. So there the listener is added/removed on each render.

2. UseEscKey is meant to be used also in other HDS-components (there are comments in other components about creating a hook)

Yes there a lot of hooks. One adds the esc listener, one creates a function that moves the focus, one that gets the selector for the element to move the focus to.

Selector has to be passed as a prop to the component, so it has to be in the context. Hook is the easiest way to get. It could be a function too (I guess), but the function would be identical.

FocusSetter is a hook, because it uses the context hook. Making it a non-hook is impossible I think. Now it also caches the callback, so document.listener is not reset on each render.

So all hooks would be same functions anyway, if they were stripped out of their "hook status".

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
```diff
@@ -0,0 +1,283 @@
...
+export const getConsentsFromCookieGroup = (groups: CookieGroup[]): ConsentList => {
+  return groups.reduce((ids, currentGroup) => {
+    currentGroup.cookies.forEach((consentData) => {
+      ids.push(consentData.id);
+    });
+    return ids;
+  }, []);
+};
+
+const forceFocusToElement = (elementSelector: string): void => {
```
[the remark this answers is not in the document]
NikoHelle: Name has "force" in it, because tabIndex is forced, if first focus() fails. It fails, if element is not a focusable element. Setting tabindex fixes this. This way the focus always succeeds.

What would exactly be gained by making this function strictly usable only with one selector/element? Only the argument would be cleared and element selector would be fetched inside the function.

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: The "setIsDomReady" inside the useEffect causes this. It is related to the React Portal and probably required by it.

I'm not sure why Portal must be used for all sites, if the Portal is just required for Gatsby? Can't it have its own component?

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: React Context is commonly used in React Apps instead of prop drilling. With context the data is in one place data and also actions that change data, like button clicks etc. Without the context, passing translations, consents and other data would require Redux, MobX or passing down objects from parent to grand-grand-grand-grand-children.

Because of context, the root component is re-rendered every time user does something. But that happens quite rarely - maybe once in 5 seconds for a fast user. Usually user just closes the modal with a single button click. Some optimization/memoizing can be applied. Mainly for the buttons, because they never change. 

One checkbox click can affect multiple components. Optimizing changes that have multiple side-effects may not be worth the effort when the component re-renders so rarely.

Re-rendering does not always update the DOM, so running functional components is not slow.

I have no experience with the Portal, but recreating that on each context change seems excessive. I would remove the portal usage to another component. if possible.

packages/react/src/components/cookieConsent/cookieConsentController.ts
```diff
@@ -0,0 +1,239 @@
...
+
+function convertStringArrayToKeyConsentObject(array: string[]): ConsentObject {
+  return array.reduce((current, key) => {
+    // eslint-disable-next-line no-param-reassign
+    current[key] = false;
+    return current;
+  }, {} as ConsentObject);
+}
+
+function mergeConsents(set1: ConsentObject, set2: ConsentObject, set3?: ConsentObject): ConsentObject {
```
VilleMiekkoja, line 46: We might be able to improve the function order in this file. It seems the exported and regular functions are in mixed order currently.

packages/react/src/components/cookieConsent/cookieModal/CookieModal.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+  describe('Cookie consent ', () => {
+    it('and child components are rendered when consents have not been handled', () => {
+      const result = renderCookieConsent(defaultConsentData);
+      verifyElementExistsByTestId(result, dataTestIds.container);
+      verifyElementExistsByTestId(result, dataTestIds.languageSwitcher);
+      verifyElementExistsByTestId(result, dataTestIds.approveButton);
+      verifyElementExistsByTestId(result, dataTestIds.approveRequiredButton);
+    });
+
+    it('is rendered if a required consent has not been approved. It could have been optional before', () => {
```
VilleMiekkoja, line 120: If there's only required cookies, then I thought it was not necessary to show the cookie consent modal then?

packages/react/src/components/cookieConsent/useCookies.ts
```diff
@@ -0,0 +1,28 @@
+import { useMemo } from 'react';
+
+import { ConsentObject, COOKIE_NAME, getCookieDomainFromUrl, parseConsents } from './cookieConsentController';
+import { createCookieController } from './cookieController';
+
+type ReturnType = {
```
VilleMiekkoja, line 6: I wonder if we would wish to export this ReturnType? Would the users of the `useCookie` hook benefit from getting the ReturnType? If we do export it, we probably need to give it a more specific name.

packages/react/src/components/cookieConsent/modal/Modal.tsx
VilleMiekkoja: Good points! 

> One checkbox click can affect multiple components. Optimizing changes that have multiple side-effects may not be worth the effort when the component re-renders so rarely.

Do you mean components that are outside of the modal? They could use the cookie consent context, and yeah that way clicking a checkbox could affect multiple components outside the modal. But inside the modal not that many components change after clicking the checkbox, or have I understood it wrong? But I think you are right, if it's hard to optimize it might not be worth it 🤔 

> Re-rendering does not always update the DOM, so running functional components is not slow.

True, but it still will run unnecessary javascript.

packages/react/src/components/cookieConsent/modal/Modal.tsx
VilleMiekkoja: Yeah, this is probably by design.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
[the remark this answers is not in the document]
VilleMiekkoja: The `tabindex` should only matter on whether the element is reachable via sequential keyboard navigation. You should be always able to focus element through javascript: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

> What would exactly be gained by making this function strictly usable only with one selector/element? Only the argument would be cleared and element selector would be fetched inside the function.

It would reveal more information to the reader of the code, without having to look for the implementation. Currently from the api user thinks that this is a general `foruceFocusToElement` function, which implies that there are multiple use cases for it. We could then rename it to be more revealing of the intented usage. Just a minor readability improvement imo, but nothing too fancy so if you think it's better this way, we can leave it as it is.

VilleMiekkoja:
CookieConsent is missing from hds documentation components overview page.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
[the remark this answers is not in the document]
NikoHelle: "You should be always able to focus element through javascript". What do you mean?

`document.querySelector("div").focus()`does nothing, if div has no tabindex set.

We can remove the failsafe and just silently fail, if the implementing developer fails to provide a selector for an element which can be focused. Usually a heading is focused after the modal is closed, and <h1> are not focusable by default.

site/src/docs/components/cookie-consent/index.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/index.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/index.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/index.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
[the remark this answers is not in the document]
VilleMiekkoja: > "You should be always able to focus element through javascript". What do you mean?

You're right, it does seem require the tabindex 👍

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/code.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Improved

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Added information about optionality

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Improved

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: Fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: fixed

site/src/docs/components/cookie-consent/api.mdx
VilleMiekkoja: fixed

VilleMiekkoja:
> CookieConsent is missing from hds documentation components overview page.

Fixed

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+      if (!containerElement) {
+        const htmlContainer = document.createElement('div');
+        htmlContainer.setAttribute('id', containerId);
+        htmlContainer.setAttribute('data-testid', 'html-cookie-consent-container');
+        document.body.insertBefore(htmlContainer, document.body.firstChild);
+      }
+      setIsDomReady(true);
+    } else if (!shouldShowModal && isDomReady) {
+      if (containerElement) {
+        document.body.removeChild(containerElement);
```
NikoHelle, line 42: Just noticed that destroying an element directly from DOM does not call unmount listeners created in React.components. Meaning functions returned from useEffect() won't be called.

Seems like Portal should use useLayoutEffect() to make components work properly.

Would be nice to make a parent component which is the portal and it adds the modal.

site/src/docs/patterns/cookies/basics.mdx
VilleMiekkoja: Fixed

site/src/docs/patterns/cookies/basics.mdx
VilleMiekkoja: Fixed

site/src/docs/patterns/cookies/common-helsinki-cookies.mdx
VilleMiekkoja: We agreed we would start with this amount and then increase in the future.

site/src/docs/patterns/cookies/basics.mdx
VilleMiekkoja: Added CookieModal/CookiePage to the header

site/src/docs/patterns/cookies/basics.mdx
VilleMiekkoja: Cleared

packages/react/src/components/cookieConsent/CookieConsent.module.scss
VilleMiekkoja: Fixed

site/src/docs/patterns/cookies/common-helsinki-cookies.mdx
VilleMiekkoja: Fixed

packages/react/src/components/cookieConsent/modal/Modal.tsx
VilleMiekkoja: Very good catch! This needs to be fixed somehow. Now if you close the modal, and press esc, the heading element outside the modal will still be focused.

VilleMiekkoja: > Just noticed that destroying an element directly from DOM does not call unmount listeners created in React.components. Meaning functions returned from useEffect() won't be called.

This is true, but we are lucky, because when the modal is closed, this Modal component will render the screen notification before the element is removed from DOM. Which means the removal of event listeners from useEffect  inside the `Content` component will be called.

But there was a real bug here, because this `Modal` component was never unmounted in our examples. Which led to a situation where the event listeners set in Modal's own useEffect were never removed. This is now fixed with one if-clause

site/src/docs/patterns/cookies/using-the-cookieconsent.mdx
NikoHelle: Maybe it is enough just to say: "current ones changed" -> "ids/names of the current ones are changed"

packages/react/src/components/cookieConsent/cookieModal/CookieModal.test.tsx
NikoHelle: In this case the test data has optional cookies. Should clarify that, or change test case where only required cookies and one is false.

packages/react/src/components/cookieConsent/useCookies.ts
NikoHelle: The returned type should be exposed to the IDE without exporting it. But might be good to export it if devs pass the prop somewhere.

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: I mean clicking a group's checkbox will affect checkboxes in its child consents and their checkboxes. So one click must update state of the child components too. Splitting context will solve that, but still the state of all checkboxes must be known to the controller which writes the cookies. Optimizing state is very hard in this kind of component where all children affect the "root state"

Adding React.Memo to optional consents and its every consent group can prevent some re-rendering.  Required consents cannot be changed, so React.Memo can be used there too.

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: I think this is fixed in the portal PR.

packages/react/src/components/cookieConsent/CookieConsentContext.tsx
NikoHelle: Changed to getApprovalPercentageForOptional

packages/react/src/components/cookieConsent/useCookies.ts
NikoHelle: Exported as UseCookiesHookReturnType

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: document.body.removeChild will throw an error when there is an existing container element which is not child of document.body. Fixed in Portal PR.

NikoHelle: The Portal PR also removes all keyup listeners

packages/react/src/components/cookieConsent/modal/Modal.tsx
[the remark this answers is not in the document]
NikoHelle: Just tried this. If component uses React.context memoizing is useless, because it will re-render when context changes.

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: Tried a bit of memoizing on some of the components. But... User actions update the context. Context updates causes components to re-render. So, memoization is useless unless context is removed from the render process.

What could be done:
- split texts in to their own context which will change only when language is changed. This won't actually decrease re-renderings, because about all components which use texts also use other data from the context. But anyway some optimisation is gained.
- memoize all functions which use consentController. ConsentController itself is memoized, so functions calling it can also be memoized. This way some components can be memoized. For example buttons.
- split components into two: a part which uses context and a part which only creates elements . Latter part might be memoizable.

hannu-kivimaki:
cookie-consent-screenshots

I reviewed the cookie consent modal design from Storybook. Seems that Show/Hide cookie consent button is a little bit off to the left (see image attached). The collapse icons left edge should be aligned even with the rest of the stuff.

Also, the heading typography seems to be from the HDS version 1.X. New headings use Medium weight instead of Bold.

hannu-kivimaki (changes requested):
cookie-consent-screenshots

I reviewed the cookie consent modal design from Storybook. Seems that the "Show/Hide cookie consent" button is a little bit off to the left (see image attached). The collapse icon's left edge should be aligned even with the rest of the stuff.

Also, the heading typography seems to be from the HDS version 1.X. New headings use Medium weight instead of Bold.

NikoHelle:
> cookie-consent-screenshots
> 
> I reviewed the cookie consent modal design from Storybook. Seems that the "Show/Hide cookie consent" button is a little bit off to the left (see image attached). The collapse icon's left edge should be aligned even with the rest of the stuff.
> 
> Also, the heading typography seems to be from the HDS version 1.X. New headings use Medium weight instead of Bold.

What browser is that screenshot from?

Chrome/Mac:

image

site/src/docs/components/cookie-consent/api.mdx
NikoHelle: This is good. :)

NikoHelle: Ok - as long as the next sentence about subdomains is kept.

site/src/docs/components/cookie-consent/api.mdx
NikoHelle: Yeah, good to know the order of function calls.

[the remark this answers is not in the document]
NikoHelle: Replace with: "If true, the modal will not be shown."

site/src/docs/patterns/cookies/basics.mdx
NikoHelle: Optional cookies **could be** changed to required or new cookies **could be** added, but that would require re-rendering the cookie component with new props. This is quite hard to implement currently, but not impossible. Probably should not be documented. :)

site/src/docs/patterns/cookies/using-the-cookieconsent.mdx
NikoHelle: Sorry, the "id" of the cookie must be changed to show the modal again. Changing the "name" has no effect as it is just text in the UI.

packages/react/src/components/cookieConsent/cookieModal/CookieModal.test.tsx
NikoHelle: CookieModal.test.tsx is removed in another PR. Replaced by PortalModal

hannu-kivimaki (approved):

minevala (commented):
Added small Grammarly issues

site/src/docs/components/cookie-consent/tabs.mdx
```diff
@@ -0,0 +1,34 @@
...
+# CookieConsent
+
+<StatusLabel type="alert">WIP</StatusLabel>
+
+<LeadParagraph>
+  The cookie compliance component informs users about cookie usage. This banner is shown when they visit a website or an application for the first time.
+</LeadParagraph>
+
+<Notification label="The component is work in progress!" type="alert" className="siteNotification">
+  This component is work in progress. Breaking changes are possible and the final accessibility audit has not been done. It has been released for early users and to gather feedback.
```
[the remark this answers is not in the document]
NikoHelle: Koko work in progress -block pitäisi kai poistaa?

packages/react/src/components/cookieConsent/modal/Modal.tsx
NikoHelle: Outdated. Modal has no domready anymore

packages/react/src/components/cookieConsent/modal/Modal.tsx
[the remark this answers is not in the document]
NikoHelle: Memoizing is done now after contexts was split.

site/src/docs/components/cookie-consent/tabs.mdx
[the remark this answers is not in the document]
minevala: Aa, niin joo. Voihan sen poistaa

site/src/docs/components/cookie-consent/index.mdx
```diff
@@ -0,0 +1,39 @@
...
+---
+
+import { withPrefix } from 'gatsby';
+import TabsLayout from './tabs.mdx';
+
+export default ({ children, pageContext }) => <TabsLayout pageContext={pageContext}>{children}</TabsLayout>;
+
+## Usage
+
+This documentation page is about HDS CookieCompliance component. If you are looking for documentation about cookies in general, please refer to [HDS cookie pattern documentation page](/patterns/cookies/basics).
```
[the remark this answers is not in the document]
NikoHelle: Cannot resolve this from the button below...

PR #795, opened 2022-07-13
Feature/whole page is viewable when cookie modal is open

minevala:
## Description
- Add placeholder for CookieConsent modal in bottom of the page

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1345

## Motivation and Context
- Now the user is able to see the whole page (including foooter) when Cookie Consent modal is visible

## How Has This Been Tested?
- Locally on dev machine

[Demo](https://city-of-helsinki.github.io/hds-demo/cookie-consent-placeholder/)

### DIFF
--- a/packages/react/src/components/cookieConsent/modal/Modal.tsx
+++ b/packages/react/src/components/cookieConsent/modal/Modal.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
 import ReactDOM from 'react-dom';
 import { VisuallyHidden } from '@react-aria/visually-hidden';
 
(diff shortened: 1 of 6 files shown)


packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -19,32 +19,63 @@ export function Modal(): React.ReactElement | null {
...
   const { settingsSaved } = useCookieConsentUiTexts();
-  const containerId = 'HdsCookieConsentContainer';
+  const containerId = 'HdsCookieConsentModalContainer';
+  const placeholderId = 'HdsCookieConsentModalPlaceholder';
   const getContainerElement = (): HTMLElement | null => document.getElementById(containerId);
+  const getPlaceholderElement = (): HTMLElement | null => document.getElementById(placeholderId);
   const [isDomReady, setIsDomReady] = useState<boolean>(false);
+  const [contentHeight, setContentHeight] = useState<number>(0);
+  const modalContentRef = useRef<HTMLDivElement>();
+  const bodyBottomPaddingStyleRef = React.useRef<string>(null);
```
VilleMiekkoja, line 29: We do not seem to add this ref to any react node, or update its value? Do we need it?

packages/react/src/components/cookieConsent/modal/Modal.tsx
```diff
@@ -19,32 +19,63 @@ export function Modal(): React.ReactElement | null {
...
         document.body.removeChild(containerElement);
       }
+      if (placeholderElement) {
+        document.body.removeChild(placeholderElement);
+      }
       setIsDomReady(false);
     }
   }, [shouldShowModal, isDomReady, setIsDomReady]);
 
+  useEffect(() => {
```
VilleMiekkoja, line 72: This useEffect is probably causing additional re-renderings of the Modal component, also therefore `ReactDOM.createPortal` gets called again. In the main cookie-consent branch, in initial render, the `ReactDOM.createPortal` gets called 2 times. In this branch, it gets called 4 times in the initial render.

Also in the main branch, if you press esc when the modal is open, the `ReactDOM.createPortal` is not called. Neither if you press 'Näytä evästeasetukset'. But in this branch, if you press esc or switch focus out of the modal, then the `ReactDOM.createPortal` will be called again. I wonder if we could reduce the amount of re-renderings? Would combining the `useEffects` help?

VilleMiekkoja:
Seem to be working, nice 👍 . Added couple comments about increase of re-renderings, and one unused react ref.

PR #812, opened 2022-08-11
Feature/notification custom aria label

VilleMiekkoja:
## Description

User is able to insert a custom aria-label for the notification section element

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1352

Closes #

## Motivation and Context

Before this, the aria-label of the notification section was hard-coded as 'Notification'. This PR adds support for more languages and fixes the a11y issue that occurs when there's 2 notifications on screen at the same time (2 identical aria-labels)

https://helsinkicity.slack.com/archives/CHCV3KTHA/p1659082296491799

## How Has This Been Tested?

- Locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/notification-custom-aria-label/?path=/story/components-notification--with-custom-aria-label)


ronijaakkola:
@VilleMiekkoja This looks good! 

Remember to add this new property to the documentation as well.

VilleMiekkoja:
> @VilleMiekkoja This looks good!
> 
> Remember to add this new property to the documentation as well.

- [ ] Add the notificationAriaLabel property to documentation

packages/react/src/components/notification/Notification.tsx
```diff
@@ -64,6 +64,11 @@ type CommonProps = React.PropsWithChildren<{
    * Note: Labels are not displayed visually for small notifications, but they are still accessible to assistive technology. This could be used to help screen reader users to better understand the context of the notification.
    */
   label?: string | React.ReactNode;
+  /**
+   * The aria label of the notification region
```
minevala, line 68: I guess we are using two ways to write aria-label 
I think I have used both of them 😁

minevala (approved):
Nice improvement 👍

packages/react/src/components/notification/Notification.tsx
VilleMiekkoja: Maybe we go with aria-label? I can change this:

- [ ] Change aria label -> aria-label

PR #814, opened 2022-08-22
Feature/cookie consent split contexts

NikoHelle:
## Description

Optimise rendering by splitting context into two:

ContentContext holds ui texts and callbacks. Props which do not change, unless language is changed.

ConsentContext holds consents related to cookies (stored in ContentContext) and functions to change consents.

Consents change when user makes selections.

Many functions are now also memoized with useMemo/useCallback so those remain unchanged when contexts change. This way component memoization can be done with React.Memo

## Motivation and Context

Solves some useless re-renderings.


packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
```diff
@@ -1,12 +1,8 @@
 import React, { useState } from 'react';
 
-import {
-  Category,
-  CookieData,
-  Provider as CookieContextProvider,
-  SupportedLanguage,
-  useCookieConsentContext,
-} from './CookieConsentContext';
+import { useConsentContext } from './contexts/ConsentContext';
```
VilleMiekkoja, line 3: Since this useConsentContext is globally exported variable, I think it should probably include the word __cookie__ in it for enough context for the user

packages/react/src/components/cookieConsent/CookieConsent.stories.tsx
```diff
@@ -1,12 +1,8 @@
...
+import { Category, CookieData, SupportedLanguage, useContentContext } from './contexts/ContentContext';
+import { Context } from './contexts/ContextComponent';
```
VilleMiekkoja, line 5: This `Context` is probably going to be also a globally exported function? Therefore I think the name requires more context to it. I would suggest we add the word _Cookie_ in it

NikoHelle:
Did not add "CookieConsent" to every type in the folder. Can't see when those would be used outside CookieConsent.

PR #815, opened 2022-08-26
Feature/react pagination

VilleMiekkoja:
## Description

Implements Pagination component for hds-react

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1259

## How Has This Been Tested?

- Locally on dev machine
- Automated unit tests

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/react-pagination/?path=/story/components-pagination--basic)

image
image
image


packages/react/src/components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+      pageCount={25}
+      pageHref={() => '#'}
+      pageIndex={pageIndex}
+      paginationAriaLabel="Pagination"
+      siblingCount={1}
+    />
+  );
+};
+
+InteractableExample.storyName = 'Interactable example';
```
minevala, line 102: Should all the examples be interactable?

packages/core/src/components/pagination/pagination.css
```diff
@@ -86,7 +86,7 @@ we can not use this position for the actual button as it's in incorrect source o
 
 .hds-pagination__item-ellipsis {
   height: var(--link-height);
-  width: 24px;
+  width: 40px;
```
minevala, line 89: This changes the outlook a bit. Is it a layout fix? 
I guess we have considered similar changes as a breaking change

packages/react/src/components/index.ts
```diff
@@ -20,20 +20,21 @@ export * from './logo';
 export * from './navigation';
 export * from './notification';
 export * from './numberInput';
+export * from './pagination';
```
minevala, line 23: Finally in alphabetic order 👍

packages/react/src/components/pagination/Pagination.test.tsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { axe } from 'jest-axe';
+
+import { Pagination, createPaginationItemList } from './Pagination';
+
+describe('createPaginationItemList', () => {
```
minevala, line 7: Would it possible to test these using the public component API?

packages/react/src/components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+  title: 'Components/Pagination',
+  parameters: {
+    controls: { expanded: true },
+  },
+  args: {},
+};
+
+// args is required for docs tab to show source code
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
+export const Basic = (args) => (
```
minevala, line 16: It would be nice if the control tab's args would work, at least on some level

minevala (commented):
Nice work 👍 
A few comments

packages/core/src/components/pagination/pagination.css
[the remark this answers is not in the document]
minevala: I guess we could decide when it is not a breaking change. I would say it is not a breaking change when the change is to a new component or it fixes the layout somehow.

packages/react/src/components/pagination/Pagination.stories.tsx
VilleMiekkoja: Good idea, I see no reason why not 👍

packages/react/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,250 @@
...
+  pageIndex: number;
+  /**
+   * Aria-label for the pagination nav element
+   */
+  paginationAriaLabel: string;
+  /**
+   * Number of always visible pages before and after the current page
+   * @default 1
+   */
+  siblingCount?: number;
```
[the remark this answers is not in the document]
minevala: Let's go with siblingCount since it is used in other libraries as well, for example in Material UI

packages/core/src/components/pagination/pagination.css
minevala: Not a breaking change since it is for the new component and fixes the layout

packages/react/src/components/pagination/Pagination.tsx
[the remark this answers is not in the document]
VilleMiekkoja: We had a discussion about this and agreed that the term is not easy to understand, but is a one that is commonly used in other paginations as well, so we decided to leave it like this.

packages/core/src/components/pagination/pagination.css
VilleMiekkoja: We had a discussion about this and decided that we would not consider it a breaking change this time.

packages/react/src/components/pagination/Pagination.test.tsx
VilleMiekkoja: It should be possible:

- [ ] Test pagination item list using the public component API

packages/react/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  return pageTitle[language];
+};
+
+const range = (start, end) => {
+  const length = end - start + 1;
+  return Array.from({ length }, (_, i) => start + i);
+};
+
+export const createPaginationItemList = ({ pageCount, pageIndex, siblingCount }: Partial<PaginationProps>) => {
```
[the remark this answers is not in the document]
VilleMiekkoja: Good addition:

- [ ] Add typings for the ellipsis and the function output

packages/react/src/components/pagination/Pagination.stories.tsx
VilleMiekkoja: This is hard because then the props do not show up in the storybook's story tab. Let's try to create at least one Playground story where the controls work:

- [ ] Create Playground story where storybook's controls work

packages/react/src/components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+      hideNextButton={args.hideNextButton}
+      hidePrevButton={args.hidePrevButton}
+      language={args.language}
+      onChange={(event, index) => {
+        event.preventDefault();
+        setPageIndex(index);
+      }}
+      pageCount={args.pageCount}
+      pageHref={() => '#'}
+      pageIndex={pageIndex}
```
minevala, line 148: Setting pageIndex through controls do not work at the moment

packages/react/src/components/pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+      paginationAriaLabel={args.paginationAriaLabel}
+      siblingCount={args.siblingCount}
```
minevala, line 150: Would it be possible to add theme also here? So it could be tested in the Playground

minevala (commented):
Good improvements 👍 
A small suggestion to Playground component

packages/react/src/components/pagination/Pagination.stories.tsx
VilleMiekkoja: Good idea:

- [ ] Add theme to playground story

packages/react/src/components/pagination/Pagination.stories.tsx
minevala: Lets skip this since it breaks the buttons when it is initially set

packages/react/src/components/pagination/Pagination.test.tsx
```diff
@@ -132,4 +132,46 @@ describe('<Pagination /> spec', () => {
...
+        pageCount={25}
+        pageIndex={3}
+        pageHref={() => '#'}
+        language="en"
+        siblingCount={1}
+      />,
+    );
+    expect(queryByText('Opened page 4')).not.toBeInTheDocument();
+
+    rerender(
```
minevala, line 150: Would this work with a user click?

VilleMiekkoja: The user click handling is handled by the user of the component. So the test would only work on specific case, and it would basically test the same thing, as changing the prop is the trigger for the notification.

minevala: Yes true 👍

minevala (approved):
Nice work 👍 A new component 🎉

minevala (approved):

PR #816, opened 2022-08-31
Feature/Add callback onSort to Table component to support sort side effects

VilleMiekkoja:
## Description

Add callback onSort to Table component to support sorting related side effects, for example sorting in backend. Tilavarauspalvelu had a need for this.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1357

## How Has This Been Tested?

- Locally on dev machine
- Automated unit tests

### DIFF
--- a/packages/react/src/components/table/Table.test.tsx
+++ b/packages/react/src/components/table/Table.test.tsx
@@ -137,21 +137,6 @@ describe('<Table /> spec', () => {
     userEvent.click(container.querySelector('[data-testid="hds-table-sorting-header-age"]'));
 
     expect(mockOnSort).toHaveBeenCalledTimes(1);
-    expect(mockOnSort).toHaveBeenCalledWith(
-      expect.objectContaining({
-        direction: 'asc',
-        colKey: 'age',
-      }),
-    );
-
-    userEvent.click(container.querySelector('[data-testid="hds-table-sorting-header-age"]'));
-    expect(mockOnSort).toHaveBeenCalledTimes(2);
-    expect(mockOnSort).toHaveBeenCalledWith(
-      expect.objectContaining({
-        direction: 'desc',
-        colKey: 'age',
-      }),
-    );
   });
 
   it('Should successfully select and deselect a single row', () => {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/table/Table.test.tsx
```diff
@@ -102,6 +102,43 @@ describe('<Table /> spec', () => {
...
+        rows={rows}
+        indexKey={indexKey}
+        onSort={mockOnSort}
+        renderIndexCol={renderIndexCol}
+      />,
+    );
+
+    userEvent.click(container.querySelector('[data-testid="hds-table-sorting-header-age"]'));
+
+    expect(mockOnSort).toHaveBeenCalledTimes(1);
```
minevala, line 139: This could be tested without mock by:
1. Providing initial rows, for example, names could be D, E, F
2. Providing a custom onSort which replaces those initial rows with rows from "server" that have names A, B, C
3. User clicks name sort
4. Asserting that the server rows are visible or have replaced the initial rows

packages/react/src/components/table/Table.tsx
```diff
@@ -138,6 +138,13 @@ export type TableProps = React.ComponentPropsWithoutRef<'table'> & {
    * Sorting order applied for initial sorting.
    */
   initialSortingOrder?: 'asc' | 'desc';
+  /**
+   * Function for custom sorting or providing side effects before or after the sort. If you want to use the component's own sorting function, call the handleSort callback function.
+   * @param order
+   * @param colKey
+   * @param handleSort
+   */
+  onSort?: (order: 'asc' | 'desc', colKey: string, handleSort: () => void) => void;
```
VilleMiekkoja, line 147: Remember to add this new prop to documentation site code tab props.

minevala (commented):
Nice work 👍 
One suggestion for testing but nothing mandatory

packages/react/src/components/table/Table.test.tsx
harriplappalainen: I really like this as we talked just this morning about avoiding mocking in tests but should we make this test in a separate branch after the release? Maybe look for other mocks and refactor them as well.

packages/react/src/components/table/Table.tsx
harriplappalainen: Done. I put the same description as in the component prop doc.

minevala (approved):
Nice addition 👍

PR #820, opened 2022-09-06
Show notification after successful or failing Playground code copy

minevala:
## Description
- Show notification when user copies code using Documentation Playground component's "copy code"-button
- This functionality requires that Notification accessibility screen reader issue is fixed, see related issue HDS-1375

Depends on #824 

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1281
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1375

## Motivation and Context
- This improves Documentation's accessibility when user gets feedback after a button click 

## How Has This Been Tested?
- locally on dev machine

[Demo](https://city-of-helsinki.github.io/hds-demo/playground-copy-notification/components/buttons/code)

### DIFF
--- a/site/src/components/Playground.js
+++ b/site/src/components/Playground.js
@@ -135,15 +135,14 @@ const Editor = ({ onChange, initialCode, code, language }) => {
     [getTextArea],
   );
 
-  const copy = () => {
+  const copy = async () => {
     if (viewPortRef.current && copyButtonRef.current) {
       const textArea = getTextArea(viewPortRef.current);
       textArea.focus();
       textArea.select();
       setCopyState('');
-
       try {
-        document.execCommand('copy');
+        await navigator.clipboard.writeText(textArea.value);
         if (copyButtonRef.current) {
           copyButtonRef.current.focus();
         }
--- a/site/src/components/Playground.js
+++ b/site/src/components/Playground.js
@@ -244,13 +244,8 @@ const Editor = ({ onChange, initialCode, code, language }) => {
           Reset example
         </Button>
       </div>
-      {language === 'jsx' &&
-        <LiveError />
-      }
-      {language !== 'jsx' &&
-        <LiveErrorCore code={code} />
-      }
-
+      {language === 'jsx' && <LiveError />}
+      {language !== 'jsx' && <LiveErrorCore code={code} />}
     </>
   );
 };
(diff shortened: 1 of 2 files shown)


site/src/components/Playground.js
```diff
@@ -137,14 +140,17 @@ const Editor = ({ onChange, initialCode, code, language }) => {
       const textArea = getTextArea(viewPortRef.current);
       textArea.focus();
       textArea.select();
+      setCopyState('');
 
       try {
         document.execCommand('copy');
```
VilleMiekkoja, line 146: Off-topic from the PR, Intellij IDEA warns about document.execCommandy being deprecated:

image

site/src/components/Playground.js
```diff
@@ -207,6 +213,27 @@ const Editor = ({ onChange, initialCode, code, language }) => {
...
+            onClose={() => setCopyState('')}
+          />
+        )}
+        {copyState === copyErrorState && (
+          <Notification
+            type="error"
+            label="Code copy failed."
+            position="bottom-right"
+            displayAutoCloseProgress={false}
+            dismissible
```
VilleMiekkoja, line 232: Is there a reason why the notifications differ in case of success vs error? I mean the error has the close button, and it won't automatically close?

minevala: I think error messages are something we require users to see because the copy failed, and they either have to try it again or try to copy the code in some other way. If they autoclose, users might miss them in some cases.
I would rather remove the autoclose from success message, but then it would require two clicks to continue when everything went smoothly.

site/src/components/Playground.js
minevala: I'll check if it would be an easy to change 👍

site/src/components/Playground.js
minevala: I think the error message could suggest to manually copy the code once the button functionality fails 🤔

harriplappalainen: Now I think the success and error texts are too heavy since it's in bold 😅 It looks a bit tacky in my opinion. I fiddled with it and I think it looks better with shorter label and using the children prop for more information. Try these ones for size and see what you think:
- on success:`label="Code copied" children={<span>Example code was copied to clipboard.</span>}`
- on error: `label="Code copying failed" children={<span>Try again or copy the code using the clipboard.</span>}`

minevala: - [x] update notification texts

harriplappalainen (approved):
This is a good addition! 🏅 I approve as well.

PR #824, opened 2022-09-12
Fix Notification content readability in screen readers

minevala:
## Description
- Replace aria-atomic with role=alert and remove heading wai-aria attributes in toast and invisible notifications. 
- Use the above convention in the Core examples
- Set focus to Notification toggle button in toast examples
- Add accessibility info about focus handling into documentation
- Remove close button from autoclose toast examples

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1375

## Motivation and Context
- Some screen reader and browser combinations do not read toast notification content at the moment. It might be fixed by moving the role-attribute into the same element as the area-atomic attribute.

## How Has This Been Tested?
- Locally using VoiceOver. The VoiceOver reads the toast's notification title which is not read in the current production version (this is also an issue with some screen reader and browser combinations)
- Component snapshot test

[Documentation Demo](https://city-of-helsinki.github.io/hds-demo/fix-notification-accessibility-site/components/notification)

[React Demo](https://city-of-helsinki.github.io/hds-demo/fix-notification-accessibility/iframe.html?id=components-notification--playground&args=position:top-center&viewMode=story)

[Core Notfications Demo](https://city-of-helsinki.github.io/hds-demo/fix-notification-accessibility-core/?path=/story/components-notification--default)


harriplappalainen:
Using Mac's VoiceOver, the React component's aria labels were read beautifully and automatically on Chrome but not on Safari or Firefox. Safari and Firefox focus first on the Toggle button but Chrome started from the notification. Notification is also not traversable to by tab button but it can be navigated to by Ctrl+Option+Left/Right arrow key.

minevala:
> Using Mac's VoiceOver, the React component's aria labels were read beautifully and automatically on Chrome but not on Safari or Firefox, I had to navigate to the notification myself. Safari and Firefox focus first on the Toggle button but Chrome started from the notification. Notification is also not traversable to by the tab button but it can be navigated to by Ctrl+Option+Left/Right arrow key. Are these problems? I'm not sure. It would be nice if tab would work though, right? Or is it enough if navigation works at least by ctr-option-arrow keys?

Good find 👍 I'm not sure, I think it should navigate to Notification automatically

VilleMiekkoja (approved):
Good job! 🥇

harriplappalainen:
Ship it! :shipit:

PR #826, opened 2022-09-14
Element ids for NumberInputs in site documentation

harriplappalainen:
## Description

Add element ids for NumberInputs in site documentation in component Usage and Code pages for accessibility.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1376

## How Has This Been Tested?

Locally on developer's machine

### DIFF
--- a/site/src/docs/components/number-input/code.mdx
+++ b/site/src/docs/components/number-input/code.mdx
@@ -11,39 +11,39 @@ export default ({ children, pageContext }) => <TabsLayout pageContext={pageConte
 
 ## Code
 
-### Code examples 
+### Code examples
 
 #### Default
-<Playground>
 
+<Playground>
 
 ```jsx
 <>
-<NumberInput
-  helperText="Assistive text"
-  label="Total compensation"
-  unit="€"
-  defaultValue={1000}
-  style={{ maxWidth: '320px' }}
-/>
+  <NumberInput
+    id="input"
+    helperText="Assistive text"
+    label="Total compensation"
+    unit="€"
+    defaultValue={1000}
+    style={{ maxWidth: '320px' }}
+  />
 
-<NumberInput
-  disabled
-  helperText="Assistive text"
-  label="Total compensation"
-  unit="€"
-  defaultValue={1000}
-  style={{ maxWidth: '320px', marginTop: 'var(--spacing-s)' }}
-/>
+  <NumberInput
+    disabled
+    id="input"
+    helperText="Assistive text"
+    label="Total compensation"
+    unit="€"
+    defaultValue={1000}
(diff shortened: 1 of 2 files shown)


site/src/docs/components/number-input/code.mdx
```diff
@@ -11,51 +11,56 @@ export default ({ children, pageContext }) => <TabsLayout pageContext={pageConte
...
-<NumberInput
-  disabled
-  helperText="Assistive text"
-  label="Total compensation"
-  unit="€"
-  defaultValue={1000}
-  style={{ maxWidth: '320px', marginTop: 'var(--spacing-s)' }}
-/>
+  <NumberInput
+    id="input"
```
VilleMiekkoja, line 23: Looking good! Small improvement idea: The id probably needs to be unique among one html page. Maybe we could give a name that maps with the titles? So this would be `default`, next could be `disabled` etc.

harriplappalainen: Sure, let's do it. I wondered about doing it that way but noticed that the core examples had same ids. Do I change those as well?

minevala: They could be fixed also. One weird behaviour for same ids is the focus jumping when user clicks the input label.

harriplappalainen: Requested fix has been pushed! Check if this is ok.

minevala: I think it works now 👍 
They do not even have to be different between React or Core tabs because they are never in the DOM at the same time. 
I was thinking about how realistic some of the ids are, eq. default. Do users realize that they have to replace it with their custom id? Maybe they will 🤔

harriplappalainen: It's sort of a silly name for an id, not very realistic, but it relates to the context in documentation. I think the id names are fine but it wouldn't be a bad idea to include accessibility information about label for input id somewhere? At a quick glance I didn't find anything about it in NumberInput or TextField component documentation or even at Patterns/Form building section.

minevala:
> Github claims theres 333 commits in this pr, which seems weird? I wonder if something has gone wrong with the rebase?
> 
> image

I thin it has done a merge

VilleMiekkoja:
> > Github claims theres 333 commits in this pr, which seems weird? I wonder if something has gone wrong with the rebase?
> > image
> 
> I think it has done a merge

Something is not right. Github also shows that 97 files has been changed. Github should be able to show the difference to master only 🤔

VilleMiekkoja (approved):
Good job with adding the ids ✌️

minevala (approved):
Nice fix 👍

PR #832, opened 2022-09-19
Analytics script for finding usage of hds-react versions

harriplappalainen:
## Description

Bash script for calling Github API and fetching information about repositories under City of Helsinki organisation that use hds-react package. Saves a JSON file in the same directory as the script was run.

Updated Confluence as well: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HDS/pages/8028225540/HDS+React+version+usage

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1369

## How Has This Been Tested?
Running the script locally. You need a Github account and Personal Access Token to run it. The script uses an environment variable called `GITHUB_TOKEN`. You can add that to your environment variable or just add your PAT into the script code.

Screenshot of result file: 
Screenshot 2022-09-19 at 12 34 45

### DIFF
--- a/analytics/README.md
+++ b/analytics/README.md
@@ -29,9 +29,24 @@ To collect analytics from all found repositories, run:
 The bash script file `find-hds-react-version-usage.sh` calls Github API and finds all `hds-react` occurrences in code under the City of Helsinki organisation and gets the corresponding repository name, Github URL and hds-react version and puts them in a JSON file. This could possibly be merged into the analytics scripts to provide even more refined analytics data.
 
 ### How to use it
-You need bash and a Github account.
 
+Requirements for running the script:
+- Github account
+- Personal Access Token (PAT for short) for Github
+- [brew](https://brew.sh/)
+- [jq](https://stedolan.github.io/jq/) - for JSON handling
+- sponge library - for file management
+
+#### Github Personal Access Token
 The script uses an environment variable called `GITHUB_TOKEN` which is Personal Access Token (PAT for short) which you can obtain from your [Github settings](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token). If you can't be bothered, you can replace the `GITHUB_TOKEN` in the script with your PAT.
+The PAT token requires the following permissions: repo, admin:public_key and project.
+
+#### Brew, jq and sponge
+Install brew for MacOS and Linux with this command: `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"``
+
+jq: `brew install jq`
+
+sponge: `brew install sponge`
 
 Run the script in the terminal with `bash find-hds-react-version-usage.sh`
 
--- a/analytics/README.md
+++ b/analytics/README.md
@@ -28,7 +28,7 @@ To collect analytics from a subset of found repositories (for testing purposes),
 To collect analytics from all found repositories, run:
 ```TOKEN=$GITHUB_TOKEN node run-analytics.js```
 
-## Bonus content: Search which City of Helsinki Github repositories use hds-react and what version
+## HDS version usage analytics: collect data of HDS versions in repos
 The bash script file `find-hds-react-version-usage.sh` calls Github API and finds all `hds-react` occurrences in code under the City of Helsinki organisation and gets the corresponding repository name, Github URL, path to the package.json file and hds-react version and puts them in a JSON file.
 
 ### How to use it
@@ -52,6 +52,6 @@ jq: `brew install jq`
 sponge: `brew install sponge`
 
 ### Running the script
-Run the script in the terminal with `bash find-hds-react-version-usage.sh`
+Run the script in the terminal with `TOKEN=$GITHUB_TOKEN bash find-hds-react-version-usage.sh`
 
 The script produces a `hds-react-version-usage.json` file which contains an array of objects with the name and Github URL of the repository, which hds-react version it uses and the path to the package.json file where hds-react was mentioned. The path is useful information to determine if subfolders of a repository use the same hds-react version.
--- a/analytics/README.md
+++ b/analytics/README.md
@@ -15,6 +15,7 @@ It will strain the network and might exceed some Github rate limits. Running scr
 
 ### Prerequisites
 - [Node](https://nodejs.org/en/)
+- [Github cli](https://cli.github.com/)
 - Github account
 - [Github PAT for authentication](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
 - [React-scanner](https://github.com/moroshko/react-scanner) as globally or locally installed
--- a/analytics/README.md
+++ b/analytics/README.md
@@ -26,7 +26,7 @@ To collect analytics from all found repositories, run:
 ```node run-analytics.js```
 
 ## Bonus content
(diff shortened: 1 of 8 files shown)


harriplappalainen:
@minevala found out that the result set is much smaller than what the search on Github pages gives. I will see what that's about.

minevala (commented):
Nice bash magic 🪄 
I think it would be nice to merge this into existing analytic scripts

analytics/README.md
```diff
@@ -24,3 +24,15 @@ To collect analytics from a subset of found repositories (for testing purposes),
...
 To collect analytics from all found repositories, run:
 ```node run-analytics.js```
+
+## Bonus content
+The bash script file `find-hds-react-version-usage.sh` calls Github API and finds all `hds-react` occurrences in code under the City of Helsinki organisation and gets the corresponding repository name, Github URL and hds-react version and puts them in a JSON file. This could possibly be merged into the analytics scripts to provide even more refined analytics data.
+
+### How to use it
+You need bash and a Github account.
+
+The script uses an environment variable called `GITHUB_TOKEN` which is Personal Access Token (PAT for short) which you can obtain from your [Github settings](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token). If you can't be bothered, you can replace the `GITHUB_TOKEN` in the script with your PAT.
```
VilleMiekkoja, line 34: We could maybe add that what scopes are required for the personal access token:

image

analytics/find-hds-react-version-usage.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Call to Github API to find all occurrences of "hds-react" under City-of-Helsinki org.
+curl -H "Accept: application/vnd.github+json"  -H "Authorization: token $GITHUB_TOKEN" "https://api.github.com/search/code?q=hds-react+in:file+filename:package.json+org:City-of-Helsinki" | jq '[.items[] | {name: .repository.name, html_url: .repository.html_url, package_url: .git_url}]' > hds-react-version-usage.json
```
VilleMiekkoja, line 3: Small improvement idea: Since the GITHUB_TOKEN is used in this line and also at line 13, we could maybe store it as a global variable in the beginning of the file if possible? Then (user who has not setup the env variable) user would only have to replace the variable to one place

analytics/README.md
```diff
@@ -24,3 +24,15 @@ To collect analytics from a subset of found repositories (for testing purposes),
...
+
+Run the script in the terminal with `bash find-hds-react-version-usage.sh`
```
VilleMiekkoja, line 36: Tried running the script, but got these errors:

> find-hds-react-version-usage.sh: line 3: jq: command not found
> curl: (23) Failed writing body (0 != 1370)
> find-hds-react-version-usage.sh: line 8: jq: command not found
> find-hds-react-version-usage.sh: line 24: jq: command not found
> find-hds-react-version-usage.sh: line 24: sponge: command not found

jq and sponge probably needs to be installed separately? We could maybe add instructions on how to install those to this readme file

analytics/README.md
harriplappalainen: A very good point, I'll add it in the comments!

harriplappalainen:
Review away! 

Interestingly enough this script and the old script return about the same number of repositories but wholly different number of statistics for component usages; for example old one returns 549 instances of Button component and new one returns 717 instances. This is because old one fetches all Typescript projects that have a .json file in them under the City of Helsinki organisation. New script has a more precise query, it returns repos that have hds-react in package.json files under the CoH org. So while the number of repos the scripts fetch are about the same, they are not quite the same results.

analytics/README.md
```diff
@@ -16,11 +16,42 @@ It will strain the network and might exceed some Github rate limits. Running scr
...
 - [Github cli](https://cli.github.com/)
+- Github account
+- [Github PAT for authentication](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
 - [React-scanner](https://github.com/moroshko/react-scanner) as globally or locally installed
 
 ### Running the script
+Set your Github PAT to environment variable called `GITHUB_TOKEN` or inject it into the command below.
 To collect analytics from a subset of found repositories (for testing purposes), run:
-```OFFSET=0 LIMIT=1 node run-analytics.js```
+```OFFSET=0 LIMIT=1 TOKEN=$GITHUB_TOKEN node run-analytics.js```
```
minevala, line 26: Should we use TOKEN or GITHUB_TOKEN everywhere?
In this script the env variable is TOKEN and in the bash script it is GITHUB_TOKEN

analytics/find-hds-react-version-usage.sh
```diff
@@ -0,0 +1,27 @@
...
+    version=$(jq '.dependencies."hds-react"' <<< "${content}")
+    # If version is null, read it from devDependencies instead
+    [[ "$version" = "null" ]] && properVersion=$(jq '.devDependencies."hds-react"' <<< "${content}") || properVersion=$(jq '.dependencies."hds-react"' <<< "${content}");
+    # Insert the key value pair "version": "xx.xx" into the hds-react-version-usage.json with sponge
+    jq --argjson properVersion "$properVersion" --argjson count "$count" '.repositories[$count] += { version: $properVersion }' hds-react-version-usage.json | sponge hds-react-version-usage.json;
+    
+    ((count++))
+done
+# Remove package helsinki-design-system from the list
+jq 'del(.repositories[] | select(.name == "helsinki-design-system"))' hds-react-version-usage.json | sponge hds-react-version-usage.json
```
minevala, line 27: It would be nice if the result would be in the same results folder as other results.
It would be also useful if the name would include date as other results.

minevala: This script does not seem to filter out duplicates 🤔 
Maybe it does not need to if the subfolders might have different versions

minevala (commented):
Nice work 🏅 
I added a few minor suggestions but the script is working and it generates nice results 👍

analytics/find-hds-react-version-usage.sh
harriplappalainen: Yes, I thought it could be useful to see all the versions inside a project if it using multiple versions. Then we could even holla at them to fix it.

analytics/find-hds-react-version-usage.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# Call to Github API to find all occurrences of "hds-react" under City-of-Helsinki org.
+curl -H "Accept: application/vnd.github+json"  -H "Authorization: token $TOKEN" "https://api.github.com/search/code?q=hds-react+in:file+filename:package.json+org:City-of-Helsinki&per_page=100" | jq '. | .total_count as $total_count | { total_count: $total_count, repositories: [.items[] | {name: .repository.name, url: .repository.html_url, package_url: .git_url, path: .path}]}' > hds-react-version-usage.json
```
minevala, line 3: Could this be combined with the curl in the search-repos script? That way it would easier to develop and fix later?

harriplappalainen:
Review away! Changes made:
- bash script JSON file is stored in results file in the same naming format (except for milliseconds, please don't demand those)
- github API call is in an own bash script
- hds-react bash script calls search-repos.js to get all the data if the results are on multiple pages
- README updated

minevala (approved):
Good job, nice analytics improvement 🏅 🥇 

I longed for some console about the steps info when I ran the bash script, for example
```searching repos..``` ```generating results..```
But we can go without them too

PR #838, opened 2022-09-26
HDS-1387: refocus cookie consent title on language change

NikoHelle:
## Description

When user changes the language, focus is lost and the modal shrinks.

Fixed by adding the modal title as a dependency to the useEffect which handles the initial focusing.

If user selects the same language again, the focus is moved back to the language selector button for better accessibility.

## How Has This Been Tested?

Added tests for both scenarios: language is changed and same language is reselected.

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/cookie-consent/?path=/story/components-cookieconsent--english-modal-version)

### DIFF
--- a/packages/react/src/components/cookieConsent/content/Content.tsx
+++ b/packages/react/src/components/cookieConsent/content/Content.tsx
@@ -32,10 +32,10 @@ export function Content(): React.ReactElement {
       );
 
   const setFocusToTitle = useCallback(() => {
-    if (titleRef.current) {
+    if (title && titleRef.current) {
       titleRef.current.focus();
     }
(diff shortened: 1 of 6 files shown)


VilleMiekkoja:
Could you add a demo of this?

NikoHelle:
> 

What kind of demo? Do you mean a story?

VilleMiekkoja:
I think it's a bit weird that it works differently depending on which language is selected. I mean the difference in selecting current language or a new language. But it works ok and we can go with this 👍

VilleMiekkoja (approved):
Nice clean fix! 🕺

PR #839, opened 2022-09-26
HDS-1385 Doc site: New visuals on component overview page

hannu-kivimaki:
## Description

- Added component overview images as pngs --> /site/static/images/components/overview/.
- Changed text-only linkboxes to linkboxes with images.
- Made the linkboxes a bit wider than the previous ones (from 204px to 280 px). The added width gives the visuals a bit more breathing room.

## Motivation and Context

Make the doc site's component overview page visually more appealing.

## How Has This Been Tested?

Locally on the designer's workstation.

updated-component-overview

### DIFF
--- a/site/src/components/LinkboxList.js
+++ b/site/src/components/LinkboxList.js
@@ -21,7 +21,7 @@ const LinkboxList = ({ data, className }) => (
             external={item.external}
             heading={item.name}
             text={item.text}
-            imgProps={item.imgProps ? { ...item.imgProps, src: withPrefix(item.imgProps.src) } : undefined}
+            imgProps={{ src: withPrefix("/images/components/overview/" + item.name + "@2x.png"), width: 280, height: 180}}
             onClick={(event) => {
               if (!item.external) {
                 event.preventDefault();
--- a/site/src/components/LinkboxList.js
+++ b/site/src/components/LinkboxList.js
@@ -21,7 +21,7 @@ const LinkboxList = ({ data, className }) => (
             external={item.external}
             heading={item.name}
             text={item.text}
-            imgProps={{ src: withPrefix("/images/components/overview/" + item.name + "@2x.png"), width: 280, height: 180}}
+            imgProps={item.imgProps ? { ...item.imgProps, src: withPrefix(item.imgProps.src) } : undefined}
             onClick={(event) => {
               if (!item.external) {
                 event.preventDefault();
--- a/site/src/components/ComponentsList.js
+++ b/site/src/components/ComponentsList.js
@@ -12,7 +12,9 @@ const ComponentsList = () => {
       href: '/components/accordion',
       imgProps: { 
         src: '/images/components/overview/accordion@2x.png', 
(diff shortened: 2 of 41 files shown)


site/src/components/LinkboxList.js
```diff
@@ -21,7 +21,7 @@ const LinkboxList = ({ data, className }) => (
             external={item.external}
             heading={item.name}
             text={item.text}
-            imgProps={item.imgProps ? { ...item.imgProps, src: withPrefix(item.imgProps.src) } : undefined}
+            imgProps={{ src: withPrefix("/images/components/overview/" + item.name + "@2x.png"), width: 280, height: 180}}
```
minevala, line 24: Maybe we could add the imageUrl in the ComponentsList component configuration? 
I think we need an alternative text for the image too

minevala (commented):
Looks really nice 👍 
I added a suggestion

site/src/components/LinkboxList.js
minevala: The format could be: 
```js
 const componentData = [
    {
      name: 'Accordion',
      text: 'Accordions can be used to hide and reveal information.',
      linkboxAriaLabel: 'Accordion component',
      linkAriaLabel: 'Go to the Accordion component page',
      href: '/components/accordion',
      imageProps: { 
         src: '/images/components/overview/Accordion@2x.png', 
         altText: "An illustration of Accordion component"
      }
    },
```

hannu-kivimaki: Thanks for pointing that out! Now I understand the structure and the dependencies of the site better.

minevala: The change might be tricky but we probably need to map the component data in ComponentsList like this:
```js
const componentDataWithImageProps = componentData.map(component => 
  ({...component, 
     imgProps: { 
       src: '/images/components/overview/" + item.name + "@2x.png', 
       alt: `An illustration of ${component.name} component`
     }
  })
)
return <LinkboxList data={componentDataWithImageProps} />;

```
And the make the LinkBoxList to accept imgProp.alt if there is one.

hannu-kivimaki: Thanks! 👍 I added imgProps to the component data in ComponentsList and restored LinkboxList back to the way it was. Now the image URLs on the front page work fine and the component overview page shows the images and their alternative texts correctly.

hannu-kivimaki: Still wondering one more thing. Should I add the height and width to imgProps in ComponentList like in FrontPageLinkList (see below)?

`{
      name: 'Components',
      text: 'UI components ready to use for complete layouts',
      linkboxAriaLabel: 'Components',
      linkAriaLabel: 'Go to the Components page',
      href: '/components',
      imgProps: {
        src: '/images/components/components.svg',
        height: 198,
        width: 300,
      },
    },`

minevala: Yes, it would make sense to add those too 👍

minevala (approved):
Nice work 👍

PR #848, opened 2022-10-06
HDS-1386: Increased yarn network-timeout

NikoHelle:
## Description

Increased yarn timeout in development.yml. Same fix was done in Profile when its builds failed with same yarn error.

## How Has This Been Tested?

Hard to test, because builds fail rarely anyway. Ran build 5 times without hiccups.


minevala (approved):
I did not manage to find this setting in the Yarn docs but according to GitHub Action's logs the timeout is configured successfully. 👍

PR #852, opened 2022-10-13
HDS-1377: number input accessibility fix

NikoHelle:
## Description

When screen reader user clicks stepper buttons, the user receives no information about the new value.
Also, an input and its control buttons should be grouped. 

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1377

## How Has This Been Tested?

Added new tests for new features.

## Demo

https://city-of-helsinki.github.io/hds-demo/numberinput-accessibility/iframe.html?id=components-numberinput--custom-step&args=&viewMode=story

### DIFF
--- a/packages/react/src/components/numberInput/NumberInput.test.tsx
+++ b/packages/react/src/components/numberInput/NumberInput.test.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, RenderResult, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
 import { act } from 'react-dom/test-utils';
@@ -14,6 +14,14 @@ const numberInputProps = {
 };
 
 describe('<NumberInput /> spec', () => {
+  const getAriaLiveElement = (result: RenderResult) => {
+    return result.container.querySelector('[aria-live="assertive"]');
+  };
+
+  const getNumberInputElement = () => {
+    return screen.getByLabelText('Test label number input', { selector: 'input' });
+  };
+
   it('renders the component', () => {
     const { asFragment } = render(<NumberInput {...numberInputProps} />);
     expect(asFragment()).toMatchSnapshot();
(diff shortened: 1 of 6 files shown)


packages/react/src/components/numberInput/NumberInput.test.tsx
```diff
@@ -95,7 +103,36 @@ describe('<NumberInput /> spec', () => {
...
+
+  it('input and buttons are grouped when component has the step-attribute', async () => {
+    const resultWithStep = render(<NumberInput defaultValue={10} step={10} {...numberInputProps} />);
+    const group = resultWithStep.container.querySelector('[role="group"]');
+    expect(group).not.toBeNull();
+
+    const resultWithoutStep = render(<NumberInput defaultValue={10} {...numberInputProps} />);
+    const groupNotFound = resultWithoutStep.container.querySelector('[role="group"]');
+    expect(groupNotFound).toBeNull();
+  });
```
harriplappalainen, line 137: Nice accessibility tests. I'm just not sure if the last one about checking for role="group" is a bit too implementation detailed test? If we or unicus decides that actually role="group" isn't good enough and we have to change them, then this will fail. Although that might not happen, at least for a good while. What do you think?

harriplappalainen:
Nice fix! Great improvement to this component's accessibility. I had one comment about one specific test and I just have one question. 

In the ticket it was mentioned that "When focusing to a button from outside the group, it should be read something like this: ”Number of attendees group, 5 chosen, Increase by one, button”.” Right now this doesn't happen or I don't know how to test this properly? 😄 I tried just by clicking the button with a mouse. Did Unicus check how this is working and it's ok for them? If so then that's good enough for me.

packages/react/src/components/numberInput/NumberInput.test.tsx
NikoHelle: Role="group" is required in this case, so it should not be removed and the test should therefore be reliable. The element where the role-attribute is added, exists in all cases (just without the group attribute), so adding a data-testid would not work

NikoHelle:
> Nice fix! Great improvement to this component's accessibility. I had one comment about one specific test and I just have one question.
> 
> In the ticket it was mentioned that "When focusing to a button from outside the group, it should be read something like this: ”Number of attendees group, 5 chosen, Increase by one, button”.” Right now this doesn't happen or I don't know how to test this properly? 😄 I tried just by clicking the button with a mouse. Did Unicus check how this is working and it's ok for them? If so then that's good enough for me.

Unicus: "Tuo on sellainen ominaisuus, joka voisi olla kiva olla, mutta toisaalta voi aiheuttaa liikaa puhetta. Tarkoitin tuolla kohdistamisella sitä, että kun käyttäjä selaa lomaketta ylöspäin, niin ensin kohdistuu painikkeet ja vasta sen jälkeen syötekenttä, jossa on arvot. Voidaan varmaan jättää tuo pois."

harriplappalainen (approved):

PR #855, opened 2022-10-14
HDS-1431 Add possibility to use external label for Select and Combobox components

VilleMiekkoja:
## Description

Add possibility to use external label for Select and Combobox components. This enables custom labels that are placed and styled differently than the components own label.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1431

## How Has This Been Tested?

- Locally in dev machine

👉 [Combobox demo](https://city-of-helsinki.github.io/hds-demo/external-label-for-dropdowns/?path=/story/components-dropdowns-combobox--with-external-label)
👉 [Select demo](https://city-of-helsinki.github.io/hds-demo/external-label-for-dropdowns/?path=/story/components-dropdowns-select--with-external-label)

image

### DIFF
--- a/packages/react/src/components/dropdown/select/Select.tsx
+++ b/packages/react/src/components/dropdown/select/Select.tsx
@@ -65,6 +65,7 @@ export interface SelectCustomTheme {
 }
 
 export type CommonSelectProps<OptionType> = {
+  'aria-labelledby'?: string;
   /**
    * When `true`, allows moving from the first item to the last item with Arrow Up, and vice versa using Arrow Down.
    */
@@ -116,7 +117,7 @@ export type CommonSelectProps<OptionType> = {
   /**
    * The label for the dropdown
    */
-  label: React.ReactNode;
+  label?: React.ReactNode;
   /**
    * Callback function fired when the state is changed
    */
@@ -299,6 +300,7 @@ export const Select = <OptionType,>(props: SelectProps<OptionType>) => {
   // we can't destructure all the props. after destructuring, the link
   // between the multiselect prop and the value, onChange etc. props would vanish
   const {
+    'aria-labelledby': ariaLabelledBy,
     circularNavigation = false,
     className,
     clearable = props.multiselect,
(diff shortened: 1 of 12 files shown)


packages/react/src/components/dropdown/select/Select.tsx
```diff
@@ -65,6 +65,7 @@ export interface SelectCustomTheme {
 }
 
 export type CommonSelectProps<OptionType> = {
+  'aria-labelledby'?: string;
```
minevala, line 68: Should we have some description text for this property?

packages/react/src/components/dropdown/select/Select.tsx
```diff
@@ -116,7 +117,7 @@ export type CommonSelectProps<OptionType> = {
   /**
    * The label for the dropdown
    */
-  label: React.ReactNode;
+  label?: React.ReactNode;
```
minevala, line 120: Could we have some guidance or typing that component requires either label or aria-labelledby?

minevala (commented):
A nice addition 👍 
I was thinking should we rather have a functional test for component than regression tests?
The feature feels more like a functional addition rather than a visual change even tough it hides the default label. And the functional tests are way faster

packages/react/src/components/dropdown/select/Select.tsx
VilleMiekkoja: - [ ] Add description for aria-labelledby property

packages/react/src/components/dropdown/select/Select.tsx
VilleMiekkoja: - [ ] Add typing for conditionally requiring eitgher label or aria-labelledby

VilleMiekkoja:
> A nice addition 👍 I was thinking should we rather have a functional test for component than regression tests? The feature feels more like a functional addition rather than a visual change even tough it hides the default label. And the functional tests are way faster

- [ ] Add unit test for the case without label (with aria-labelledby)

VilleMiekkoja:
Unicus tested this and it should be ok.

minevala (approved):
🏅

PR #856, opened 2022-10-18
Hds 1420 accordions heading is not accessible

tuomashatakka:
## Description

Fixes the issue with accordion component being not accessible by moving the button inside the heading in the title.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1420

## How Has This Been Tested?

- Locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/accordion-header-a11y-fix/?path=/story/components-accordion--medium)

### DIFF
--- a/packages/react/src/components/accordion/Accordion.module.scss
+++ b/packages/react/src/components/accordion/Accordion.module.scss
@@ -76,6 +76,10 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
+
+  & > div {
+    flex: 1 1 auto;
+  }
 }
 
 .accordionContent {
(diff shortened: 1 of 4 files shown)


packages/react/src/components/accordion/Accordion.module.scss
```diff
@@ -76,6 +76,10 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
+
+  & > div {
```
minevala, line 80: I think it would be more future-proof to target the element using a plain class selector.  But to be honest, I did not come up with any good class name for this. Could it be just  ```.accordionHeading```? But we already have accordionHeadingContainer 😬 
This is more of a suggestion

minevala (commented):
Nice fix 👍 
One small suggestion. 
I was also wondering does this require an accessibility audit?

PR #859, opened 2022-10-21
HDS-1274 Use span tag instead of i tag for icons.

VilleMiekkoja:
## Description

Fixes a11y problem in documentation site and hds-core icon examples by using span tag instead of i tag.

This can be merged without a release.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1274

## How Has This Been Tested?

- Locally on dev machine

👉 [Demo](https://city-of-helsinki.github.io/hds-demo/docsite-link-icon-a11y-fix/components#components-overview)


minevala (commented):
Should we fix the core code examples as well?

minevala:
> Should we fix the core code examples as well?
I'll fix the examples as well

harriplappalainen:
Good addition 👍 i-tags were also used in some other code examples but I went ahead and fixed them. Btw, should we edit Accessibility statement regarding use i-tags?

harriplappalainen (approved):
This is 💎

PR #860, opened 2022-10-25
Core components' style importing made easier

harriplappalainen:
## Description

Instead of importing several CSS files when using core components, the user needs to import only one file which takes care of importing the rest of the needed styles. Only a small handful of components was affected (Link, Notification, Pagination, SearchInput and Tag.)

Are you worried about double style definitions? Don't be! `cssnano` takes care of removing double styles during building.

## Related Issue
https://helsinkisolutionoffice.atlassian.net/browse/HDS-1263

## Motivation and Context

Using core components should be simple and therefore importing the component's styles should be relatively straightforward.

## How Has This Been Tested?
Locally with `hds-cra` project and storybook
[Demo](https://city-of-helsinki.github.io/hds-demo/core-css-imports/?path=/story/components-link--external-links)

### DIFF
--- a/packages/core/src/components/link/link.css
+++ b/packages/core/src/components/link/link.css
@@ -1,3 +1,6 @@
+@import "../../icons/icon.css";
+@import "../../icons/ui/icon-link-external.css";
+
 .hds-link {
   --link-visited-color: #551a8b;
   --link-color: var(--color-bus);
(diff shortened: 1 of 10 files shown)


minevala (commented):
Some lokitests seem to fail, at least Notification and ErrorSummary. Is some dependency now broken?

harriplappalainen:
Lokitests fixed, it was caused by style overriding. Now core and react components work and look the same.

minevala (approved):
Nice improvement 👍

PR #862, opened 2022-10-27
Prevent DateInput calendar value reset on scroll

harriplappalainen:
Prevent value from resetting by disabling scroll event listener in react-popper. At least I didn't find how this could break functionality but **test with creativity please!** 

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1421

## Motivation and Context

hel.fi notified us about this in Slack. The need for a fix is great.

## How Has This Been Tested?
Locally on dev machine
[Demo](https://city-of-helsinki.github.io/hds-demo/dateinput-scroll-fix/?path=/story/components-dateinput--default)

### DIFF
--- a/packages/react/src/components/dateInput/DateInput.module.scss
+++ b/packages/react/src/components/dateInput/DateInput.module.scss
@@ -1,27 +1,3 @@
 .wrapper {
   position: relative;
 }
-
-.pickerWrapper {
-  transition: opacity 150ms;
-  display: flex;
-  justify-content: flex-end;
-  pointer-events: none;
-  opacity: 0;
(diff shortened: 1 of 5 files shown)


minevala (commented):
I tested this on mobile Safari and the problem still exists. The reason is that the mobile Safari browser also triggers resize event on scroll.

minevala (commented):
I tested the latest fix with an iPhone and the problem still occurs on the demo site.
Is the demo site up-to-date?

harriplappalainen:
My bad, I forgot to update it. Now it's up to date!

PR #866, opened 2022-11-03
Combobox scroll bug fix

harriplappalainen:
## Description

Combobox automatically scrolls to hovered menu items. This creates a confusing user experience when user scrolls down the list and (accidentally or not) moves the mouse up and the Combobox keeps scrolling even if user didn't mean to. I think this functionality is more trouble than it's worth. 

I checked if this serves any purpose accessibility-wise with Emmi Nyqvist and she thought no, if anything it may confuse people with problems perceiving things or using the magnifying glass functionality.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1210

## Motivation and Context

People in the Jira ticket's comments and Slack are waiting for us to fix the behaviour.

## How Has This Been Tested?
Locally on dev machine.
[Demo](https://city-of-helsinki.github.io/hds-demo/combobox-scroll-fix/?path=/story/components-dropdowns-combobox--multi-select-example)

### DIFF
--- a/packages/react/src/components/dropdown/combobox/Combobox.tsx
+++ b/packages/react/src/components/dropdown/combobox/Combobox.tsx
@@ -447,6 +447,10 @@ export const Combobox = <OptionType,>(props: ComboboxProps<OptionType>) => {
         onFocus={ignoreFocusHandlerWhenClickingItem(handleWrapperFocus)}
         onBlur={ignoreFocusHandlerWhenClickingItem(handleWrapperBlur)}
         onClick={handleWrapperClick}
+        onMouseUp={() => {
+          setIsClicking(false);
+          focusInput();
+        }}
         className={classNames(styles.wrapper, props.multiselect && props.icon && styles.wrapperWithMultiSelectAndIcon)}
         ref={getComboboxProps().ref}
       >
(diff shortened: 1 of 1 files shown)


tuomashatakka (changes requested):
Seems to be working well on any viable browser. However, enter Safari...:

Safari still has a minor issue with the dropbox jumping slightly (and losing focus) when the scrollbar is dragged with a mouse.

Steps to reproduce:

1. Select one or more items from the list
2. Try to move the scroll position by dragging the scrollbar with the cursor.

safari-combobox-probelma

harriplappalainen:
That is an annoying little bug. I could only get it fixed by adding onMouseDown event listener on the menu's parent div which sets the input focus after a 50ms time out (!!) but then it breaks the component other browsers 😆 As timeouts are nasty hack fixes already I wouldn't want to add a flimsy browser check as well (so as to only do the timeout on Safari). I'm inclined to let this one slide and make a ticket about it to fix it later because the bug exists on production release already and this PR's fix is eagerly anticipated by users. What do you think?

minevala:
Does Select component have similar issues?
I think we could check that as well if the fix is this simple 👍

harriplappalainen:
>Does Select component have similar issues?
No, Select all fine & dandy. It didn't use onHighlightedChange which was causing the problem. And it doesn't have Safari issues either.

harriplappalainen:
@tuomashatakka The safari problem has been fixed and can be tested in the demo.

minevala (approved):
It works nicely. A good little fix 🎉

tuomashatakka (approved):

PR #869, opened 2022-11-07
Add missing enterkeyhint-attribute to search input

minevala:
## Description
- Add missing enterhintKey to search input

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1429

## Motivation and Context
- The enterhint key enumeration tells the mobile browsers what kind of label enter key needs. The browser will resolve the text with the help of the enterhintkey-attribute.

## How Has This Been Tested?
- locally on the dev machine
- snapshot test

[Demo](https://city-of-helsinki.github.io/hds-demo/hds-1429-search-input-mobile-enter-fix/iframe.html?args=&id=components-searchinput--default&viewMode=story)

### Old in Mobile Chrome:
FF8F8982-518C-4BF2-874D-71B2C0F376A2

### With this fix in Mobile Chrome:
A4E33B6A-5097-40AF-ACED-E2755EB9AA2A

### DIFF
--- a/packages/react/src/components/searchInput/SearchInput.tsx
+++ b/packages/react/src/components/searchInput/SearchInput.tsx
@@ -192,6 +192,7 @@ export const SearchInput = <SuggestionItem,>({
           })}
           className={classNames(styles.input)}
           placeholder={placeholder}
+          enterKeyHint="search"
         />
         <div className={styles.buttons}>
           {inputValue.length > 0 && (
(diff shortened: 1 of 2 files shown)


harriplappalainen (approved):
Yes, I think this is a good and enough of an improvement as the value `search` is recognised and supported by browsers. I don't think there is a need to give the user the ability to change it in this component's case and if there ever will be a need, we'll make a ticket about it. I approve!

PR #871, opened 2022-11-08
Make the DateInput button area larger

tuomashatakka:
## Description

Update the DateInput button to have a more sizeable click area.

## Motivation and Context

Makes it easier to open the date picker.

## How Has This Been Tested?

- Automated tests
- On a local machine


harriplappalainen:
While this does make the calendar button larger and easier to press even on mobile, the resulting look deviates from other such input+icon components such as PasswordInput and SearchInput. The icon calendar in this case is positioned to the left much more than the other components' icons. For unified look they should all be at the exact same spot. 

I included an image from the ticket at the end of this comment which demonstrates the area that should be clickable. So if you can keep the icon where it is and make the green area in the image be clickable, that would be great. If the changes should be done on TextInput's end, which is likely, it would be great if the other input+icon components would have bigger clickable areas as well. If that proves to be a bit too much work for now, we can make a ticket about it.

I found that removing `margin-right: var(--spacing-s);` from button's parent div (with class "buttonWrapper") sets the icon to it's old place with the exception that it's a bit higher up than the old one. If this is a good lead, just check that the other input icon components work as well.

icon-calendar-bounding-box

harriplappalainen:
I noticed that when the icon has focus, the focus borders break outside of parent element which looks a bit wonky. I don't think other hds components have that? If there are others then it might be ok.

packages/react/src/components/dateInput/DateInput.tsx
```diff
@@ -187,8 +187,9 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
       <div lang={language} className={styles.wrapper}>
         <TextInput
           {...textInputProps}
-          buttonIcon={disableDatePicker ? undefined : <IconCalendar aria-hidden />}
+          buttonIcon={disableDatePicker ? undefined : <IconCalendar aria-hidden style={{ padding: '1em' }} />}
```
harriplappalainen, line 190: I think you should use spacing design token here, spacing-s in this case.

PR #874, opened 2022-11-09
Change the select input error icon

tuomashatakka:
## Description

Unify the error icons in input components

## How Has This Been Tested?

- Automated tests
- On a local machine


harriplappalainen (approved):
Nice fix 👍 You only forgot the updated snapshots but I went ahead and pushed them.

PR #877, opened 2022-11-09
HDS-1439 favicon documentation improvement

laitine:
## Description

* New favicon kit assets are to be used in the documentations site
* The documentation needs to reflect the latest favicon guidelines and specification
* Old favicon files removed

## Motivation and Context

With the new favicon we should follow guidelines from it also on the documentation site.


harriplappalainen:
The documentation seems solid 👍 But I found that the favicon doesn't display at all on Safari 😞 And the favicon on doc site doesn't change based on browser theme. It would be nice to have that functionality just to prove the point about automatically changing favicon. Seems like a fix for that would be to fiddle with gatsby configurations to preferably use the svg? Or do the switch manually in gatsby-browser.js and `prefer-color-scheme`.

PR #878, opened 2022-11-10
Hds 1401 header component

harriplappalainen:
## Description

Header component to act as a wrapper for upcoming new navigation components. I imagine this will gain more meat around the bones as child components come to life but this component mainly serves as means to pass on information relevant to the context. As of now it only provides information whether view is in mobile.

I removed everything else that didn't seem sense to keep here when looking at the designs. I thought about Header accepting `theme` property but it might make more sense to give them straight to the subcomponents, as some of them are very specific. 

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1401

## How Has This Been Tested?
Locally on dev machine
[Demo](https://city-of-helsinki.github.io/hds-demo/header/?path=/story/components-header--example)


packages/react/src/components/header/Header.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { render } from '@testing-library/react';
+import { axe } from 'jest-axe';
+
+import { Header } from './Header';
+
+describe('<Header /> spec', () => {
+  it('renders the component', () => {
+    const { asFragment } = render(<Header />);
+    expect(asFragment()).toMatchSnapshot();
+  });
```
minevala, line 11: new line?

packages/react/src/components/header/Header.tsx
```diff
@@ -0,0 +1,31 @@
...
+   */
+  id?: string;
+}>;
+
+export const Header = ({ children, className }: HeaderProps) => {
+  const isMobile = useMobile();
+  const context: HeaderContextProps = { isMobile };
+  return (
+    <HeaderContext.Provider value={context}>
+      <div className={classNames(styles.header, className)}>
```
minevala, line 26: Should this be a header element?

minevala (commented):
A nice start 👍 
A couple of small comments

minevala:
Snapshot test seems to fail

packages/react/src/components/header/Header.tsx
harriplappalainen: Yes! Totally forgot.

harriplappalainen:
Good remarks! I changed the code a bit and added the snapshot tests which I forgot to do initially.

minevala (approved):
A nice start 🏅

PR #879, opened 2022-11-11
Date input calendar button clickable area is too small

tuomashatakka:
## Description

- Add padding to the input buttons. 
- Update focused input to use outline css property instead of box-shadow.

## Related issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1432

## Motivation and Context

The clickable area was too small for touch screens.

When testing this PR's changes locally, the hds-core library should be built locally before starting the hds-react storybook.

## How Has This Been Tested?

- Local dev environment
- Automated tests


minevala (commented):
There is slight differences with the placing of the icon in Lokitests
Maybe we could just increase the click area without moving the icon?

harriplappalainen:
>There is slight differences with the placing of the icon in Lokitests
Maybe we could just increase the click area without moving the icon?

To specify Mika's comment, I noticed by comparing this build's and productions storybook that only the SearchInput's search icon has moved a few pixels to the right. And a bigger move happened with the value clear icon which can be seen in story With default value (file referenced: packages/react/.loki/reference/chrome_iphone7_Components_SearchInput_With_default_value.png)

packages/core/src/components/text-input/text-input.css
```diff
@@ -258,18 +258,23 @@
...
+  border: none;
+  color: inherit;
   cursor: pointer;
-  border: 0;
-  width: 1.5rem;
-  height: 1.5rem;
-  overflow: hidden;
-  font-size: 1rem;
+  font: inherit;
+  color: var(--dropdown-color-default);
```
harriplappalainen, line 272: Color is defined twice? And this color definition might be the reason for loki images change because as Hannu pointed out, the color is different in new images. It used to be #0000 (color-black) but it has changed to #1a1a1a (black-90). Because it's a lighter color it might seem like there is an anti-aliasing effect because the difference is more noticeable to the naked eye around the borders of the icon.

tuomashatakka: Hmm, I removed the duplicate color definition but that had no difference in loki tests.

tuomashatakka: Oh, there was no inheritance before. I added it to unify the input stylings. Should we leave it as it is now or revert the `color: inherit`?

tuomashatakka: Reverted. Changed files of this PR went from 24 to 4 😆

harriplappalainen (approved):
Now everything is cool and where it should be. Stamp of approval! 👍

packages/core/src/components/text-input/text-input.css
```diff
@@ -258,18 +258,21 @@
...
   cursor: pointer;
-  border: 0;
-  width: 1.5rem;
-  height: 1.5rem;
-  overflow: hidden;
-  font-size: 1rem;
+  font: inherit;
+  outline: none;
+  display: flex;
+  padding: var(--spacing-xs) calc(var(--spacing-xs) / 2);
```
minevala, line 273: The button area would be bigger if we could use just:
`padding: var(--spacing-s);` and remove the `margin-right`
The current implementation:

Screenshot 2022-11-22 at 9 24 05

With the changes:
Screenshot 2022-11-22 at 9 23 43

**But** this means that the layout of the Search Input changes a bit because there are two buttons when the input has a value. They require more space now.
So it is either or a **breaking change** or we just accept that the change is so small that it does not require major release. Or we leave this as it is.

minevala: Screenshot 2022-11-22 at 9 33 44
Screenshot 2022-11-22 at 9 32 38

PR #887, opened 2022-11-17
Hds 1425 use theme fix

minevala:
## Description
- Add isomorphicLayoutEffect hook
- Use isomorphicLayoutEffect hook in useTheme hook
- Mention CSS specificity in the SSR guidelines customization part

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1425

## Motivation and Context
- Using useLayoutEffect hook will minimize the time of unstyled content flashing with CSR
- Documentation addition will help to resolve theming problems with SSR

## How Has This Been Tested?
- locally on dev machine

[UselayoutEffect Theming Demo and doc changes](https://city-of-helsinki.github.io/hds-demo/ssr-demo/foundation/guidelines/server-side-rendering/)

### DIFF
--- a/packages/react/src/hooks/useIsomorphicLayoutEffect.ts
+++ b/packages/react/src/hooks/useIsomorphicLayoutEffect.ts
@@ -1,4 +1,3 @@
-// use-isomorphic-layout-effect.js
 import { useLayoutEffect, useEffect } from 'react';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useIsomorphicLayoutEffect.ts
```diff
@@ -0,0 +1,8 @@
+// use-isomorphic-layout-effect.js
```
harriplappalainen, line 1: I think this comment can be removed. I might've accidentally left it here?

harriplappalainen (approved):
Good fix and nice addition to documentation! 👍 And you got the navigation theming working in doc site! 

I only had one tiny comment about removing a comment but otherwise this is solid.

packages/react/src/hooks/useIsomorphicLayoutEffect.ts
minevala: - [x] remove obsolete comment

PR #889, opened 2022-11-17
Hds 1483 fix broken path

minevala:
## Description
- Replace node module path extname  with local custom implementation

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1483

## Motivation and Context
- In some environment (Gatsby) the FileInput throws an error which breaks the application


harriplappalainen (approved):
I think the fix is good and clear. Nicely done 🏅

PR #891, opened 2022-11-21
Hds 1463 fix menuitems list styles

minevala:
## Description
- Use CSS overflow-y to let browsers resolve if the scrollbar is needed

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1463

## Motivation and Context
- When there are 5 or fewer options in the Combobox, the scrollbar does not exist even though it might be needed. For example in cases where a single option's label requires multiple lines.

## How Has This Been Tested?
- Locally on dev machine

[Demo](https://city-of-helsinki.github.io/hds-demo/hds-1463-fix-menuitems-list-styles/?path=/story/components-dropdowns-combobox--combobox-example)

### DIFF
--- a/packages/react/src/components/dropdown/_dropdown.common.scss
+++ b/packages/react/src/components/dropdown/_dropdown.common.scss
@@ -125,16 +125,13 @@
   left: 0;
   list-style: none;
   margin: 0;
+  overflow-y: auto;
   outline: none;
   padding: 0;
   position: absolute;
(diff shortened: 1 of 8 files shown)


harriplappalainen (approved):
Clever fix! I tried and tried to make it break on any of the related components but no, it stands tall 😄 Well done 🎖️

PR #898, opened 2022-11-23
Remove unnecessary TimeInput logging

minevala:
## Description
- Remove console.warn log when user types into TimeInput (does not bring user value)

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1480

## Motivation and Context
- At the moment TimeInput component warns on every value change even tough user was not ready typing the value. Validation warnings in the console are not very useful for end user.

## How Has This Been Tested?
- Locally on dev machine


harriplappalainen (approved):
I approveth thee

PR #899, opened 2022-11-25
HDS-1490 cookie modal bug with portal

NikoHelle:
## Description

If cookie consent modal is re-initialized or re-rendered after it has been closed, an error will be thrown from React.Portal. It tries to render Portal to a non-existing DOM element.

The component’s child checker does not handle screen reader notification properly and tries to re-render even it should not.

**Note**
Fixed also `onConsentsParsed` being called multiple times. It was called on each context change.
Now it is set as `undefined` after first call. The callback is copied, so any original objects are not mutated.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1490

## How Has This Been Tested?

Added failing tests before changes, which succeeded after the changes were done.


NikoHelle:
Git LFS re-init messed this branch and PR. New one https://github.com/City-of-Helsinki/helsinki-design-system/pull/904

PR #911, opened 2022-12-12
Feature/hds 1434 build a storybook for webcomponents

tuomashatakka:
## Description

Web components initial implementation with a working storybook

## Related Issues

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1434

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1426

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1442


packages/web-components/src/components/accordion/Accordion.ts
```diff
@@ -0,0 +1,179 @@
...
+      </div>
+
+      <div
+        id=${this.identifiers.content}
+        style="display: ${this.isOpen ? 'block' : 'none'}"
+        role="region"
+        class='accordionContent'
+        aria-labelledby=${this.identifiers.heading}>
+        <slot></slot>
+        ${this.renderCloseButton()}
```
harriplappalainen, line 115: Close button should be on the next line of the accordion content. Now it comes on top and hover style overlaps the text a bit.

packages/web-components/src/components/button/Button.ts
```diff
@@ -0,0 +1,206 @@
...
+        ${this.renderRightIcon()}
+      </button>
+    `;
+  }
+
+  renderLabel () {
+    return html`
+      <span class='label'>
+        ${this.isLoading ? this.loadingText : ''}
+        <slot></slot>
```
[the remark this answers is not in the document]
harriplappalainen: Ooh I got it! This is like `{props.children}`, right? This text should be in bold or something by the way, the difference is subtle but noticable if you compare this to React button in storybook.

packages/web-components/src/components/accordion/Accordion.ts
```diff
@@ -0,0 +1,179 @@
...
+      <hds-button
+        class=${className}
+        variant='supplementary'
+        aria-label=${closeLabel}
+        theme='black'
+        size='small'
+        @keypress=${handleKeyPress}
+        @click=${this.toggleOpen}>
+        ${closeMessage}
+      </hds-button>
```
harriplappalainen, line 158: Button is missing chevron but I think we discussed that we will go without it for now because it would need more work.

packages/web-components/stories/Introduction.stories.mdx
```diff
@@ -0,0 +1,211 @@
+import { Meta } from '@storybook/addon-docs';
```
harriplappalainen, line 1: Is this whole stories folder necessary? Was this generated automatically?

minevala (commented):
Nice work 🏅 
It needs some cleaning up and we probably should find a way to import Core styles before publishing

minevala:
I noticed that my yarn.lock changed compared to the one in the branch. Maybe that is causing the broken build?

PR #913, opened 2022-12-13
HDS-1481: set notification heading level

NikoHelle:
## Description

The arial-level is fixed to be "2", which may cause accessibility errors. The default is now "2", but can be overridden.

Closes HDS-1481

## How Has This Been Tested?

Tests included.


minevala:
Otherwise this looks good 👍 but one Lokitest fails

NikoHelle:
Failed test was a Cookie consent test. Which does not use notification. And the test passed on second run. Weird.

minevala (approved):

minevala:
Do not merge yet 😁

PR #914, opened 2022-12-16
HDS-1299: overly long dropdown tags

NikoHelle:
## Description

If the element of the first selected item is wider that the container,  all items are hidden. The component hides all children that overflow the container.

After the fix, if there are selected items and all are hidden (indicating the first one is too wide), then a css class is added to the container which will truncate the label of the first item. 

## How Has This Been Tested?

Loki tests only. Jest tests cannot be used, because jsdom does not calculate element sizes.

### DIFF
--- a/packages/react/src/internal/selectedItems/SelectedItems.module.scss
+++ b/packages/react/src/internal/selectedItems/SelectedItems.module.scss
@@ -34,10 +34,10 @@
 
   &.truncateFirstChild {
     flex-wrap: nowrap;
-    & > :first-child {
+    & > .tag:first-child {
       justify-content: flex-end;
       overflow: hidden;
-      > span {
+      .tagLabel {
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
--- a/packages/react/src/internal/selectedItems/SelectedItems.module.scss
+++ b/packages/react/src/internal/selectedItems/SelectedItems.module.scss
@@ -5,12 +5,12 @@
   visibility: hidden;
 }
 
-
 .selectedItems {
   align-items: center;
   display: flex;
   flex-wrap: wrap;
-  padding: var(--spacing-4-xs) calc((var(--spacing-s) - var(--border-width)) * 2 + (var(--icon-size) * 2)) calc(var(--spacing-2-xs) + var(--border-width)) var(--spacing-4-xs);
+  padding: var(--spacing-4-xs) calc((var(--spacing-s) - var(--border-width)) * 2 + (var(--icon-size) * 2))
+    calc(var(--spacing-2-xs) + var(--border-width)) var(--spacing-4-xs);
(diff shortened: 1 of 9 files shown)


packages/react/src/internal/selectedItems/SelectedItems.module.scss
```diff
@@ -32,6 +32,19 @@
     }
   }
 
+  &.truncateFirstChild {
+    flex-wrap: nowrap;
+    & > :first-child {
+      justify-content: flex-end;
+      overflow: hidden;
+      > span {
```
harriplappalainen, line 40: It's a shame that the span inside Tag component doesn't have a class and styles because browsers parse the selectors bottom-up, meaning the selectors are read from right to left. So first the browser finds all the span elements in the DOM and then it continues with the selectors. Should we refactor the Tag component and give a prop to Tag's span element?

harriplappalainen (commented):
Really solid fix 👍 I just had one comment about CSS selector parsing, check it out and let me know what you think.

minevala (commented):
Nice fix 👍 
One question/suggestion

packages/react/src/internal/selectedItems/SelectedItems.module.scss
minevala: Could we use the Tag's className for this?

minevala: With hds-prefix, `.hds-dropdown-tag` etc?

NikoHelle: I actually tried to pass className to <Tag> via "labelProps", but that overrides all tag's classNames. So I rejected that approach, because tag loses all base styles. 

How can I pass "hds-dropdown-tag" to the Tag-component?

harriplappalainen: I guess a new props like `labelClassName` would need to made to Tag component. I'm not sure if the styling needs to be at the innermost span or the "label" span that wraps it? If the wrapper is enough => prop `labelClassName` and merge that with the existing className. If the innermost, then the naming gets a bit confusing imo. Maybe `textClassName` ?

NikoHelle: This is to get rid of `> span` and replace it with `> .hds-dropdown-tag`. For performance gain?

minevala: And also for clarity and for better selector 👍 
The class tells more than span that we are targeting the "tag" 😁

NikoHelle: Done.

harriplappalainen (approved):
Now it's magnifico 👌

PR #916, opened 2022-12-20
HDS-1509: add Tunnistamo login cookies to cookie content

NikoHelle:
Note: merge https://github.com/City-of-Helsinki/helsinki-design-system/pull/933 first!
## Description

Tunnistamo is the common login provider for Helsinki services. The login flow uses multiple other services (Tunnistus, Suomi.fi) to provide strong authentication. Those services use cookies and have no cookie consent tools in the service, so the services using Tunnistamo need to ask for consents.

All used cookies are added to HDS, so it is easier to ask for consents and same consents does not have to be asked again and again.

Added also a common group "tunnistamoLogin", which lists all cookies related to Tunnistamo login cookies. Load balancer, security and language cookies must be grouped manually as the service probably needs to add its own related cookies to those groups.

There is a new story to demo the usage.

## How Has This Been Tested?

Unit tests. Most of the changes are just textual additions.


laurakarhu (approved):
Very tasty cookies 🍪

packages/react/src/components/cookieConsent/getContent.ts
```diff
@@ -0,0 +1,806 @@
...
+    fi: {
+      description: 'Eväste vaaditaan jotta käyttäjän kielivalinta säilyisi.',
+      expiration: 'Istunto',
+    },
+    sv: {
+      description: 'Kakan krävs för att spara användarens språkval.',
+      expiration: 'Session',
+    },
+    en: {
+      description: "Required to persist the user's chosen language..",
```
harriplappalainen, line 12: I don't think there's a need for passive aggressive dots at the end of sentence 😆

harriplappalainen (commented):
Looks good 👍 My only remark is one typo in the texts.

packages/react/src/components/cookieConsent/getContent.ts
NikoHelle: :) Good catch!

PR #923, opened 2022-12-28
HDS-1516: fix search input cursor jumping

NikoHelle:
## Description

When SearchInput had text and the text was edited somewhere else than end of the string, the cursor jumped to the end.

Downshift rendered the input multiple times when input was changed: once with new value, once with old value and then new again. React could not track the cursor position.

Changed how changes are tracked with DownShift. Link to suggested fixes are in the ticket.

Also fixed how home/end keys work inside the input as commented in the ticket

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1516

## How Has This Been Tested?

Added new tests

### DIFF
--- a/packages/react/src/components/searchInput/SearchInput.test.tsx
+++ b/packages/react/src/components/searchInput/SearchInput.test.tsx
@@ -1,8 +1,8 @@
 import React, { useState } from 'react';
-import { render, waitFor, screen } from '@testing-library/react';
+import { render, waitFor, screen, RenderResult } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
-import { SearchInput } from './SearchInput';
+import { SearchInput, SearchInputProps } from './SearchInput';
 
 type SuggestionItemType = {
   value: string;
@@ -21,6 +21,51 @@ const getSuggestions = (inputValue: string): Promise<SuggestionItemType[]> => {
 };
 
 describe('<SearchInput /> spec', () => {
+  let renderResult: RenderResult;
+  const defaultProps: Pick<
+    SearchInputProps<SuggestionItemType>,
+    'label' | 'suggestionLabelField' | 'getSuggestions'
+  > = {
+    label: 'search',
+    suggestionLabelField: 'value',
+    getSuggestions,
+  };
+
+  const getInputByDefaultLabel = () => {
+    return renderResult.getByLabelText(defaultProps.label as string, { selector: 'input' });
+  };
+
+  const clickResetButton = () => {
+    userEvent.click(renderResult.getByLabelText('Clear', { selector: 'button' }));
+  };
+
+  const initTests = (props: Partial<SearchInputProps<SuggestionItemType>>, defaultValue?: string): RenderResult => {
+    const mergedProps = {
+      ...defaultProps,
+      ...props,
+    } as SearchInputProps<SuggestionItemType>;
+    if (defaultValue) {
+      const ControlledSearchInput = () => {
+        const [stateValue, updateStateValue] = useState(defaultValue);
+        const onChange = (value: string) => {
+          if (props.onChange) {
+            props.onChange(value);
+          }
+          updateStateValue(value);
+        };
+        return (
+          <div>
+            <SearchInput<SuggestionItemType> {...mergedProps} onChange={onChange} value={stateValue} />
+          </div>
+        );
+      };
+      renderResult = render(<ControlledSearchInput />);
+    } else {
+      renderResult = render(<SearchInput<SuggestionItemType> {...mergedProps} />);
+    }
+    return renderResult;
+  };
+
   it('renders the component', () => {
     const { asFragment } = render(<SearchInput label="Search" onSubmit={() => null} />);
     expect(asFragment()).toMatchSnapshot();
@@ -30,17 +75,9 @@ describe('<SearchInput /> spec', () => {
     const onChange = jest.fn();
     // eslint-disable-next-line no-console
     const onSubmit = () => console.log('submit');
-    const { getAllByLabelText } = render(
-      <SearchInput<SuggestionItemType>
-        label="search"
-        suggestionLabelField="value"
-        getSuggestions={getSuggestions}
-        onChange={onChange}
-        onSubmit={onSubmit}
-      />,
-    );
+    initTests({ onSubmit, onChange });
 
-    const input = getAllByLabelText('search')[0];
+    const input = getInputByDefaultLabel();
     userEvent.type(input, 't');
     expect(onChange.mock.calls[0][0]).toBe('t');
     userEvent.type(input, 'e');
@@ -53,16 +90,9 @@ describe('<SearchInput /> spec', () => {
 
   it('submits the selected item on mouse click', async () => {
     const onSubmit = jest.fn();
-    const { getAllByLabelText, getAllByRole } = render(
-      <SearchInput<SuggestionItemType>
-        label="search"
-        suggestionLabelField="value"
-        getSuggestions={getSuggestions}
-        onSubmit={onSubmit}
-      />,
-    );
+    const { getAllByRole } = initTests({ onSubmit });
 
-    const input = getAllByLabelText('search')[0];
+    const input = getInputByDefaultLabel();
     userEvent.type(input, 'a');
     await waitFor(() => {
       const options = getAllByRole('option');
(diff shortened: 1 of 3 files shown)


packages/react/src/components/searchInput/SearchInput.test.tsx
```diff
@@ -71,6 +71,36 @@ describe('<SearchInput /> spec', () => {
...
+    const { getAllByLabelText, container } = render(
+      <SearchInput<SuggestionItemType>
+        label="search"
+        suggestionLabelField="value"
+        getSuggestions={getSuggestions}
+        onSubmit={onSubmit}
+      />,
+    );
+
+    const input = getAllByLabelText('search')[0];
```
minevala, line 85: Could this work in here:
```const input = screen.getByLabelText('search', { selector: 'input' })```

File: packages/react/src/components/searchInput/SearchInput.test.tsx [hunk 2]
```diff
@@ -71,6 +71,36 @@ describe('<SearchInput /> spec', () => {
...
+    userEvent.type(input, 'abc{enter}');
+    await waitFor(() => {
+      expect(onSubmit.mock.calls[0][0]).toBe('abc');
+    });
+    const buttons = container.querySelectorAll('button');
+    const clearButton = buttons[0];
```
minevala, line 91: Could this work in here:
const input = screen.getByLabelText('Clear', { selector: 'button' })

packages/react/src/components/searchInput/SearchInput.test.tsx
```diff
@@ -71,6 +71,36 @@ describe('<SearchInput /> spec', () => {
...
+    const submitButton = buttons[1];
```
minevala, line 92: Could this work in here:
const input = screen.getByLabelText('Search', { selector: 'button' })

minevala (commented):
Nice fix 👍 
A few small suggestions or comments

packages/react/src/components/searchInput/SearchInput.tsx
```diff
@@ -133,43 +151,40 @@ export const SearchInput = <SuggestionItem,>({
...
-    if (isItemClicked.current) {
-      isItemClicked.current = false;
-      submit();
+  const onInputKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+    // isOpen is sometimes true for no reason
+    if (isOpen && suggestions.length) {
+      return;
+    }
+    if (event.key === 'Home' || event.key === 'End') {
+      // When preventDownshiftDefault = true, downshift does not alter native behavior
```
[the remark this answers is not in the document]
NikoHelle: From the issue:

> Home (Optional): Either moves focus to and selects the first option or returns focus to the textbox and places the cursor on the first character.
> End (Optional): Either moves focus to the last option or returns focus to the textbox and places the cursor after the last character.
> This is from https://www.w3.org/TR/wai-aria-practices/#listbox-popup-keyboard-interaction. Yours is from an implementation that chose the second approach. So neither of us is wrong.
> 
> I would prefer to leave it like it is, and you can use preventDownshiftDefault on Home/End (https://github.com/downshift-js/downshift#customizing-handlers) and have your scenario fixed.

**"So neither of us is wrong."** They kept their implementation as there is a way to do it in another way (adopted to HDS).

I didn't follow throught the re-open/close discussion.

packages/react/src/components/searchInput/SearchInput.test.tsx
NikoHelle: Works.

packages/react/src/components/searchInput/SearchInput.test.tsx
NikoHelle: Works

packages/react/src/components/searchInput/SearchInput.test.tsx
NikoHelle: Works

packages/react/src/components/searchInput/SearchInput.tsx
```diff
@@ -122,6 +122,24 @@ export const SearchInput = <SuggestionItem,>({
...
   const isControlledComponent = value !== undefined && onChange;
 
+  const dispatchValueChange = (changedValue: string) => {
+    if (isControlledComponent) {
+      onChange(changedValue);
+    } else {
+      setInternalValue(changedValue);
+    }
+  };
+  const onInputValueChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
NikoHelle: I added a comment ...but the input value change never was tracked in the  useCombobox's onStateChange. Not even before this PR, so maybe comment does not provide the information you want.

packages/react/src/components/searchInput/SearchInput.tsx
```diff
@@ -133,43 +156,40 @@ export const SearchInput = <SuggestionItem,>({
...
-    onInputValueChange: (e) => {
-      if (isControlledComponent) {
-        onChange(e.inputValue);
-      } else {
-        setInternalValue(e.inputValue);
-      }
-    },
     onStateChange(props) {
       const { ItemClick } = useCombobox.stateChangeTypes;
       if (props.type === ItemClick) {
```
harriplappalainen, line 161: By testing this locally, I detected that only having values submitted via ItemClick removes some other functionalities when SearchInput is used as a controlled component (check story WithDefaultValue). At least resetting value or selecting the value from the list with keyboard. Luckily the docs have [stateChangeTypes](https://github.com/downshift-js/downshift/tree/master/src/hooks/useCombobox#statechangetypes) that can help cover this! Check them out if there's something else that might help, there is stuff about Home and End keys as well.

Here's a suggestion to support at least resetting and keyboard selection, feel free to add any more that you find useful:
`const { InputKeyDownEnter, ItemClick, FunctionReset } = useCombobox.stateChangeTypes;
if (props.type === ItemClick || props.type === FunctionReset || props.type === InputKeyDownEnter) {
        if (props.type !== FunctionReset) isItemClicked.current = true;
        dispatchValueChange(props.inputValue);
        submitValue(props.inputValue);
 }`

harriplappalainen (commented):
Well done! 👍 Your fixes seem to work well. Only problem I found is with controlled component usage under With Default Value -story.

packages/react/src/components/searchInput/SearchInput.tsx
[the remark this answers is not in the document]
minevala: I think it is good, it clarifies why the input handler is not part of the onStateChange.

packages/react/src/components/searchInput/SearchInput.tsx
NikoHelle: Oh, what a bug! The reset check is even in the tests. But it tested uncontrolled component which worked.

The state change tracks now also `FunctionReset`  and `InputKeyDownEnter`. This change caused more side-effects and changes to the component. There were double submits.

I added a more tests so these won't occur again undetected.

Home and End keys must be trapped onKeyDown to prevent Downshift default behaviour.

harriplappalainen: Nice! Now those are covered. But unfortunately I found another bug and console.error.. 🙈 Bit of edge-cases though but still.

1. `SearchInput.tsx:245 Uncaught TypeError: Cannot read properties of undefined (reading 'length')` crash occurs when you type a value which is suggested -> click it -> enter text into input and remove it -> set focus elsewhere on page
2. related this is a console error about switching to controlled -> uncontrolled component. If you do the same steps with WithDefaultValue story there's a console.error about this

NikoHelle: Wow... I couldn't quite follow your steps, but found easier way

If you type **exactly** the same value that is suggested, like "Apple", and click the suggestion, it crashes. The object passed to "onStateChange" has no inputValue-prop when it matches the suggested value.

NikoHelle: Fixed now. The behaviour now slightly differs from the current version in master.
The value is now submitted when the value in the input matches the value of the selected item in the dropdown. 

But this is actually  a fix, because value is submitted in all other cases when item is selected. For example if "Appl" is written and "Apple" is selected. :D 

Added more tests.

harriplappalainen: Nicely fixed! I'm not finding anymore bugs in functionality. This whole PR is a Major improvement to this component! 🥇 Has @minevala anything to add to the latest commits?

tuomashatakka (approved):

packages/react/src/components/searchInput/SearchInput.tsx
tuomashatakka: I marked this approved for now as it seems that the tests cover the prior issues nicely and they seem to be passing :)

PR #924, opened 2022-12-28
Hds 1493 Remove LFS

minevala:
## Description
- Run LFS export and make a custom commit without export's autocommit
- Remove .gitattributes
- Use git-checkout in GitHub Actions' configuration
- Remove all texts mentioning git lfs

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1493

## Motivation and Context
- Git LFS has caused issues in CI-pipeline and local development as well. Especially when trying to add new Lokitest images. We also had to buy more space for our LFS cache from GitHub. Let's try to live without it for a while.


harriplappalainen (approved):
I don't see much to review here, the changes seem fine 😅 Mentions about git lfs are removed from docs as well. Great work! 🤝

PR #926, opened 2022-12-30
HDS-1523: noscript localization

laitine:
## Description

* Noscript feature with text content in Finnish, Swedish and English

## Motivation and Context

To make the noscript content more inclusive they need to be displayed according to the browsers language in FI, SV and EN.

## How Has This Been Tested?

On latest Mac Chrome, Firefox and Safari browsers with JavaScript disabled.


laurakarhu (approved):
Hyvä, jättebra, very nice!

minevala (commented):
I could tweak the example HTML to make it a bit simpler

minevala (approved):
Nice, works nicely when I was able to test it on a separate HTML page 👍

PR #930, opened 2023-01-03
HDS-1011: Prevent clearing multi-item ComboBox on selection

NikoHelle:
The input was cleared on selection, which is annoying when wanting to select multiple items and items need to be filtered first.

The input value is now kept until blurred or menu is closed.

## Description

When item is selected with pointer / spacebar, the input value is not reset like before.
If item is selected with enter key, then the menu closes and input is cleared. This behaviour has not changed in this PR.
Input is cleared on blur, if it is a multi-select combobox

## Related Issue

Closes #

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1011

## How Has This Been Tested?

Added tests.

[HDS-1011]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/dropdown/combobox/Combobox.test.tsx
+++ b/packages/react/src/components/dropdown/combobox/Combobox.test.tsx
@@ -121,7 +121,7 @@ describe('<Combobox />', () => {
         multiselect: true,
       });
 
-      const input = getAllByLabelText(label)[0];
+      const input = getAllByLabelText(label)[0] as HTMLInputElement;
 
       // Search an option
       userEvent.type(input, 'Fi');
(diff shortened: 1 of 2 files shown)


tuomashatakka (approved):

minevala (approved):
It would make reviews faster if there were a live demo 👍 
Seems to work when I ran this locally 🏅

PR #931, opened 2023-01-03
HDS-1521: Deleting tags in combobox has weird side-effects

NikoHelle:
## Description
If options are selected and then tags are deleted while options are open, items are deleted quite randomly and item can even be added. Depends on the indexes of the selected and deleted items.

Preventing selections onBlur fixes the problem.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1521

## How Has This Been Tested?

Added tests which fail without these changes.

[HDS-1011]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-1521]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/dropdown/combobox/Combobox.test.tsx
+++ b/packages/react/src/components/dropdown/combobox/Combobox.test.tsx
@@ -225,5 +225,52 @@ describe('<Combobox />', () => {
       const visibleOptions = getAllByRole('option');
       expect(visibleOptions.length).toBe(1);
     });
+
+    it('deleting an item while suggestions are shown, should only delete the item and not add an option', async () => {
+      const onChange = jest.fn();
+      const { getAllByLabelText, getAllByRole, queryAllByText } = getWrapper({
+        onChange,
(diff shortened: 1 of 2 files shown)


minevala (approved):
There seems to be something weird with the Combobox state with Downshift but this fixes the issue 👍 Testing Combobox is a bit hard but you managed to do it 🏅 I tried some user oriented selectors for tests but they did not work because of the complex structure of the Combobox. Good job 🏅

PR #945, opened 2023-01-20
JSON5 upgrade

harriplappalainen:
## Description

Set JSON5 dependency as a resolution in package.json to v 2.2.2 which fixes a security issue.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1544

## How Has This Been Tested?
Locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   },
   "resolutions": {
     "multer": "1.4.4-lts.1",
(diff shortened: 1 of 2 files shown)


minevala (approved):
Seems to work 👍 
And based on a quick glance at releases, the library seems to follow semver so this is not a breaking change.

PR #951, opened 2023-01-27
Update table

laurakarhu:
remove futurice from table

## Motivation and Context
Futurice hasn't been involved after 1.0.0 anymore

## How Has This Been Tested?
running locally


minevala (approved):
👍
There are empty spaces around string but they were already there

PR #952, opened 2023-01-30
HDS-1360 Storybook React component extended properties

mikkojamG:
## Description

Fix missing components properties

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1360

## Motivation and Context

As a default `react-docgen-typescript` parser filters out extended properties for `node_modules`. Some HDS components have properties named identical to native HTML attributes e.g. `className`. Changing component types in such a way that normal properties are treated as a first priority and won't then get filtered out with extended properties.

## How Has This Been Tested?
- Tested on local machine
- Unit tests

### DIFF
--- a/packages/react/.storybook/main.js
+++ b/packages/react/.storybook/main.js
@@ -1,5 +1,3 @@
-const path = require('path');
-
 module.exports = {
   core: {
     builder: 'webpack5',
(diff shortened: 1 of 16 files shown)


laurakarhu (approved):
Looks like quite simple change with settings. :)

harriplappalainen (approved):
Just perfect 💯 🎉

PR #955, opened 2023-02-03
HDS-1367-dialog-page-jump

mrTuomoK:
## Description
Fixes the nasty content and scroll position jumping due to Dialog-component rendering. Also did some rework on the accessibility-related containers controlling the focus. (~40 lines of code less now)

## Related Issue
[HDS-1367](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1367)

## How Has This Been Tested?
- tests
- on local machine
- on iOS simulator

[HDS-1367]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/dialog/Dialog.module.scss
+++ b/packages/react/src/components/dialog/Dialog.module.scss
@@ -24,7 +24,6 @@
 
   .dialogBackdrop {
     background: var(--overlay-color);
-    pointer-events: none;
   }
 
   .dialog {
--- a/packages/react/src/components/dialog/Dialog.module.scss
+++ b/packages/react/src/components/dialog/Dialog.module.scss
@@ -9,17 +9,18 @@
 .dialogContainer {
   --accent-line-color: var(--color-bus);
   --overlay-color: rgba(0, 0, 0, 0.3);
-  -webkit-overflow-scrolling: touch;
+  -webkit-overflow-scrolling: none;
   align-items: center;
   bottom: 0;
   display: flex;
   justify-content: center;
   left: 0;
   overflow: auto;
-  overscroll-behavior: contain;
+  overscroll-behavior: none;
   position: fixed;
   right: 0;
   top: 0;
+  touch-action: none;
   z-index: 800;
 
   .dialogBackdrop {
--- a/packages/react/src/components/dialog/Dialog.module.scss
+++ b/packages/react/src/components/dialog/Dialog.module.scss
@@ -3,7 +3,7 @@
 @value medium-up from "../../styles/breakpoints.css";
 
 .dialogVisibleBodyWithHiddenScrollbars {
-    overflow: hidden;
+    overflow: hidden !important;
 }
 
 .dialogContainer {
--- a/packages/react/src/components/dialog/Dialog.module.scss
+++ b/packages/react/src/components/dialog/Dialog.module.scss
@@ -3,26 +3,24 @@
 @value medium-up from "../../styles/breakpoints.css";
 
 .dialogVisibleBodyWithHiddenScrollbars {
-  touch-action: none;
-  -webkit-overflow-scrolling: none;
-  overflow: hidden !important;
-  overscroll-behavior: none;
+    overflow: hidden;
 }
 
 .dialogContainer {
   --accent-line-color: var(--color-bus);
   --overlay-color: rgba(0, 0, 0, 0.3);
-
+  -webkit-overflow-scrolling: touch;
   align-items: center;
+  bottom: 0;
   display: flex;
-  height: 100%;
   justify-content: center;
   left: 0;
-  position: fixed;
-  top: 0;
-  width: 100%;
-  z-index: 800;
   overflow: auto;
+  overscroll-behavior: contain;
+  position: fixed;
+  right: 0;
+  top: 0;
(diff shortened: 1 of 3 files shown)


minevala:
There is a small difference with current implementation in tab focus behaviour when the focus shifts outside the browser viewport. In current implementation the focus shifts to last element from first browser tab with Shift + Tab. In this implementation the focus shifts to the close cross button. 
I'm not sure how crucial this kind of behaviour is, I think we could check this with our accessibility experts anyways.

minevala (commented):
Nice work, lots of deleted lines 👍 
I had small question about the implementation and we could check the behaviour with our accessibility experts as well.

mrTuomoK:
Yeah lets discuss next week 👌🏻 And the focus from elsewhere -> dialog, it's easy to change where it focuses, I thought it was the 1st one that's wanted (if I remember the tests wanted it that way 🧐🤷‍♂️)

packages/react/src/components/dialog/Dialog.tsx
```diff
@@ -191,11 +148,29 @@ export const Dialog = ({
...
+    }
+    return null;
+  }, []);
+
+  useEffect(() => {
+    // if currently focused element is not inside the Dialog
+    if (!dialogRef?.current?.contains(focusedElement)) {
+      focusFirstDialogElement(dialogRef.current);
+    }
+  }, [focusedElement]);
```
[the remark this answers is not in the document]
mrTuomoK: Ah yes I could combine some to the pre-existing useEffect 👌🏻 will refine next week.

mrTuomoK:
Ok now this should be done, fallbacked a bit more to the original solution with a bit of a shorter code than that. Now focuses to the last element with Shift + Tab (re-added the body boundary elements since impossible else)

packages/react/src/components/dialog/Dialog.tsx
```diff
@@ -32,73 +32,34 @@ const findFocusableDialogElements = (dialogElement: HTMLElement): NodeList =>
...
-  element.addEventListener('focus', () => focusFirstDialogElement(dialogElement));
-  document.body.insertBefore(element, document.body.firstChild);
-  return element;
+const ContentTabBarrier = ({ onFocus }: { onFocus: () => void }): JSX.Element => {
+  /* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */
+  return <div {...defaultBarrierProps} onFocus={onFocus} />;
 };
 
-const addDocumentEndTabBarrier = (dialogElement?: HTMLElement): HTMLDivElement => {
+const addDocumentTabBarrier = (position: 'top' | 'bottom', dialogElement?: HTMLElement): HTMLDivElement => {
```
minevala, line 49: A suggestion: this could be const or enum to avoid typing mistakes:
```ts
enum TabBarrierPosition {
  top = 'top',
  bottom = 'bottom'
}

if(TabBarrierPosition.top) {
 ...
}
```

minevala (commented):
Nice job, works nicely in almost every example :thumbsup: 🏅 
I noticed that the background scrolls in iPhone simulator in the documentation when the danger dialog is opened. This does not happen with other dialogs. Does it miss some class or something?

Otherwise, it is pretty much ready. I'm not sure do we need an accessibility audit. This PR did not change the functionality, it only fixed some of it.

https://user-images.githubusercontent.com/1610860/216995085-eaadd07a-f9dd-480e-b22a-d7c3f6600b02.mov

packages/react/src/components/dialog/Dialog.tsx
mrTuomoK: Yeah I'll do this quickly tomorrow 👍

mrTuomoK: on iOS there's really somthing weird happing, sometimes it scrolls, sometimes it doesn't and sometimes it suddenly stops scrolling after it 1st started scrolling 🤦

packages/react/src/components/dialog/Dialog.module.scss
```diff
@@ -3,24 +3,25 @@
...
-  height: 100vh !important;
-  overflow: hidden !important;
+    overflow: hidden !important;
 }
 
 .dialogContainer {
   --accent-line-color: var(--color-bus);
   --overlay-color: rgba(0, 0, 0, 0.3);
-
+  -webkit-overflow-scrolling: none;
```
minevala, line 12: I guess possible values are auto and touch: 
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html
But the Safari browser also suggests "unset"

minevala: Otherwise this is good to go 👍

mrTuomoK: Well seems like it damn, good spotting 👍 I'll fix

minevala (approved):
Works really well 🏅

PR #958, opened 2023-02-06
Add FAQ about Safari focus issue

minevala:
## Description
- Add accessibility FAQ-section of the Safari focus issue

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1567

## Motivation and Context
- This issue seems to raise once in a while when the team is testing the HDS components, so we thought it would be valuable for people outside the team as well

## How Has This Been Tested?
- Locally on dev machine


laurakarhu (approved):
Thank you for adding this, maybe now we'll remember those settings.

PR #960, opened 2023-02-07
Hds 1495 exportable mediaquery hooks

harriplappalainen:
## Description

- useMediaQueryLessThan and useMediaQueryGreaterThan hooks are exported and tested
- Removed console build warning about changed package name: @wessberg/rollup-plugin-ts => rollup-plugin-ts

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1495

## Motivation and Context

useMediaQueryLessThan and useMediaQueryGreaterThan hooks can be used by users as well as they support the hds breakpoints by default.

## How Has This Been Tested?
Locally checking the lib-folder's content and with hds-cra app to check that the hooks can be used

[Documentation page](https://city-of-helsinki.github.io/hds-demo/breakpoint-hooks/foundation/design-tokens/breakpoints)

To do:
- [x] Mention hooks in documentation

### DIFF
--- a/packages/react/rollup.config.js
+++ b/packages/react/rollup.config.js
@@ -3,7 +3,7 @@ import fs from 'fs';
 
 import includePaths from 'rollup-plugin-includepaths';
 import resolve from '@rollup/plugin-node-resolve';
-import ts from '@wessberg/rollup-plugin-ts';
+import ts from 'rollup-plugin-ts';
 import babel from '@rollup/plugin-babel';
 import commonjs from '@rollup/plugin-commonjs';
 import json from '@rollup/plugin-json';
--- a/packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
+++ b/packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
@@ -1,0 +1,24 @@
+import { cleanup } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+
+import { useMediaQueryLessThan, useMediaQueryGreaterThan } from '../useMediaQuery';
+
+describe('useMediaQuery', () => {
+  beforeEach(cleanup);
+  it('useMediaQueryLessThan returns correct value', () => {
+    // By default the window.innerWidth is 1024px
+    const isSmallView = renderHook(() => useMediaQueryLessThan('s'));
+    const isLargeView = renderHook(() => useMediaQueryLessThan('xl'));
+
+    expect(isSmallView.result.current).toBe(false);
+    expect(isLargeView.result.current).toBe(true);
+  });
+  it('useMediaQueryGreaterThan returns correct value', () => {
+    // By default the window.innerWidth is 1024px
+    const isLargeView = renderHook(() => useMediaQueryGreaterThan('s'));
+    const isXLargeView = renderHook(() => useMediaQueryGreaterThan('xl'));
+
+    expect(isLargeView.result.current).toBe(true);
+    expect(isXLargeView.result.current).toBe(false);
+  });
+});
--- a/packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
+++ b/packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
@@ -1,17 +1,23 @@
-import { cleanup } from '@testing-library/react';
-import { renderHook } from '@testing-library/react-hooks';
+import { cleanup, fireEvent } from '@testing-library/react';
+import { act, renderHook } from '@testing-library/react-hooks';
 
 import { useMediaQueryLessThan, useMediaQueryGreaterThan } from '../useMediaQuery';
 
 /* This test suite changes window width but it's set to default 1024 after this is done. */
 describe('useMediaQuery', () => {
-  afterAll(cleanup);
+  const defaultInnerWidth = global.innerWidth;
+  afterAll(() => {
+    cleanup();
+    global.innerWidth = defaultInnerWidth;
+  });
 
   it('useMediaQueryLessThan returns correct value', () => {
-    // Change the viewport to 500px
-    global.innerWidth = 500;
     // Trigger the window resize event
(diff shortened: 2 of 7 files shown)


packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { cleanup } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+
+import { useMediaQueryLessThan, useMediaQueryGreaterThan } from '../useMediaQuery';
+
+describe('useMediaQuery', () => {
+  beforeEach(cleanup);
+  it('useMediaQueryLessThan returns correct value', () => {
+    // By default the window.innerWidth is 1024px
```
minevala, line 9: Can the tests control the window innerWidth value? Then it could be set before the tests and reset after.
I was just thinking of a scenario when the JSDom updates and the width becomes undefined or it changes

packages/react/rollup.config.js
```diff
@@ -3,7 +3,7 @@ import fs from 'fs';
 
 import includePaths from 'rollup-plugin-includepaths';
 import resolve from '@rollup/plugin-node-resolve';
-import ts from '@wessberg/rollup-plugin-ts';
+import ts from 'rollup-plugin-ts';
```
minevala, line 6: I guess this library delivers but I was thinking can we use @rollup/plugin-typescript here?
Is it missing something?

minevala (commented):
Nice work 👍 This has been missing from HDS React lib 🎉 
I had couple of questions

packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
harriplappalainen: You're right, it would be more safe. I saw people having problems with responsive testing so I just "went where the fence was at the lowest" 😄 I'll try this out.

packages/react/rollup.config.js
harriplappalainen: I was wondering that too that why was this chosen instead of the rollup plugin. I didn't dare to try and change it 🙈 I can test it out and search why people would use this.

packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
harriplappalainen: Updated, [check it out](https://github.com/City-of-Helsinki/helsinki-design-system/pull/960/commits/ed60051466e9f36c1ecd88dea89280747ad8e449).

minevala: I'm not sure if the cleanup resets the global object, the source code defines it as: 
`Unmounts React trees that were mounted with render.`
I also get act-warnings when I run the tests. I'm not sure if it is this test or some other test that causes those 🤔 
`console.error
    Warning: An update to TestComponent inside a test was not wrapped in act(...)`

packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    // Trigger the window resize event
+    global.dispatchEvent(new Event('resize'));
+
+    const isMediumView = renderHook(() => useMediaQueryLessThan('m'));
+    expect(isMediumView.result.current).toBe(true);
+
+    // Set the viewport width again just to be safe
+    global.innerWidth = 1000;
+    // Trigger the window resize event
+    global.dispatchEvent(new Event('resize'));
```
tuomashatakka, line 35: Would [fireEvent](https://testing-library.com/docs/dom-testing-library/api-events#fireevent) be a better alternative here?

tuomashatakka (changes requested):

packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
harriplappalainen: Gah, you're right. Thanks for keeping me in check! Now these both are [fixed.](https://github.com/City-of-Helsinki/helsinki-design-system/pull/960/commits/1bf41b423e34bdf3f6483d328b72871bcaf941d4)

packages/react/src/hooks/__tests__/useMediaQuery.test.tsx
harriplappalainen: Good find! [Implemented.](https://github.com/City-of-Helsinki/helsinki-design-system/pull/960/commits/1bf41b423e34bdf3f6483d328b72871bcaf941d4)

packages/react/src/hooks/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './useMediaQuery';
+export * from './useMobile';
+export * from './useTheme';
```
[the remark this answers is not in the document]
harriplappalainen: > I guess this will export other hooks as well?
Exports only what is mentioned here in the file. All the other hooks look like they're imported but they aren't actually usable. It's weird.
> I'm not sure how useTheme should work. 
Ah maybe I should remove that altogether from here then. Better keep it just for us, for now 😄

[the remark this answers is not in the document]
harriplappalainen: I removed both of these because users don't really have use for them (as far as we know). I don't remember why I thought that they would have, I probably confused these with something else.

laurakarhu (approved):
Hooks look alright, but I guess they should be mentioned somewhere in documentation that they exist so that users would know about this fantastic job?

harriplappalainen:
> Hooks look alright, but I guess they should be mentioned somewhere in documentation that they exist so that users would know about this fantastic job?

That's a very good point 😂 Only fitting place that comes to my mind is in the Breakpoints token section, though this will only benefit React users. I'll dig around some more.

laurakarhu:
Looks great with documentation and all! 🎉

PR #961, opened 2023-02-07
HDS-831: Check components with UniqueId

timwessman:
## Description

Replaced not working `lodash.uniqueid` with local util function. `utils/getUniqueId()` gives running unique id numbers for ui components. This approach works also with server side rendered mode (which was the problem with `lodash.uniqueid`

This isn't a breaking change as it is replacement bug fix

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-831

Closes #

## Motivation and Context

`lodash.uniqueid` didn't work with SSR, but it could be replaced with this simple function

## How Has This Been Tested?
All HDS tests pass and this can be tested also with [https://github.com/City-of-Helsinki/hds-next](https://github.com/City-of-Helsinki/hds-next)

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -124,7 +124,6 @@
     "lodash.isundefined": "3.0.1",
     "lodash.pick": "^4.4.0",
     "lodash.pickby": "^4.6.0",
-    "lodash.uniqueid": "4.0.1",
     "lodash.xor": "^4.5.0",
     "memoize-one": "5.2.1",
     "postcss": "7.0.36",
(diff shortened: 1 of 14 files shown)


tuomashatakka (approved):

harriplappalainen (changes requested):
Oops nothing was discussed here, only in Slack 😅 It seems that this doesn't work in hds-next, the same error pops up. Tim also mentioned that seems like this can't be done with current React version but in React v.18+. Maybe we should cancel this or whatsamacallit?

laurakarhu:
Let's fix this with the React version 18 update.

PR #966, opened 2023-02-13
Hds 1580 helper text support

minnas:
## Description

This PR adds helper text support to below components:

- checkbox
- radiobutton
- selection group
- fieldset

## Motivation and Context

Jira: [HDS-1580](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1580) 

## How Has This Been Tested?

Tested locally


packages/core/src/components/checkbox/checkbox.stories.js
```diff
@@ -120,3 +121,11 @@ export const GroupWithParent = () => {
 }
 
 GroupWithParent.storyName = 'Group with a parent';
+
+export const WithHelperText = () => `
+    <div class="hds-checkbox">
+      <input type="checkbox" id="checkbox7" class="hds-checkbox__input" value="foo" />
+      ${getLabel('checkbox7')}
+      ${getHelperText()}
+    </div>
```
minevala, line 130: I noticed that our HDS Core examples have some accessibility flaws. The helper-text should be linked with the input. Otherwise the screen readers read it separately and the connection to the input is lost. Often the helper text is extra info for the input. 
This can be done by adding aria-describedby attribute to input like in HDS React library. The attribute can hold multiple element ids, for example <input aria-describedby="helper-text-id error-text-id" type="text"/>

In here it could be like:
```suggestion
const helperTextId = 'helper-text-id';
const getHelperText = (text = 'Assistive text') => `<span id="${helperTextId}" class="hds-radio-button__helper-text">${text}</span>`;

export const WithHelperText = () => `
    <div class="hds-checkbox">
      <input type="checkbox" id="checkbox7" class="hds-checkbox__input" value="foo" aria-describedby="${helperTextId}" />
      ${getLabel('checkbox7')}
      ${getHelperText()}
    </div>
```

packages/core/src/components/fieldset/fieldset.stories.js
```diff
@@ -74,3 +75,19 @@ export const WithBorder = () => `
...
+        Applicant information
+      </legend>
+      <div style='display: grid; grid-gap: 12px 16px; grid-template-columns: repeat(auto-fill, minmax(245px, 1fr));'>
+       ${firstNameInput}
+       ${lastNameInput}
+       </div>
+      <div style='margin-top: 12px'>
+       ${socialSecurityNumberInput}
+      </div>
+      ${getHelperText()}
```
minevala, line 91: I guess we need to ask accessibility experts how we connect the helper-text into the fieldset. Or does it even need to be connected, is it enough that it is inside the element?

packages/core/src/components/radio-button/radio-button.stories.js
```diff
@@ -35,3 +36,11 @@ export const SelectedDisabled = () => `
     </div>
 `;
 SelectedDisabled.storyName = 'Selected & disabled';
+
+export const WithHelperText = () => `
+    <div class="hds-radio-button">
+      <input type="radio" id="radio3" class="hds-radio-button__input" name="example" value="baz" disabled />
+      ${getLabel('radio5')}
+      ${getHelperText()}
```
minevala, line 44: This has a similar accessibility issue with the missing aria-describedby than Checkbox

packages/core/src/components/selection-group/selection-group.stories.js
```diff
@@ -202,3 +204,22 @@ export const WithParent = () => {
...
 
 WithParent.storyName = 'With a parent';
+
+export const WithHelperText = () => `
+  <fieldset class="hds-selection-group">
+    <legend class="hds-selection-group__legend">Label</legend>
+    <div class="hds-selection-group__items">
+      ${getCheckboxes('Default')}
+    </div>
+    ${getHelperText()}
```
minevala, line 214: We also need to ask accessibility experts how we connect the helper-text into the selection-group.

packages/core/src/components/checkbox/checkbox.css
```diff
@@ -115,6 +116,15 @@
   width: var(--icon-size);
 }
 
+/* HELPER TEXT */
+.hds-checkbox__helper-text {
+  color: var(--helper-color-default);
+  display: block;
+  font-size: var(--fontsize-body-m);
+  margin-top: var(--spacing-2-xs);
+  flex-basis: 100%;
```
minevala, line 125: Could we have a shared helper for the helper text, for example in the Core utils?
Then it could be imported and extended when needed.

packages/react/src/components/checkbox/Checkbox.tsx
```diff
@@ -108,6 +113,11 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
             {errorText}
           </div>
         )}
+        {helperText && (
+          <div className={styles.helperText} id={`${id}-helper`}>
```
minevala, line 117: The id needs to be added to the input's aria-describedby attribute.

packages/react/src/components/radioButton/RadioButton.tsx
```diff
@@ -78,6 +83,11 @@ export const RadioButton = React.forwardRef<HTMLInputElement, RadioButtonProps>(
         <label htmlFor={id} className={classNames(styles.label)}>
           {label}
         </label>
+        {helperText && (
+          <div className={styles.helperText} id={`${id}-helper`}>
```
minevala, line 87: The id needs to be added to the input's aria-describedby attribute.

minevala (commented):
Good improvements 👍 
I think we need some review from the accessibility experts as well

packages/core/src/components/fieldset/fieldset.stories.js
minnas: It's enough that text is inside <fieldset> tags

laurakarhu (approved):

packages/react/src/components/checkbox/Checkbox.tsx
```diff
@@ -24,6 +24,10 @@ export type CheckboxProps = {
    * The error text content that will be shown below the checkbox
    */
   errorText?: string;
+  /**
+   * The helper text content that will be shown below the radiobutton
```
laurakarhu, line 28: below the checkbox

packages/react/src/components/checkbox/Checkbox.tsx
```diff
@@ -113,7 +126,7 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
           type="checkbox"
           disabled={disabled}
           checked={checked}
-          aria-describedby={errorText ? `${id}-error` : null}
+          aria-describedby={ariaDescribedby.length > 0 ? ariaDescribedby.join(' ') : null}
```
laurakarhu, line 129: There is a util for creating the aria-describedBy label
https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/utils/composeAriaDescribedBy.ts
Probably would be wise to use it here as well

laurakarhu (commented):
Using the common util will solve also (at least partly) Mikas comments.

harriplappalainen (commented):
Seems to work really well 👍 I noticed minor accessibility issues in core's storybook that I pushed a fix for.

PR #975, opened 2023-02-27
Hds-1588 - normalize breaks page styles

minevala:
## Description
- Add base CSS for HDS React components to import HDS common styles without normalize.css
- Add rules from Normalize.css to each component in Core and React
- Use :where selector in Core for lower CSS specifity

## Related Issue
- https://helsinkisolutionoffice.atlassian.net/browse/HDS-1588

## Motivation and Context
- HDS React loads HDS common styles (including normalize.css) asynchronously. The normalize.css overrides some of the page custom styles unintentionally. Now the HDS common styles for React components consist of HDS CSS variables and helper classes without the normalize.css

## How Has This Been Tested?
- Locally on the dev machine

[Demo](https://city-of-helsinki.github.io/hds-demo/hds-1588-issue-normalize/?path=/story/components-accordion--card-accordion)

- [x] Add notes about React-package style changes and some guidance of how to get base styles using Core
- [x] Check build size (does not have a noticeable effect on build size)
- [x] Check SSR support


laurakarhu (approved):
Looks good and consistent to me :)

packages/react/src/styles/base.css
```diff
@@ -0,0 +1,12 @@
+@import 'hds-design-tokens';
+@import 'hds-core/lib/utils/animations.css';
+@import 'hds-core/lib/utils/helpers.css';
```
harriplappalainen, line 3: HDS doesn't work in hds-next project and there's console errors (404 Not found) about these specific packages and files. Is this hds-next settings problem or should this file be included somewhere in hds settings/build?

harriplappalainen (commented):
Amazing work 🚀 🎸 🤩  I like the normalize comments. Also at a quick glance this seems to fix the need for `.class.class` syntax when overriding styles! So this might be an even greater improvement then thought at first. 

I tested this by browsing Storybooks, running and updating tests and on hds-cra and hds-next apps and all worked except for hds-next. There's a comment about it below.

packages/react/src/styles/base.css
harriplappalainen: Oh and it would be good to test this in SSR. How do we do that again?

minevala: > HDS doesn't work in hds-next project and there's console errors (404 Not found) about these specific packages and files. Is this hds-next settings problem or should this file be included somewhere in hds settings/build?

I need to check this

> Oh and it would be good to test this in SSR. How do we do that again?

The site uses SSR support of HDS. I think we could test that with/without js

harriplappalainen: > The site uses SSR support of HDS. I think we could test that with/without js
Oh right! Doc site works well and theming too. Without javascript it doesn't load anything though.

minevala: I fixed this by addin postcss-import plugin into rollup config. It will inline imported css.

harriplappalainen (approved):
Approved! 🥇

harriplappalainen:
The .where syntax is a great change since the normalized styles are now applied with lower priority. They appear in the console as though they are targeted via element selector.

Now this works in SSR as well! I approve this yet again!

PR #982, opened 2023-03-09
fix: null cannot be returned from useEffect

sirjuan:
## Description

Returning null is not allowed from useEffect which breaks apps using current React. Returning undefined however is fine.

## How Has This Been Tested?

I tested this by modifying built code from app node_modules and running tests + app locally and verifying the problem does not exists anymore.

telegram-cloud-photo-size-4-5784963267586800381-y

telegram-cloud-photo-size-4-5784963267586800383-y


laurakarhu:
Thank your for the contribution! HDS team will review the PR and when everything is alright, proceed to include this to next possible release.

harriplappalainen:
[Here's our ticket](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1599?atlOrigin=eyJpIjoiNjMwZjE0OGJiYmJhNDY1ZmI4Y2Q1NzAwMDNiM2JiYzQiLCJwIjoiaiJ9) about this in our backlog.

mrTuomoK:
Could you @sirjuan test removing the return altogether (I'd say useEffect doesn't need to return anything)? Always tidier to remove lines of code :)

sirjuan:
Linter complains if return is removed: `Expected to return a value at the end of arrow function  consistent-return`. Also I feel that in this case explicit return with comment might be better to keep that happening again. Although I wonder whether there is a linter rule enforcing returning correct value from useEffect.

mrTuomoK:
Ah good point, there might be some other catch in it then, we need to look it a bit closer.

sirjuan:
Enabling strict mode in tsconfig would catch that it seems, but it also catches a few other things as well 😄 

`Found 348 errors in 74 files.`

harriplappalainen (approved):
Since something must be returned because there's a return inside the if, this is a good fix 👍 Thank you for your contribution! We had already received feedback about this so your help is greatly appreciated 😊 

We'll merge this to master when we're preparing for a new release.

PR #985, opened 2023-03-10
Fix dialog for react 18 users

VilleMiekkoja:
Dialog crashed when opened with react 18. I guess react 18 does not like that useEffect returns null, so changed it to undefined which fixes the issue.


harriplappalainen:
Long time no see, Ville! Thanks for the contribution but this is actually a [duplicate](https://github.com/City-of-Helsinki/helsinki-design-system/pull/982). But keep them coming, we welcome HDS-alumnis PRs anyday 😄

PR #989, opened 2023-03-15
Hds 1576 rectangular tag and footer deprecation

harriplappalainen:
## Description

Mark rectangular Tag as deprecated in code and doc site.

I set RoundedTag and LargeRoundedTags as deprecated because the Tag will be rounded by default and the large size can be handled with `size` prop.

**Bonus:** I changed the Footer component's doc site notification to alert and reworded it about upcoming breaking changes.  

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1576

## Motivation and Context

It's good to notify users beforehand of upcoming breaking changes so they can mull things over and adapt.

## How Has This Been Tested?
Doc site on localhost, testing the components' use in hds-cra
[Demo](https://city-of-helsinki.github.io/hds-demo/tag-deprecate/components/tag)

**How the RoundedTag looks to users:**
Screenshot 2023-03-15 at 14 12 47

**How the default Tag looks to users**.
Screenshot 2023-03-15 at 14 13 58

### DIFF
--- a/packages/core/src/components/tag/tag.css
+++ b/packages/core/src/components/tag/tag.css
@@ -48,7 +48,7 @@
   padding-left:  var(--tag-padding);
 }
 
-.hds-tag--rounded-corners > button {
+.hds-tag--rounded-corners .hds-tag__delete-button {
   padding-right: calc(var(--tag-padding) / 2);
 }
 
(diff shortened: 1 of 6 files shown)


harriplappalainen:
Yes, it should be deprecated. I created own [ticket](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1649) for it.

PR #1002, opened 2023-03-23
Fix normalize styles

minevala:
## Description
- Use class-selector instead of where-selector

## Motivation and Context
- Postcss compose only works with single class-selector styles. 
## How Has This Been Tested?
- locally on dev machine


harriplappalainen (approved):
Approved 👍

PR #1009, opened 2023-03-28
Feature 447 Token for focus color

timwessman:
## Description

Created `color-focus-outline` -token for focus color. Modified all focusable core components to use it

## Related Issue

[https://helsinkisolutionoffice.atlassian.net/browse/HDS-447](https://helsinkisolutionoffice.atlassian.net/browse/HDS-447)

Closes #

## Motivation and Context

This makes a change of focus color possible.

## How Has This Been Tested?

All focus colors should look the same as before. 

To test if the new token works, you can try to change the focus color of all focusable components just by changing the color value in `/design-tokens/tokens/color/ui/focus.json`.  Build project and go to see your color change in Storybook

### DIFF
--- a/packages/core/src/components/button/button.css
+++ b/packages/core/src/components/button/button.css
@@ -306,7 +306,7 @@ input[type="submit"].hds-button--small {
   --color-focus: var(--color-white);
   --color-hover-focus: var(--color-white);
   --color-disabled: var(--color-white);
-  --focus-outline-color: var(--color-coat-of-arms);
+  --focus-outline-color: var(--color-focus-outline);
   --submit-input-focus-gutter-color: var(--color-white);
 }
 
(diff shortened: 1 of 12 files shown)


mrTuomoK:
Is it truly that they are the same color for both focus and background? Well in the ticket it says: "Create separate **tokens** for focus colour (background and border) and change components to use them." 🤷 Dunno 🤷

timwessman:
I didn't understand `background` -part. Focus don't normally change the background of components (only outline or border), and I didn't find that kind of behavior from HDS either.

mrTuomoK:
Yeah I don't know, maybe we'll need to ask the HDS-people too.

laurakarhu:
That background thing was a weird blurb, I don't know how it got there :D I removed it from the issue as well, so you can forget that part.

mrTuomoK (approved):
Seems legit

PR #1029, opened 2023-05-10
Revert "Icon library build"

harriplappalainen:
Reverts City-of-Helsinki/helsinki-design-system#968.

The PR caused hds-react build to fail.


mrTuomoK (approved):
Works, maybe another review by @laitine and then it's good to go.

PR #1032, opened 2023-05-16
HDS-1631 Refactor getChildElementsEvenIfContainersInbetween with flatten

mikkojamG:
## Description

Refactoring `getChildElementsEvenIfContainerInbetween` which does not work correctly at the moment if there are more than one containers within children. Also change the function name to use plural `containers`, since theoretically there can be more than on container.

## Motivation and Context

Tested on local machine and running unit tests.


harriplappalainen (approved):
Everything works now, great job 🤝 🚀

PR #1036, opened 2023-05-22
Feature: Added optionKeyField to select and combobox

### LINKED ISSUE #836: Using optionLabel as key in DropdownMenuItem causes issue when options don't have unique labels
## Describe the bug
[DropdownMenuItem](https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/internal/dropdownMenu/DropdownMenu.tsx#L41) uses item label (optionLabel) [as the key](https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/internal/dropdownMenu/DropdownMenu.tsx#L179) for the list elements. Although it isn't ideal to have items with same labels there are situations where it might happen (such as dynamic lists with peoples names).

If the said situation happens where there are more than one element with the same key (optionLabel), and the DropdownMenu options change, the DropdownMenu breaks.

## How to reproduce
1. DropdownMenu contains more than one element with the same label
2. DropdownMenu options are changed
3. DropdownMenu ends up in a broken state and only recovers from it when rendered again

## Expected behavior
DropdownMenu keeps working normally if options change regardless of duplicate labels

## Possible solution
optionLabel should not be used as the key for the DropdownMenuItem component. Instead a value from items which will always be unique between the list elements should be used instead. This could e.g. be id or key variable delivered with the item and then if neither are present the optionLabel could be used as a fallback.

## Screenshots
Combobox (DropdownMenu) in a broken state
hds-broken-dropdownmenu

## Device information
**Desktop (please complete the following information)**:
> - OS: iOS
> - Browser: Chrome
> - Version: 105.0.5195.125

lauriLukkarinen:
Added optional optionKeyField property to Select and Combobox.

## Description

`Select` and `Combobox` dropdown components now have an optional string property `optionKeyField`. If provided, it is used as a key of single option object to get a unique key for a `DropdownMenuItem`. If not provided, the previous implementation of using the `optionLabelField`, which is by default `label`, is used.

## Related Issue

#836 

Closes #836 

## Motivation and Context

Select components are throwing errors, when multiple options have the same key. This is because select components use 'label' field from option data as key. When handling data, what is visible often does not match what is handled programmatically, e.g. there might be multiple options with the same label but with different internal value. Select component should provide an option to define the key field to allow multiple identical labels.

## How Has This Been Tested?

Tested locally

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -68,6 +68,7 @@
     "@typescript-eslint/eslint-plugin": "5.10.2",
     "@wessberg/rollup-plugin-ts": "2.0.4",
     "babel-jest": "^26.0.1",
+    "babel-plugin-polyfill-corejs2": "^0.4.1",
     "babel-plugin-require-context-hook": "1.0.0",
     "chalk": "4.0.0",
     "eslint": "7.1.0",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -68,7 +68,6 @@
     "@typescript-eslint/eslint-plugin": "5.10.2",
     "@wessberg/rollup-plugin-ts": "2.0.4",
     "babel-jest": "^26.0.1",
-    "babel-plugin-polyfill-corejs2": "^0.4.1",
     "babel-plugin-require-context-hook": "1.0.0",
     "chalk": "4.0.0",
     "eslint": "7.1.0",
--- a/packages/react/src/components/dropdown/combobox/Combobox.tsx
+++ b/packages/react/src/components/dropdown/combobox/Combobox.tsx
@@ -97,6 +97,7 @@ export const Combobox = <OptionType,>(props: ComboboxProps<OptionType>) => {
     label,
     onBlur = () => null,
     onFocus = () => null,
+    optionKeyField,
     optionLabelField = 'label',
     options = [],
     placeholder,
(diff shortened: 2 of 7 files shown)


packages/react/package.json
```diff
@@ -68,6 +68,7 @@
     "@typescript-eslint/eslint-plugin": "5.10.2",
     "@wessberg/rollup-plugin-ts": "2.0.4",
     "babel-jest": "^26.0.1",
+    "babel-plugin-polyfill-corejs2": "^0.4.1",
```
harriplappalainen, line 71: I don't think this is needed in this PR? It looks like it could be an improvement but maybe in a separate PR where it's implemented and tested.

harriplappalainen (commented):
Big thanks for the contribution! 🙏 🥳 Your code looks good and works but I was wondering about the addition of `babel-plugin-polyfill-corejs2` package.

packages/react/package.json
lauriLukkarinen: Sure! I had some problems to start the project without adding this plugin, but I ran the builds again just now and everything seems to work fine now. Not sure what caused the issue then, but I'll remove this as unnecessary 🙂

harriplappalainen (approved):
Approved! 🎉

mrTuomoK (approved):
Looks good to go 👍  thx :)

PR #1039, opened 2023-05-24
HDS-1718-fix-fileInput-validateAccept

mrTuomoK:
## Description

Fix FileInput not accepting files that are missing `file.type` though given the correct accept-prop. Not all files have the `file.type` (in this case `.dwg` -files were the ones that were noticed). In this case the `file.type`: string is and empty string which caused our `validateAccept` to return an error.

Also fixed the documentation site code example and added info on FileInput accept-parameter.

## Related Issue

[HDS-1718](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1718)

## How Has This Been Tested?
On local machine

image

[HDS-1718]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/site/src/docs/components/file-input/index.mdx
+++ b/site/src/docs/components/file-input/index.mdx
@@ -47,7 +47,7 @@ export const FileInputExample = () => {
 - Dragging and dropping files can be allowed by using the `dragAndDrop` property.
 - The FileInput uses an info text element to inform the user about the status of the component. You can read more about the usage of the info text element in the <InternalLink href="/patterns/forms/form-validation#3-info-message">HDS Form validation pattern documentation page</InternalLink>.
 - The FileInput component resolves human-readable file size abbreviations based on a binary system. The file size texts are shown after the selected file name and in maxSize messages if maxSize validation is used.
-- The accept parameter only checks and allows selecting the correct file types according to the file extension(s) or file types and does not do any in-depth analysis. Any safety and/or other checks should be done on the receiving end if necessary.
+- The accept parameter only checks and allows correct file types according to the file extension(s) or file types, instead of doing any in-depth analysis.
 
 ### Variations
 


mrTuomoK:
Seems that no matter what the mimeType/file.type is, the check has always passed through 🤦 ... this is also very problematic to even do properly since `file.type` doesn't necessarily have almost anything in common with the file extension. (for example `.mp3` => mimeType is `audio/mpeg` or `.ppt` => `application/vnd.ms-powerpoint` and so on)

site/src/docs/components/file-input/index.mdx
```diff
@@ -47,6 +47,7 @@ export const FileInputExample = () => {
 - Dragging and dropping files can be allowed by using the `dragAndDrop` property.
 - The FileInput uses an info text element to inform the user about the status of the component. You can read more about the usage of the info text element in the <InternalLink href="/patterns/forms/form-validation#3-info-message">HDS Form validation pattern documentation page</InternalLink>.
 - The FileInput component resolves human-readable file size abbreviations based on a binary system. The file size texts are shown after the selected file name and in maxSize messages if maxSize validation is used.
+- The accept parameter only checks and allows selecting the correct file types according to the file extension(s) or file types and does not do any in-depth analysis. Any safety and/or other checks should be done on the receiving end if necessary.
```
mikkojamG, line 50: I find the first sentence bit long and hard to understand 🙂 Would suggest simplifying it a little bit, something like: `The accept property only checks and allows correct file types according to the file extension(s) or file types, instead of doing any in-depth analysis.`

mikkojamG (approved):
Left one comment, overall looks good 👍

site/src/docs/components/file-input/index.mdx
mrTuomoK: Goood good, I'll mod 👍

PR #1041, opened 2023-05-26
Site hero fix

NikoHelle:
## Description

Wrong hero type on front page.

Also theme does not work always, so moved theme styles to style attribute.


mrTuomoK (approved):
Fine by me 👍

PR #1046, opened 2023-06-06
Fix FileInput maximum file size to maxSize

corvidian:
## Description

Currently, if maxSize is set, the file added in FileInput needs to be smaller than the file. If adding a file exactly the size of the limit, we get a confusing validation message:
> - File, test.png, is too large (100 B). The maximum file size is 100 B.

Fix this by allowing the file to be smaller or equal to maxSize.


laurakarhu (approved):
Thank you!

mrTuomoK (approved):
ok

harriplappalainen (approved):
Nice find and fix! Thank you for contributing 🎉

Issue #1040, opened 2023-05-24
DayInput/DatePicker custom style for specific dates


kieferbonk:
## To which feature the request relates to
DatePicker in DayInput component

## What
In tilavarauspalvelu, we were plannig a date picker with a prop to modify styles or element class of specific dates. Essentially a function that could return styles or class to individual date in the datepicker component - much like the `isDateDisabledBy` prop. 

## Why
The component could display booking status much more expressively in event style calendars.

## Related material
![Screenshot 2023-05-24 at 14 04 46](https://github.com/City-of-Helsinki/helsinki-design-system/assets/1003723/7e84c608-ac11-4650-9aed-64e97644c42d)

kieferbonk:
I will make a pr shortly to showcase the functionality.

kieferbonk:
PR can be found here: https://github.com/City-of-Helsinki/helsinki-design-system/pull/1047

PR #1047, opened 2023-06-09
Feature: Datepicker custom styling

kieferbonk:
## Description

In Tilavaraus-project, we have a need for some availability information in our event calendar date picker and would like to use the DateInput component for it. Thus I added callback prop to inject custom class names to specific dates.

## Motivation and Context

Ability to change day element appearance in date picker.

## How Has This Been Tested?

Locally; with basic intergration tests; added a story to demonstrate the feature
[Demo](https://city-of-helsinki.github.io/hds-demo/dateinput-custom-styles/?path=/story/components-dateinput--with-custom-day-styles) (styles are working here with the <style> tag)
[Docs demo](https://city-of-helsinki.github.io/hds-demo/dateinput-custom-styles-docs/components/date-input/customisation/)

Screenshot 2023-06-09 at 16 17 25


laurakarhu (changes requested):
https://github.com/City-of-Helsinki/helsinki-design-system/issues/1040 In the issue related to this PR, there is also the legend for different classes. This is important feature for accessibility, so the support for legends should be added as well. 

- [x] change example color
- [x] add a legend area
- [x] create legend items for all used classes (automatically or so that developer creates them)

laurakarhu:
@kieferbonk do you wish to continue on this or should HDS team take it from here?

kieferbonk:
> @kieferbonk do you wish to continue on this or should HDS team take it from here?

Sure, I can follow up on this. Thanks for the comments!

kieferbonk:
Added a legend like so:
Screenshot 2023-06-15 at 13 05 44
Screenshot 2023-06-15 at 13 05 17

laurakarhu (approved):
Looks good to me, thank you for the contribution!

harriplappalainen (commented):
Top notch work! 🥇 Hard to find anything to improve, except for prop docs. See my comments below.

harriplappalainen (commented):
Top notch work! 🥇 Hard to find anything to improve, except for prop docs. See my comments below.

laurakarhu:
We'll check the accessibility with our specialists and then get this to our release queue.

packages/react/src/components/dateInput/DateInput.stories.scss
```diff
@@ -0,0 +1,9 @@
+.odd--day,
+.odd--color {
+    background: var(--color-suomenlinna);
+}
+
+.even--day,
+.even--color {
+    background: var(--color-black-10);
+}
```
harriplappalainen, line 9: Actually these styles aren't loaded in remote environment, at least in our demo site 😮 They do work if they are included inside a style tag in the story's return however.

kieferbonk: Seems like a loader issue, but maybe sorting that is out of scope and I'll move the styles inline as you suggested.

laurakarhu:
We received feedback on this contribution from accessibility audit so this contribution still requires some work.
[Accessibility notes](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HDS/pages/8347385872/2023-08-18+Saavutettavuustunti#Date-picker-kontribuutio)
- [ ]

packages/react/src/components/dateInput/DateInput.stories.tsx
```diff
@@ -190,5 +190,67 @@ export const WithSelectedDisabledDates = (args) => {
...
+      <DateInput
+        {...args}
+        value={value}
+        onChange={setValue}
+        helperText={helperText}
+        setDateClassNames={setDateClassNames}
+        setDateAriaDescribedBy={setDateAriaDescribedBy}
+        legend={[
+          { id: hasSomeSpaceLeft, label: 'Some free timeslots available', color: someSpaceColor },
+          {
```
NikoHelle, line 245: The link between `legend`, `setDateClassNames` and `setDateAriaDescribedBy` should be documented in detail and/or the examples should return legend values from the functions.

I'd hoist the legend props up and replace the code with
```

const someSpaceLeftDate: Legend =   {
  id: hasSomeSpaceLeft, 
  label: 'Some free timeslots available', 
  color: someSpaceColor
}
            
const littleSpaceLeftDate: Legend = {
  id: hasLittleSpaceLeft,
  label: 'Only a few free timeslots available',
  color: littleSpaceColor
 }            
            
const getDateLegendId = (date: Date) => {
 const day = date.getDate();
 if (dayHasSomeSpace(day)) return someSpaceLeftDate.id;
 if (dayHasLittleSpace(day)) return littleSpaceLeftDate.id;
 return undefined
}

const setDateClassNames = (date: Date) => {
   const dateLegendId = getDateLegendId(date)
   return dateLegendId? `${dateLegendId}--day` : undefined;
};

const setDateAriaDescribedBy = (date: Date) => {
   return getDateLegendId(date)
};

const legends=[someSpaceLeftDate, littleSpaceLeftDate]
```

And maybe even

```
const createStyles = ()=>{
  return legends.map(legend=>{
    return .${legend.id}--day {
              background: ${legend.color};
          }
  }).join(" ");
}
```

packages/react/src/components/dateInput/DateInput.test.tsx
```diff
@@ -411,12 +411,109 @@ describe('<DateInput /> spec', () => {
...
+    await act(async () => {
+      userEvent.click(screen.getByLabelText('Choose date'));
+    });
+
+    const oddButton = container.querySelector('button[data-date="2021-01-01"]') as HTMLElement;
+    const evenButton = container.querySelector('button[data-date="2021-01-02"]') as HTMLElement;
+
+    expect(oddButton.getAttribute('aria-describedby')).toBe('odd');
+    expect(evenButton.getAttribute('aria-describedby')).toBe('even');
+  });
```
NikoHelle, line 518: Great tests!

NikoHelle (changes requested):

NikoHelle:
VoiceOver works great with special dates!

When a special date with its own bg color is selected, the bg color stays and is seen as a 1px "borders". Looks a bit odd, but I think designers should comment is that good or bad.

packages/react/src/components/dateInput/DateInput.tsx
```diff
@@ -55,9 +55,25 @@ export type DateInputProps = Omit<TextInputProps, 'onChange'> & {
...
   isDateDisabledBy?: (date) => boolean;
+  /**
+   * Function to set class names for dates.
+   */
+  setDateClassNames?: (date: Date) => string | undefined;
+  /**
+   * Legend items for the date picker.
+   * */
+  legend?: {
+    color: string;
```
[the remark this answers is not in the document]
harriplappalainen: I think legend is the term for the section which explains the colours. Can @SillanmakiA or @laurakarhu confirm?

harriplappalainen:
> When a special date with its own bg color is selected, the bg color stays and is seen as a 1px "borders". Looks a bit odd, but I think designers should comment is that good or bad.

Designers commented that the custom bg colour should be removed because it's easier that way accessibility-wise.

packages/react/src/components/dateInput/components/datePicker/types/DayPickerProps.ts
```diff
@@ -11,6 +11,16 @@ export type DayClickEventHandler = (day: Date, e: React.MouseEvent) => void;
  */
 export type MonthChangeEventHandler = (month: Date, e: React.MouseEvent) => void;
 
+/**
+ * Props for legend item.
+ * @param {{ id: string, color: string, label: string }}
+ */
+export type LegendItem = {
+  id: string;
+  color: string;
```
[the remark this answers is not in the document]
harriplappalainen: Hmm, you have a good point. Commonly only `id` prop is used in HDS components but maybe we can make an exception here because it might not be clear what the id means in this context. The prefix shouldn't be added here because user needs the id when setting the `setDateAriaDescribedBy` function for specific dates.

packages/react/src/components/dateInput/DateInput.tsx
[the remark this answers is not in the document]
laurakarhu: Yes, as Harri said. The legend of a graph reflects the data displayed, in this case the graph happens to be a "calendar".

packages/react/src/components/dateInput/DateInput.tsx
```diff
@@ -55,9 +55,25 @@ export type DateInputProps = Omit<TextInputProps, 'onChange'> & {
...
+  /**
+   * Set custom theme for specific dates.
+   */
+  setDateTheme?: (date: Date) => DateCustomTheme | undefined;
+  /**
+   * Legend items for the date picker.
+   * @example
+   * ```ts
+   * legend=[{ elementId: 'legend-item-1', label: 'Text label', color: #c3c3c3 }, { elementId: 'legend-item-2', label: 'Text label', color: #d6d66d }}]
+   * ```
```
NikoHelle, line 70: Nice, you can add ts examples here! Should the value of the `color` also be quoted like other props? Plus there's double `}} `in the end, but only one curly brace is open. Too bad linter can't see these.

site/src/docs/components/date-input/customisation.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+
+  ```jsx
+import { DateInput } from 'hds-react';
+
+() => {
+    const dayHasSomeSpace = (day) => day === 12 || day === 15 || day === 24 || day === 25;
+    const dayHasLittleSpace = (day) => day === 3 || day === 4 || day === 17 || day === 19;
+
+    const setDateClassNames = (date) => {
```
NikoHelle, line 30: The storybook version has better example now. Too bad we have to support two sets of examples. The storybook example should be used here too, because it shows better the relations between `legend`, `setClassNames` and `setDateAriaDescribedBy`.

packages/react/src/components/dateInput/components/day/Day.tsx
```diff
@@ -42,7 +46,18 @@ export const Day = ({ day }: DayProps) => {
...
+      ? useTheme<DateCustomTheme>(styles['hds-datepicker__day'], !isDisabled && setDateTheme && setDateTheme(day))
+      : undefined;
+  const customWrapperThemeClass =
+    !isDisabled && setDateTheme
+      ? useTheme<DateCustomTheme>(
+          styles['hds-datepicker__day__wrapper'],
+          !isDisabled && setDateTheme && setDateTheme(day),
+        )
+      : undefined;
+  console.log(customButtonThemeClass);
```
NikoHelle, line 60: Missed console.log

site/src/docs/components/date-input/customisation.mdx
```diff
@@ -0,0 +1,77 @@
...
+        const day = date.getDate();
```
NikoHelle, line 31: image

I'm not sure is the demo old version or broken

packages/react/src/components/dateInput/DateInput.stories.tsx
```diff
@@ -190,5 +191,80 @@ export const WithSelectedDisabledDates = (args) => {
...
+        '--date-background': someSpaceLeftDate.color,
+      };
+
+    if (legendId === littleSpaceLeftDate.elementId)
+      return {
+        '--date-background': littleSpaceLeftDate.color,
+      };
+    return undefined;
+  };
+
```
NikoHelle, line 238: I think the theme version is better than classNames. Very good change! 👍

NikoHelle: Too bad each date has own `custom-theme-xxx`, even if colors match. But that's a `useTheme`-problem. And new  styles are created on each time picker is opened. `useTheme` should support fixed styles somehow.

harriplappalainen: I think that theme is better too even with all the troubles it brings. With className and the way DatePicker is built, it proved difficult to override the background-color in all cases. When setting the custom class to Day component, it didn't override, but in the inside span it did. But user may not want to only target the span element. With theme, the options are limited but they work as expected.

packages/react/src/components/dateInput/DateInput.tsx
harriplappalainen: Nice find, fixed!

site/src/docs/components/date-input/customisation.mdx
harriplappalainen: Forgot to update docs, now fixed.

packages/react/src/components/dateInput/DateInput.stories.tsx
harriplappalainen: I limited the DateCustomTheme variables to only `'--date-background'` and `'--date-color'` because only those could override Day component's styles. Seems like the other CSS variables inside `DatePicker.module.scss` should come through a generic `theme` prop, I reckon 🤔

site/src/docs/components/date-input/customisation.mdx
harriplappalainen: Forgot to update demo though 🤦 Will do later.

harriplappalainen:
More to do from a11y audit:

- support Day borders in theme
- document that user can use borders to display differences in days more clearly. If many colours are used, the selected colour should be explained as well in the Legend
- custom styles story breaks when changing months

site/src/docs/components/date-input/customisation.mdx
harriplappalainen: Demo updated.

packages/react/src/components/dateInput/DateInput.stories.tsx
harriplappalainen: Well I changed the implementation back to className because `useTheme` was causing the issues mentioned here and changing months caused an error `Rendered more/fewer hooks than on previous render` because the amount of days per month changed. `setDateClassName` works now with DateInput CSS variables. I tried to style Accordion component's theme variables with className which sets the variables but they aren't available there. I think the reason they work in DateInput is that the `DatePicker.module.scss` declares variables in `:root` which puts them in the root of the component it seems. The docs say it puts the included styles in the root of the DOM, which would be `<html>` but I guess with React they are encapsulated in the component.

laurakarhu (approved):
Looks good to me, and docs seem clear! Thank you!

packages/react/src/components/dateInput/DateInput.stories.tsx
NikoHelle: Customizing dates with css vars via classNames works very well. 👍  

I was curious, if there is an actual need to use :`root` as the base of the css vars for the date input. This is not related to this PR, but if vars are moved from `:root` to `.hds-datepicker` everything still works.

Some of the vars in the `:root` are too generic (also not related to this PR) like `horizontal-spacing,  vertical-spacing,  table-width` and could easily cause conflicts when same names are used somewhere else.

Bottom line: css vars specific to an element should be enclosed to that element, not root. Should we make a new ticket for moving the vars?

packages/react/src/components/dateInput/DateInput.stories.tsx
```diff
@@ -209,3 +211,61 @@ Success.args = {
...
+  const legend = [littleSpaceLeftDate];
+  return (
+    <>
+      <style>
+        {`
+          .little-space-left {
+              --date-background: var(--color-summer-medium-light);
+              --date-border: 1px solid black;
+              --horizontal-spacing: 1px;
```
NikoHelle, line 254: `horizontal-spacing` is only used in `hds-datepicker` element's padding. Setting it in here has no effect, because this is a child of that element and out of scope. Tried to remove this and also set it to 100px, but either has no effect, so this can be removed?

site/src/docs/components/date-input/customisation.mdx
```diff
@@ -0,0 +1,95 @@
...
+  const legend = [littleSpaceLeftDate];
+  return (
+    <>
+      <style>
+        {`
+          .little-space-left {
+              --date-background: var(--color-summer-medium-light);
+              --date-border: 1px solid black;
+              --horizontal-spacing: 1px;
+          }
```
NikoHelle, line 78: This var is not used either.

site/src/docs/components/date-input/customisation.mdx
```diff
@@ -0,0 +1,95 @@
...
+      </style>
+      <DateInput
+        value={value}
+        onChange={setValue}
+        helperText={helperText}
+        setDateClassName={setDateClassName}
+        setDateAriaDescribedBy={setDateAriaDescribedBy}
+        legend={legend}
+        style={{ width: 400 }}
+      />
```
NikoHelle, line 89: 400px?

NikoHelle (changes requested):

NikoHelle:
image

I tried a worst case scenario for the legend texts. Should the colour box align to the top or middle?

Should the colour boxes have the same border as the dates?

site/src/docs/components/date-input/customisation.mdx
harriplappalainen: I just set some value so the input isn't super wide.

packages/react/src/components/dateInput/DateInput.stories.tsx
harriplappalainen: > Some of the vars in the :root are too generic (also not related to this PR) like horizontal-spacing,  vertical-spacing,  table-width and could easily cause conflicts when same names are used somewhere else.
Bottom line: css vars specific to an element should be enclosed to that element, not root. Should we make a new ticket for moving the vars?

No they are enclosed to that element, not root. You can try to access DatePicker's variable in the story in another element and you'll see that they are not available.

harriplappalainen: > I was curious, if there is an actual need to use :root as the base of the css vars for the date input. This is not related to this PR, but if vars are moved from :root to .hds-datepicker everything still works.

Okay well that's good! I'll add this to the Jira doc I made about theming.

site/src/docs/components/date-input/code.mdx
```diff
@@ -160,3 +160,6 @@ Also, note that this component is an input. All features supported by the HDS Te
 | `initialMonth`                                   | The initial month open in calendar.                            | `Date`                                       | -                                                |
 | `disableDatePicker`                              | If set to true, the date picker button is hidden.              | `boolean`                                    | false                                            |
 | `disableConfirmation`                            | If set to true, the date picker does not require confirmation. | `boolean`                                    | false                                            |
+| `setDateTheme`                                   | Set CSS variables with a custom theme for specific dates.      | `function`                                   | -                                                |
```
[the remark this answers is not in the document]
harriplappalainen: Thanks, fixed now.

packages/react/src/components/dateInput/DateInput.stories.tsx
harriplappalainen: Thanks, fixed now.

site/src/docs/components/date-input/customisation.mdx
harriplappalainen: Fixed.

packages/react/src/components/dateInput/DateInput.tsx
```diff
@@ -55,9 +55,25 @@ export type DateInputProps = Omit<TextInputProps, 'onChange'> & {
...
   isDateDisabledBy?: (date) => boolean;
+  /**
+   * Set classNames for specific dates.
+   */
+  setDateClassName?: (date: Date) => string | undefined;
+  /**
+   * Legend items for the date picker.
+   * @example
+   * ```ts
+   * legend={[{ elementId: 'legend-item-1', label: 'Text label', relatedClassName: '.custom-day' }, { elementId: 'legend-item-2', label: 'Text label', color: '#d6d66d' }]}
```
NikoHelle, line 69: Typo: `relatedClassName: '.custom-day'` should be without the dot.

packages/react/src/components/dateInput/DateInput.stories.tsx
```diff
@@ -209,3 +211,66 @@ Success.args = {
...
+  const getDateLegendId = (date: Date) => {
+    const day = date.getDate();
+    if (dayHasLittleSpace(day)) return littleSpaceLeftDate.elementId;
+    if (isSameDay(parse(value, dateFormat, date), date)) return selectedDate.elementId;
+    return undefined;
+  };
+
+  const setDateClassName = (date: Date) => {
+    const legendId = getDateLegendId(date);
+    if (legendId === littleSpaceLeftDate.elementId) return 'little-space-left';
```
NikoHelle, line 245: Might be clearer if this returned  `littleSpaceLeftDate.relatedClassName`

packages/react/src/components/dateInput/DateInput.stories.tsx
NikoHelle: > No they are enclosed to that element, not root. You can try to access DatePicker's variable in the story in another element and you'll see that they are not available.

Are they enclosed? I've not used `:root`, but adding a style tag at the end of DOM with new value for `vertical-spacing` alters the DatePicker too.

image

Chrome shows the DatePicker's `:root` values in every element in the inspector. I can use `var(--selected-date-background)` outside DatePicker.

image

My previous argument was based on the specs: "The :root [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) matches the root element of a tree representing the document."

And I was disappointed that they weren't enclosed. 😄  There is `:scope`, but it is not supported very well.

harriplappalainen: Well, crap. I tested if the variable was available just in a random class, which it wasn't. So the variables are only inside the `:root` selector. That isn't handy of course in case of accidental overriding of existing variables. Well, maybe we don't make moves in this PR but I'll edit the confluence docs.

site/src/docs/components/date-input/customisation.mdx
NikoHelle: Sorry, I meant should the value be "400px" instead of "400" 😄

NikoHelle:
Looks good now. I'd make a ticket for moving the `:root` variables to the component. This is the only component that uses ":root" so it is not widely used.

NikoHelle (approved):

site/src/docs/components/date-input/customisation.mdx
harriplappalainen: I see! Set it to pixels.

harriplappalainen:
Thank you for good reviews 🙏 

> Looks good now. I'd make a ticket for moving the :root variables to the component. This is the only component that uses ":root" so it is not widely used.

Made [a ticket](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2030)!

harriplappalainen:
- [ ] Add to changelog

PR #1057, opened 2023-06-30
Change old hds-core style imports

NikoHelle:
Imports in other components were already changed in another PR.

These components were missed, because their PR were open at the same time.

## How Has This Been Tested?

Visually with Loki


harriplappalainen (approved):
All good 👍

PR #1059, opened 2023-07-03
Hds 1787 prerelease script

harriplappalainen:
## Description

New workflow for prereleases which asks for prerelease version (alpha or beta) and publishes it with that tag to npm so it is not set as the latest version. The workflow calls the normal release script in package.json but with additional options so as to not set the version as the latest one.

First I set up a test drive to see that the input parameter (alpha/beta) travels through properly before actually testing releases.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1787

## How Has This Been Tested?
By running the workflow remotely via Github CLI. [Here's the latest test drive run](https://github.com/City-of-Helsinki/helsinki-design-system/actions/runs/5440715843/jobs/9893921482)

Here's how to run in gh cli: `gh workflow run prerelease.yml --ref hds-1787-prerelease-script -f prerelease_stage=alpha`
The script is usable from any branch, the `--ref [branch]` part takes care of that. `-f prerelease_stage=alpha` is passing the parameter part.


package.json
```diff
@@ -13,6 +13,8 @@
     "build:react": "lerna run --scope hds-react build",
     "build:site": "lerna run --scope site build",
     "clean": "lerna clean --yes && rimraf node_modules/ packages/core/lib packages/core/storybook-static packages/react/lib packages/react/storybook-static packages/design-tokens/lib site/.cache site/public",
+    "prerelease": "yarn run lerna publish from-package --yes —dist-tag $npm_config_stage",
```
harriplappalainen, line 16: Dear reviewers,

Whaddya say if we remove this `prerelease` script and call the `release` script straightaway from the workflow but with `-- --yes --dist-tag ${{ github.event.inputs.prerelease_stage }}` at the end? Because anything after the first `--` is passed to the script as is.

NikoHelle: Sounds better to use the `release` script and pass more props to it via `prerelase`

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: '${{ steps.nvmrc.outputs.NODE_VERSION }}'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: install dependencies
+        run: yarn
+
+      - name: release npm packages
+        run: yarn run testdrive -- --dist-tag ${{ github.event.inputs.prerelease_stage }}
```
NikoHelle, line 31: When will this be changed to `yarn run release`?

harriplappalainen: I kept it in case reviewers want to test run the workflow. I can change it to release now.

NikoHelle:
There is this comment in the output 

"warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts."

Is the extra "--" useless in this scenario?

harriplappalainen:
> There is this comment in the output
> 
> "warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts."
> 
> Is the extra "--" useless in this scenario?

I ran it without the extra -- and it does indeed deliver the rest of the command properly:
> Run yarn run testdrive --dist-tag alpha
> yarn run v1.22.19
> $ echo "Hello, " --dist-tag alpha
> Hello,  --dist-tag alpha
> Done in 0.06s.

mrTuomoK (approved):
Goood goood 👍

PR #1061, opened 2023-07-04
HDS-X: add cookie consent HTML and CSS to core

Jussiles:
## Description

This is the first PR for adding cookie consent to HDS core. There will be more changes including JS coming in this or other PR:s. There are some TODOs in code that still need some work.

- Add cookie consent css to core
- Add cookie consent HTML example to demo site

## Motivation and Context

Create core version of cookie consent that sites that don't use React can use cookie banner without using React.

## How Has This Been Tested?

Locally

### DIFF
--- a/packages/core/src/utils/helpers.css
+++ b/packages/core/src/utils/helpers.css
@@ -1,12 +1,11 @@
-.visually-hidden {
+.hiddenFromScreen {
   border: 0;
-  clip-path: inset(50%);
+  clip: "rect(0 0 0 0)";
   height: 1px;
   margin: -1px;
   overflow: hidden;
   padding: 0;
   position: absolute;
-  white-space: nowrap;
   width: 1px;
 }
 
--- a/packages/core/src/utils/helpers.css
+++ b/packages/core/src/utils/helpers.css
@@ -9,6 +9,17 @@
   width: 1px;
 }
 
+.visually-hidden {
+  clip: rect(1px, 1px, 1px, 1px);
+  clip-path: inset(50%);
+  max-height: 1px;
+  max-width: 1px;
+  opacity: 0;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+}
+
 .button-reset {
   background: none;
   border: none;
--- a/packages/core/src/utils/helpers.css
+++ b/packages/core/src/utils/helpers.css
@@ -1,23 +1,13 @@
-.hiddenFromScreen {
+.visually-hidden {
   border: 0;
-  clip: "rect(0 0 0 0)";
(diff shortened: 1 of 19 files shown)


harriplappalainen (commented):
Looks like a good start! 👍 I think this is a good time to revisit and perhaps improve the cookie consent's CSS as well, for which we'll need to consult the cookie expert @NikoHelle. For example I see some element and descendant selectors, and can't help but wonder could they be improved upon? Here's a good article from CSS-tricks about [efficient CSS](https://css-tricks.com/efficiently-rendering-css/)

site/src/docs/components/cookie-consent/code.mdx
```diff
@@ -61,10 +63,107 @@ export const SimpleModalVersion = () => {
...
+          </div>
+        </form>
+      </div>
+    </div>
+  </div>
+```
+
+</Playground>
+
 <Link
```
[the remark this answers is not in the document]
Jussiles: I will add it to core storybook

packages/core/src/utils/helpers.css
```diff
@@ -9,6 +9,17 @@
...
+  clip: rect(1px, 1px, 1px, 1px);
+  clip-path: inset(50%);
+  max-height: 1px;
+  max-width: 1px;
+  opacity: 0;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+}
+
```
NikoHelle, line 22: There is .hiddenFromScreen css above this. I guess that does the same as this code, but in a slightly different way. Combine these two or delete this one.

Looks like this has more tricks than the old one.

Jussiles: Good points, I noticed it but didn't have courage to modify the old one. I will combine those and rename the old class to visually-hidden as I believe it's more commonly used. I try to make it without braking anything.

Jussiles: I renamed hiddenFromScreen also on react side and pushed changes. Let me know if it's good or not.

harriplappalainen: Actually renaming this is a breaking change requiring this to be pushed to major release, because this helpers.css is included in hds-core package. So maybe revert this for now and I can make a ticket about it 👍 A stable version 3 should be right around the corner and I think this core cookie consent will take quite some time to finish, so I don't think this will make it to 3.

harriplappalainen: Or we might make the change ourselves to v.3 💡 We'll see!

Jussiles: Good point! I will revert the changes from this PR

PR #1064, opened 2023-07-10
HDS-764 keyboard navigation

NikoHelle:
## Description

Generic keyboard navigation for vanilla js and React hook and component.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-764

Does not solve the issue, but creates components for it

Closes #

## How Has This Been Tested?

Local machine, Unit tests and Loki tests

Demo: 

https://city-of-helsinki.github.io/hds-demo/keyboard-navigation/?path=/story/components-keyboardnavigation--example

Docs:

https://city-of-helsinki.github.io/hds-demo/keyboard-navigation-docs/components/keyboard-navigation


harriplappalainen (commented):
I reviewed just the documentation side of this. I'll hold off on reviewing code because I have no background information about this implementation and I would rather discuss that when you're back from vacation. I'd like to know what this is for exactly. Without any background information I find that offering alternatives to native keyboard navigation sounds a bit problematic because how are end users going to know about these alternative navigation methods?

NikoHelle:
Changed to draft, because not needed at the moment.

PR #1080, opened 2023-07-28
Hds 1801 center header

harriplappalainen:
## Description

Header should be as wide as the browser is and center its contents. That way with custom background colour, the colour compasses the whole browser's width. This way the border-bottoms for ActionBar and NavigationMenu are as wide as the browser as well.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1801

## How Has This Been Tested?
Locally
[React storybook demo](https://city-of-helsinki.github.io/hds-demo/center-header)


minnas (approved):
Great work! 👍

PR #1081, opened 2023-07-31
Hds 1806 highlight Quote em-dash fix

minnas:
## Description

Quote's reference has `hyphen` which is incorrect. Should use `em-dash` instead.
React adds it automatically but core users need to insert the dash themselves. Added documentation about it.

**Bonus work:** 
- Rename prop `variant` to `size` as it's semantically correct and consistent with HDS style
- Add missing imports to Breadcrumb, Hero and Highlight docs code sections

## Motivation and Context

[HDS-1806](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1806)

## How Has This Been Tested?

Tested locally 

Quote-em-dash

## Demos

[Docsite](https://city-of-helsinki.github.io/hds-demo/dash-docs)
Core [Storybook](https://city-of-helsinki.github.io/hds-demo/quote-fix)
React [Storybook](https://city-of-helsinki.github.io/hds-demo/quote-fix-react)

[HDS-1806]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


harriplappalainen:
As I started to review this, I realised that the two css files is cumbersome for core users because consistently only one css file is needed (named after the component), so this change would have to be advertised in "kitty sized letters" as the saying goes in Finland. 

I tried all sorts of things but eventually the simplest solution is to have core users put the dash sign themselves and not provide it via CSS. To get it working in core and React, I presume the problematic CSS code (`content: '\2014';`) would have to formatted in a different way in the file `packages/react/lib/index.css-text.js`. I tried to search and replace it but couldn't get it working.

I updated [the ticket](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1842) for investigating this about how to start debugging this.

nooranen (approved):
Great job! 🥇 
The little dash caused so much headache, but the em-dash is looking beautiful. "Bull out of a fly" as the saying goes in Finland. 😇

laurakarhu (approved):
I want to highlight that this hightlight looks good :D

mrTuomoK:
I removed some duplicate css from react-side in the two commits, no need to add the same styles beneath the react-side classes since they are already picked up from the core css with compose.

PR #1083, opened 2023-08-03
Hds 701 landmarks for header and footer

harriplappalainen:
## Description

Users had expressed a need for adding their own landmarks. However the `<header>`, `nav` and `<footer>` elements provide landmarks automatically so I don't think we should add more in order to keep the page layout clear for screen readers [(more about the matter here)](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#3._landmark_roles).

- Add search landmark
- Add support for `role` and `ariaLabel` props to Header and Footer sub components
- Add information about the topic in Header and Footer documentation

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-701

## How Has This Been Tested?
Locally
[Storybook demo](https://city-of-helsinki.github.io/hds-demo/landmarks/?path=/story/components-footer--example)
[Site demo](https://city-of-helsinki.github.io/hds-demo/landmarks-docs)

### DIFF
--- a/packages/react/src/components/footer/Footer.stories.tsx
+++ b/packages/react/src/components/footer/Footer.stories.tsx
@@ -115,23 +115,6 @@ export default {
   },
 };
 
-export const Default = (args) => (
-  <Footer {...args}>
-    <Footer.Navigation>
-      {createArray(8).map((index) => (
-        <Footer.NavigationLink
-          key={index}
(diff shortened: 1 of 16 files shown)


mikkojamG (approved):
Wonderful, love how clearly using landmarks is explained 👍

PR #1086, opened 2023-08-07
HDS-1804 Move _typography.scss file

harriplappalainen:
## Description

- Move the `_typoraphy.scss` under `packages/react/src/components/header/components/headerActionBar/` folder as it's only used there for action bar title. The styles do not relate to heading styles in `packages/core/src/utils/helpers.css`.
- Use tokens instead of static values
- Remove support for `TitleStyleType.Black` because we aren't sure if it's useful

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1804

## Motivation and Context

- The file was under `spackages/react/src/styles` which gave the impression that they were generic and usable anywhere but that was not the case. 
- I didn't refactor them even though the mixins are a bit complicated because they're implementation is very easy (check `packages/react/src/components/header/components/headerActionBar/HeaderActionBar.module.scss` line 63).
- I didn't find a need to create generic style mixins for HDS React components because the styles in `packages/core/src/utils/helpers.css` don't change with viewport width apart from one class. Mixins would be useful if they did change. So seeing as the using the core heading styles is a simple matter of composing classes, I don't see the benefit for complicating it.

## How Has This Been Tested?
Locally, jest tests and visual tests

## To do
- [x] Check if heading font values can be changed to tokens. If they're not tokens, inform the designers


mrTuomoK (approved):
Oukidou 👍

laurakarhu (approved):
Looks good to me :)

PR #1088, opened 2023-08-14
HDS-1117-fix-button-styles (DRAFT)

mrTuomoK:
## Description

Unifies the styles for Button-component in Core and React.

## Related Issue

[HDS-1117](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1117)

## Motivation and Context

Although React composed Core styles, the styles were different, this PR removes the most complicated css-rules. Somehow the multiple :not, :first-of-type, :last-of-type don't seem to work correctly here despite testing them in a completely separate playground in which they do work 🤷. I removed all those pesky rules and discarded a lot of code (~70 lines) and with few changes it should now match the designs.

## How Has This Been Tested?
local machine
[Core storybook demo](https://city-of-helsinki.github.io/hds-demo/button-space-core/?path=/story/components-button--primary)
[React storybook demo](https://city-of-helsinki.github.io/hds-demo/button-space-react/?path=/story/components-button--primary)
[Button docs demo](https://city-of-helsinki.github.io/hds-demo/button-space-docs/components/buttons)

[HDS-1117]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Status
- As this affect all HDS components that have buttons, this needs further testing and should be released in a major release.


harriplappalainen (commented):
Great effort by going through the designs and CSS so extensively 🙇 Overall everything seems to be in order but by browsing the loki ref images, I found some things that need tweaking:
- [Accordion](http://localhost:6006/?path=/story/components-accordion--card-accordion)
  - [In the designs](https://www.figma.com/file/Kxwg3R0zNRHj55nQqFu6VS/HDS-UI-Kit-3.0?type=design&node-id=2907-144526&mode=design&t=uVDfeVa6OWNcehba-0) the opened accordion item's inner Close button is vertically aligned with the Accordion's close button
- [Pagination](http://localhost:6006/?path=/story/components-pagination--basic)
  -  Next button is too close to page numbers. This is especially visible in the [truncated story](https://city-of-helsinki.github.io/hds-demo/button-space-react/?path=/story/components-pagination--with-truncation) by hovering the Next button
  - In truncated story [ref image](https://github.com/City-of-Helsinki/helsinki-design-system/blob/44861f9fed9d008de75c0011f68f803c21d6bcd7/packages/react/.loki/reference/chrome_iphone7_Components_Pagination_With_truncation.png) you can see that Previous and Next buttons are now more in the center. This isn't in Figma but I don't know if this is an issue? Ask Noora
- [Navigation](http://localhost:6006/?path=/story/components-navigation--dropdown-links) (old component)
  - The dropdown button is now closer than it was before. This might actually be considered a breaking change as it would break UI tests for users 😅 I don't know if old nav designs are in Figma but here's the current [storybook](https://hds.hel.fi/storybook/react/?path=/story/components-navigation--dropdown-links) at least
  - This reminds me that we don't use HDS Button in Header 😮 I wonder if we should?
- External Show password button's text now goes on two lines instead of being a one liner. Check the [story](https://city-of-helsinki.github.io/hds-demo/button-space-react/?path=/story/components-passwordinput--with-external-show-password-button)

mrTuomoK:
External show password -but fixed. Was due to storybook css, not something with the Button itself.

mrTuomoK:
Fixed the bug in Pagination by enlarging the padding to make the Button fit again.

packages/react/src/components/accordion/__snapshots__/Accordion.test.tsx.snap
```diff
@@ -46,7 +46,7 @@ exports[`<Accordion /> spec renders the component 1`] = `
     </div>
     <div
       aria-labelledby="accordion-1-heading"
-      class="accordionContent contentWithCloseButton"
+      class="accordionContent"
```
NikoHelle, line 49: In `Accordion.tsx` and its css, the`contentWithCloseButton` was removed, `contentWithoutCloseButton` added and conditional `closeButton && styles.contentWithCloseButton` changed to `!closeButton && styles.contentWithoutCloseButton´. But the `contentWithCloseButton` does not appear in the snapshot.

So it is not tested. I'd add test for it.

packages/react/src/internal/menuButton/MenuButton.module.scss
```diff
@@ -14,6 +14,7 @@
   --menu-item-color-highlighted: var(--color-white);
 
   border-radius: 0;
+  gap: var(--spacing-2-xs);
   position: relative;
```
NikoHelle, line 18: In Chrome: "The display: block property prevents gap from having an effect.
Try setting display to something other than block."

packages/core/src/components/button/button.css
```diff
@@ -176,65 +176,25 @@
...
-  padding: 0;
-}
-
-input[type="submit"].hds-button--small {
-  padding: var(--spacing-2-xs) var(--spacing-m);
-}
-
-/* both icons */
-.hds-button--small .hds-button__label:not(:first-of-type):not(:last-of-type) {
-  padding: var(--spacing-2-xs) var(--spacing-2-xs);
```
NikoHelle, line 237: I a bit concerned there has never been stories for all of these odd cases and therefore loki tests won't cover them and their removal 🤔

mrTuomoK: These were replace by the `gap` making the css a lot more simpler

packages/react/src/internal/menuButton/MenuButton.module.scss
mrTuomoK: Need to have a look at this

packages/react/src/components/accordion/__snapshots__/Accordion.test.tsx.snap
mrTuomoK: Yeah, will do. Good spot.

PR #1094, opened 2023-08-28
HDS-1861 Fix doc site table caption

harriplappalainen:
## Description

Table captions were displayed wrong under the table and with no styles. This was fixed by implementing the old table caption logic in doc site component Table.js and making it work with the new mdx table object.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1861

## How Has This Been Tested?
Local host
[Demo](https://city-of-helsinki.github.io/hds-demo/docs-table-caption)

### DIFF
--- a/site/src/components/Table.js
+++ b/site/src/components/Table.js
@@ -12,8 +12,9 @@ const resolveCaptionStrings = (captionString) => {
 };
 
 const Table = (props) => {
-  const thead = props.children[0].type.name === 'TableHead' ? props.children[0] : [];
-  const tbody = props.children[1].type.name === 'TableBody' ? props.children[1] : []
+  // TableHead and body can't be identified by type.name because it doesn't exist in prod build.
+  const thead = props.children[0];
+  const tbody = props.children[1];
   const tbodyRows = tbody.props?.children || [];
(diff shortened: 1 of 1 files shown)


site/src/components/Table.js
```diff
@@ -1,7 +1,56 @@
...
+  const tbody = props.children[1].type.name === 'TableBody' ? props.children[1] : []
+  const tbodyRows = tbody.props?.children || [];
+  const rowsArray = Array.isArray(tbodyRows) ? tbodyRows : [tbodyRows];
+
+  const [rows, captionString] = rowsArray
+    .filter((child) => isValidElement(child))
+    .reduce(
+      (acc, row, i, arr) => {
+        if (
+          i >= arr.length - 1 &&
```
NikoHelle, line 25: `i` can never be > `arr.length -1 `. Small thing, but the whole `if` is quite complicated, so ">=" made me wonder even more what it does.

I don't like using optional chaining (or is it even possible in this environment), but it would make this more readable.

`if ( i=== arr.length - 1 && isCaption(row.props.children?.[0].props.children))`

harriplappalainen: Thanks! Fixed.

timwessman:
Works nicely!

timwessman (approved):

PR #1106, opened 2023-09-04
Hds 1850 subnav chevron fix

harriplappalainen:
## Description

Header's nested dropdowns chevron should point to the direction where the dropdown is opening. Original work àla Minna. Modified code to work with existing logic regarding the dropdown position and change the nested menu icon to point to the opposite direction when opened.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1850

Screenshot 2023-09-04 at 8 54 15

**To do**
- [x] When nested dropdown is open, the arrow should point to the opposite direction, not down
- [x] Make a fail safe for SSR cases, i.e. menu is always on right because we can't touch window object
- [x] Test on SSR (site demo counts?)

[Storybook demo](https://city-of-helsinki.github.io/hds-demo/chevron-fix-react/?path=/story/components-header--with-full-features)
[Site demo](https://city-of-helsinki.github.io/hds-demo/chevron-fix/components/header/)

### DIFF
--- a/packages/react/src/components/header/components/headerLink/HeaderLink.tsx
+++ b/packages/react/src/components/header/components/headerLink/HeaderLink.tsx
@@ -1,4 +1,12 @@
-import React, { MouseEventHandler, cloneElement, useCallback, useEffect, useRef, useState } from 'react';
+import React, {
+  MouseEventHandler,
+  cloneElement,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'react';
 import { v4 as uuidv4 } from 'uuid';
 
 // import base styles
@@ -159,6 +167,13 @@ export const HeaderLink = <T extends React.ElementType = 'a'>({
   const containerRef = useRef<HTMLSpanElement>(null);
   const isSubNavLink = openSubNavIndex !== undefined && setOpenSubNavIndex !== undefined;
 
+  const [menuSize, setMenuSize] = useState<DOMRect>(undefined);
+
+  useLayoutEffect(() => {
+    const size = containerRef?.current?.getBoundingClientRect();
+    setMenuSize(size);
+  }, []);
+
   const handleDynamicMenuPosition = (val: boolean) => {
     // Null position when false so if user resizes browser, the calculation is done again
     if (!val) setDynamicPosition(null);
@@ -175,6 +190,19 @@ export const HeaderLink = <T extends React.ElementType = 'a'>({
         setDynamicPosition(position);
       }
     }
+  };
+
+  const getPosition = () => {
+    // eslint-disable-next-line no-lonely-if
+    if (menuSize) {
+      // Calculate which side has more space for the menu
+      const { x: leftPosition, width } = menuSize;
+      const rightPosition = leftPosition + width;
+      const position1 =
+        leftPosition > window.innerWidth - rightPosition ? DropdownMenuPosition.Left : DropdownMenuPosition.Right;
+      return position1;
+    }
+    return undefined;
   };
 
   // Handle dropdown open state by calling either internal state or context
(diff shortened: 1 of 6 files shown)


mikkojamG (approved):
LGTM 👍

packages/react/src/components/header/components/headerLink/HeaderLink.tsx
```diff
@@ -159,28 +160,26 @@ export const HeaderLink = <T extends React.ElementType = 'a'>({
...
-      if (containerRef.current) {
-        // Calculate which side has more space for the menu
-        const { x: leftPosition, width } = containerRef.current.getBoundingClientRect();
-        const rightPosition = leftPosition + width;
-        const position =
-          leftPosition > window.innerWidth - rightPosition ? DropdownMenuPosition.Left : DropdownMenuPosition.Right;
-        setDynamicPosition(position);
-      }
+  const getDropdownPosition = () => {
+    let position;
```
NikoHelle, line 164: This `let` could be discarded if this function returns the value immediately when resolved.

```
const getDropdownPosition = () => {
    let position;
    // On SSR just set the menu on the right
    if (window === undefined) {
      return DropdownMenuPosition.Right;
    }
    if (containerRef.current != null) {
      const { x: leftPosition, width } = containerRef.current.getBoundingClientRect();
      const rightPosition = leftPosition + width;
      return leftPosition > window.innerWidth - rightPosition ? DropdownMenuPosition.Left : DropdownMenuPosition.Right;
    }
    return position;
  };
```

Btw, it returns undefined, if window is not undefined and containerRef.current is null ? The function or `position`  are not typed, so can't tell is that correct. There should be a ts error setting `undefined` to the `setDynamicPosition`... can't say why it accepts it.

packages/react/src/components/header/components/headerLink/headerLinkDropdown/HeaderLinkDropdown.tsx
```diff
@@ -90,6 +90,14 @@ export const HeaderLinkDropdown = ({
 
   const childElements = getChildElementsEvenIfContainersInbetween(children);
 
+  const renderIcon = () => {
+    if (depth > 1 && dynamicPosition === DropdownMenuPosition.Right)
```
NikoHelle, line 94: This has not changed in this PR, but it is a bit weird to call this prop 'dynamicPosition', because it is not dynamic at all from the components point of view :) It's just a prop.

NikoHelle:
Good, simple solution! 👍 Some minor comments added.

Not changed in this PR, but I'd change odd `calc(100%)` in HeaderLinkDropdown.module.scss.

packages/react/src/components/header/components/headerLink/HeaderLink.tsx
```diff
@@ -159,28 +160,26 @@ export const HeaderLink = <T extends React.ElementType = 'a'>({
...
+    // On SSR just set the menu on the right
+    if (window === undefined) position = DropdownMenuPosition.Right;
```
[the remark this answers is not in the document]
harriplappalainen: `window` is undefined in SSR 👍 Good looking out.

packages/react/src/components/header/components/headerLink/HeaderLink.tsx
harriplappalainen: I think it's safe to have a default `return DropdownMenuPosition.Right;` at the end.

packages/react/src/components/header/components/headerLink/headerLinkDropdown/HeaderLinkDropdown.tsx
harriplappalainen: Well, I guess.. 😄 I can rename it just to be `position`.

NikoHelle (approved):

PR #1111, opened 2023-09-05
Update RadioButton guideline to have no default value.

arpargo:
## Description

Fix for updating the guideline to reflect component suggestions.

## Related Issue

[HDS-1859](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1859)

Closes #

## Motivation and Context

Fix for updating the guideline to reflect component suggestions.

## How Has This Been Tested?

I've compared the site locally with the production site and confirmed that the radiobutton does not have default value on local instance.

[HDS-1859]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/site/src/docs/foundation/guidelines/checkbox-radiobutton-toggle.mdx
+++ b/site/src/docs/foundation/guidelines/checkbox-radiobutton-toggle.mdx
@@ -35,7 +35,7 @@ Radio buttons have a lower cognitive load than dropdown menus because they make 
 **Example use:** Binary choices like "Yes" or "No", age group or other similar choices that exclude other options.
 
 export const RadioButtonExample = () => {
-  const [selectedItem, setSelectedItem] = React.useState("1");
+  const [selectedItem, setSelectedItem] = React.useState('');
   const onChange = (event) => {
     setSelectedItem(event.target.value);
   };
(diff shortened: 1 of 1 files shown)


mikkojamG (approved):

NikoHelle:
Does the bolded text on the same page conflict with the new "no default" standard?

"HDS strongly recommends that radio buttons are never presented unticked."

https://hds.hel.fi/foundation/guidelines/checkbox-radiobutton-toggle/

And there is also "If none of the buttons is selected by default, users have no way to revert to this original state after they’ve made a selection. The lack of a standard can be confusing to users."

But I don't understand what that is really trying to tell 😄

arpargo:
> Does the bolded text on the same page conflict with the new "no default" standard?
> 
> "HDS strongly recommends that radio buttons are never presented unticked."
> 
> https://hds.hel.fi/foundation/guidelines/checkbox-radiobutton-toggle/
> 
> And there is also "If none of the buttons is selected by default, users have no way to revert to this original state after they’ve made a selection. The lack of a standard can be confusing to users."
> 
> But I don't understand what that is really trying to tell 😄

That is a good catch, I didn't even think of double checking the texts when making this change. This indeed is conflicting with the new way of doing this so we should have the text re-thought and replaced. What the initial stage is worried is that unlike in the case of an checkbox, you can always untick a selection. If you present radiobuttons without false value, you're forced to make a choice if you happen to select one option, even if the field is presented as optional.

NikoHelle:
Oh, it is trying to tell that there should be an option like "None of the above" in the radio button group?

NikoHelle (approved):

laurakarhu (approved):
Thanks, now it's according to the new standard and it's not conflicting anymore.

PR #1117, opened 2023-09-14
HDS-1906 fix language selector re-rendering and prop passing

NikoHelle:
## Description

The LanguageSelector stored passed languages and children to context and then another instances used those to actually render language options.

The LanguageSelector is now picked from children of the ActionBar and its props and children are extracted and passed to new components which render the content.

Less re-rendering and props are now also actually passed.

A list of language options are now passed in Header props with other language related props. No need to update context anymore from LanguageSelector.

No demo or docs, because nothing has changes except under the hood and visual tests pass.

## Related Issue

Closes  [HDS-1906](https://helsinkisolutionoffice.atlassian.net/browse/HHDS-1906)

## How Has This Been Tested?

Locally. Added some jest test and checked visual tests pass.

## Screenshots (if appropriate):

[HDS-1906]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1118, opened 2023-09-15
HDS-1795 split language options to primary and secondary

NikoHelle:
**Note: this is not targeted to the `release-3.0.0` branch, because this depends on another branch which is targeted to the release 3.  The target branch has its[ own PR](https://github.com/City-of-Helsinki/helsinki-design-system/pull/1117).**

## Description

Language options can be rendered in the `ActionBar` or inside a menu. Visible items are named primary and those are "fi" , "en" and "sv" and other options, named secondary, are placed in the menu.

If a secondary option is selected from the menu, then it should be the only one visible in the `ActionBar` and all other ones are moved in the menu.

This is done with a new option sorting function "sortLanguageOptions". It sorts the language options according to [designs](https://www.figma.com/file/Kxwg3R0zNRHj55nQqFu6VS/branch/3dKCA9y0tz2aVVFXuw1d8h/HDS-UI-Kit-3.0?type=design&node-id=3967-118903&mode=design&t=Qdo9xtDsu0LKF3z7-0).

Users can implement their own sorting function if necessary. 

If none of the language options are marked primary with the `isPrimary`-prop then all are rendered as before and secondary options do not exist.

[Demo](https://city-of-helsinki.github.io/hds-demo/HDS-1795-languages/?path=/story/components-header--minimal-with-localization)

This PR does not stylize the menu.

## Related Issue

Closes https://helsinkisolutionoffice.atlassian.net/browse/HDS-1795

## How Has This Been Tested?

Unit and loki tests.


packages/react/src/components/header/components/headerLanguageSelector/HeaderLanguageSelector.tsx
```diff
@@ -38,6 +48,17 @@ const renderLanguageNode = (language: LanguageOption) => {
   return <LanguageButton key={language.value} value={language.value} label={language.label} />;
 };
 
+const defaultLanguageSorter: LanguageSelectorProps['sortLanguageOptions'] = (options, selectedLanguage) => {
+  const hasPrimaryLanguages = options.findIndex((option) => typeof option.isPrimary !== 'undefined') > -1;
```
mikkojamG, line 52: Suggestion:
`const hasPrimaryLanguages = options.some((option) => typeof option.isPrimary !== 'undefined');`

mikkojamG (commented):
Me likes. Made suggestion on one line.

packages/react/src/components/header/components/headerLanguageSelector/HeaderLanguageSelector.tsx
mrTuomoK: This is good 🤌

mikkojamG (approved):

laurakarhu (approved):
The spanish language could be real there in the example, but at least it does show the text changes and that is more important for the feature.

NikoHelle:
> The spanish language could be real there in the example, but at least it does show the text changes and that is more important for the feature.
> 
> Actually the navigation (Sosiaali-ja terveyspalvelut, Kasvatus ja koulutus, Asuminen) does not react to the language change.

This feature is developed in another branch atm.

mikkojamG (commented):
Found a bug with Minimal with localization story, where selecting one of the other languages from dropdown causes an error. 🤔 

Screenshot 2023-10-02 at 10 40 16

NikoHelle:
> Found a bug with Minimal With Localization story, where selecting one of the other languages from dropdown causes an error. 🤔
> 
> Screenshot 2023-10-02 at 10 40 16

We don't have texts yet for de, fr, es and ru. There is a ticket HDS-1933 for those. I can add fake translations here.

NikoHelle:
Added placeholder translations

mikkojamG (approved):

PR #1119, opened 2023-09-18
Fix DatePicker issue: Selected month is reseted when parent component state changes

jorilindell:
## Description
Memoize `minDateToUse`, `maxDateToUse`, `selected`  ja `initialMonth` props before passing them to DatePicker component.

Link to discussion: https://helsinkicity.slack.com/archives/CHCV3KTHA/p1695025333999949

## Motivation and Context
In Linked Registration UI project we noticed that selected month and year in DatePicked was reseted always when the authentication session status was updated.  The reason seams to be `minDateToUse`, `maxDateToUse`, `selected`  ja `initialMonth` props for DatePicker component which are re-created in every render round.

Quite simple way to reproduce this problem is:
```
export const Default = (args) => {
  const [count, setCount] = useState(1);

  useEffect(() => {
    const interval = setInterval(() => setCount((c) => c + 1), 3000);

    return () => clearInterval(interval);
  }, [count]);

  return <DateInput {...args} />;
};
```

## How Has This Been Tested?
This has been tested in Storybook by using code snippet above.

### DIFF
--- a/packages/react/src/components/dateInput/DateInput.tsx
+++ b/packages/react/src/components/dateInput/DateInput.tsx
@@ -1,5 +1,5 @@
 import { format, parse, isValid, subYears, addYears, startOfMonth, endOfMonth, max } from 'date-fns';
-import React, { useState, useRef, useEffect, useCallback } from 'react';
+import React, { useState, useRef, useEffect, useCallback, useMemo } from 'react';
 
 import { IconCalendar } from '../../icons';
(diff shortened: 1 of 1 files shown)


mikkojamG (approved):
LGTM 👍 I would still get another approval just to make sure 🙂

laurakarhu (approved):
Thank you for the contribution! Looks clear to me.

harriplappalainen (approved):
Good find, good fix 👍

PR #1129, opened 2023-10-03
Hds 1948 translate universal links in mobile menu

harriplappalainen:
## Description

When changing languages, the universal links are not translated in mobile menu. This was because the universalContent property in HeaderContext was not set from UniversalBar because the component had too early of a `return null;`

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1948

## How Has This Been Tested?
Locally with a dummy test case for switching languages
[Demo](https://city-of-helsinki.github.io/hds-demo/hds-1948-universal-links-in-mobile/?path=/story/components-header--with-full-features)


mikkojamG (approved):

laurakarhu (approved):
Thanks!

PR #1145, opened 2023-10-09
(HDS-1763) fix tag border-radiuses and paddings to match

mrTuomoK:
Fix tag border-radiuses and paddings to match designs and bug in multilines, update reference images.

## Description

There was a slight difference in Tag radiuses compared to the designs and the implementation couldn't also handle Tags with long texts that go multi-line. Also fixes other styling issues and issue with React Tag large size demo in storybook.

## Related Issue

[HDS-1763](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1763)

## How Has This Been Tested?

Local machine

From: 
image

To:
image

[HDS-1763]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrTuomoK:
The tag radiuses are now calculated and match the designs (32px high tag -> / 2 = 16px and Large size -> 48px high Large / 2 -> 24px radius) The old implementation was hard-coded 20px for both sizes 🤦

mrTuomoK:
I'll revise this a bit according to discussions with designers

packages/core/src/components/tag/tag.stories.js
```diff
@@ -33,3 +39,27 @@ export const LargeDeletable = () => `
...
+export const LongTextDeletable = () => `
+  <div class="hds-tag" style="max-width: 300px;">
+  <span class="hds-tag__label" aria-hidden="true">Label - This is a tag with a very long text which is not advisable and might span into multiple lines</span>
+    <button aria-label="Delete item: Label" class="hds-tag__delete-button button-reset">
+      <span aria-hidden="true" class="hds-icon hds-icon--cross"></span>
+    </button>
+  </div>
+`;
+
+export const LargeLongTextDeletable = () => `
```
harriplappalainen, line 58: Core is missing LargeLongText story that React storybook has. Also the stories in React are in a better order as they are sort of grouped by the context, e.g. Large Tag stories are grouped, after that Long Text Tags (default). Could you put the stories in the same order in Core side as well?

packages/core/src/components/tag/tag.css
```diff
@@ -5,27 +5,32 @@
...
+  padding: var(--tag-padding)
 }
 
 .hds-tag--large {
   --tag-font-size:  var(--fontsize-body-m);
-  --tag-spacing:    var(--spacing-s);
-  --tag-height:     48px;
+  --tag-spacing: var(--spacing-s);
+  --tag-height: 48px;
+  --tag-padding: 8px 20px;
```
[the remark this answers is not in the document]
harriplappalainen: Also, why not slap this into a variable `--tag-padding-large` since there is a `--tag-padding` already, eh?

harriplappalainen (commented):
I see you have gone into the end zone again, or "päätyyn asti" as it's said in Finnish. Good fixings 🥇 I had just a couple of minor comments.

packages/core/src/components/tag/tag.stories.js
mrTuomoK: Will do

packages/core/src/components/tag/tag.css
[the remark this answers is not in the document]
mrTuomoK: Good points, I'll also again double-check the paddings 👍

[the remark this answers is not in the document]
mrTuomoK: The paddings should now be as correct as they can be since Figma has it's own quirks.

packages/core/src/components/tag/tag.stories.js
mrTuomoK: Done

harriplappalainen (approved):
It is done! 🙌

packages/core/src/components/tag/tag.css
```diff
@@ -4,29 +4,36 @@
...
+  box-sizing: border-box;
   color: var(--tag-color);
   display: inline-flex;
-  flex-direction:   row;
+  flex-direction: row;
   font-size: var(--tag-font-size);
-  line-height: var(--tag-height);
+  gap: var(--spacing-2-xs);
+  justify-content: center;
+  line-height: 24px;
```
SillanmakiA, line 26: Figmassa käytetään tarkkoja pikseleitä. Pitäisikö koodissa kuitenkin käyttää tokeneita? (Pohdinnassa vaihtaa Figmaan myös tokenien mukaiset prosentuaalit käyttöön)

mrTuomoK: changed to use line-height -vars

NikoHelle (approved):

PR #1147, opened 2023-10-11
HDS-1945 Fix input buttons focus

harriplappalainen:
## Description

None the buttons inside React input components can be focused. This was caused by using a wrong variable name. Core has no issues.

Affected components that underwent changes in this PR:
- CookieConsent
- Dropdown
- FileInput
- NumberInput
- SearchInput
- TextInput
- (internal) SelectedItems

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1945

## How Has This Been Tested?
Localhost
[Storybook demo](https://city-of-helsinki.github.io/hds-demo/fix-input-button-focus/)

To do
- [x] Fix cookie consent accordion show/hide button focus

### DIFF
--- a/packages/react/src/components/cookieConsent/CookieConsent.module.scss
+++ b/packages/react/src/components/cookieConsent/CookieConsent.module.scss
@@ -8,7 +8,7 @@
   z-index: 999;
 
   --common-spacing: var(--spacing-s);
-  --focus-outline-color: var(--color-coat-of-arms);
+  --focus-outline-color: var(--color-focus-outline);
   --outline-width: 3px;
 }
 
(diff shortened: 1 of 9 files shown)


laurakarhu (approved):
Good catch, thanks! And I'm glad there was a common easy way to fix problem.

mrTuomoK (approved):
Good easy fix :) btw if we are really really strict to use british-english everywhere, our `color` should be `colour` (Dunno if I nagged about this way back :D )

harriplappalainen:
> Good easy fix :) btw if we are really really strict to use british-english everywhere, our color should be colour (Dunno if I nagged about this way back :D )

I noticed this horrible disparity with this PR as well, that the doc site has `colour` but design tokens have `color` 😁 There was some discussion about this before. Probably users don't notice this though.

mrTuomoK:
These carets (or what are they called?), can't be focused anywhere in CookieConsent at least 
image

mrTuomoK (approved):
Couldn't spot any more bugs 👍

PR #1149, opened 2023-10-12
HDS-1909 bug header dropdown position

mikkojamG:
## Description

Fixing header dropdown overflow issues, trying to make it look better on smaller screen and when there are navi items on multiple rows

## Related Issue

Closes [HDS-1909](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1909)

## How Has This Been Tested?

[React demo](https://city-of-helsinki.github.io/hds-demo/hds-1909-react/?path=/story/components-header--with-full-features)

hds-1909-fixed

[HDS-1909]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


laurakarhu (approved):
Much better, thanks!

harriplappalainen (approved):
Good fixes 💐 

There was a [change](https://github.com/City-of-Helsinki/helsinki-design-system/pull/1149/commits/3aefc0a172e4b973c68fd9b7917605aa527e9a2f) to use `window.outerWidth` instead of `window.innerWidth` but innerWidth is better to use because it gets the width of the container DOM such as iframe, which is where stories are embedded in Storybook. `outerWidth` gets the whole browser's width. So when using `outerWidth`, 
in Storybook it looks like this fix was not implemented because the code thinks there is more space than there actually is.

PR #1152, opened 2023-10-20
Hds 1969 header fixes

harriplappalainen:
## Description

- In doc site the mobile navigation menu is broken. Doesn't show at all in mobile view and in desktop view the styles from underneath the page are leaking through
- The Header's mobile navigation menu doesn't cover the viewport height as it should. Check [hel.fi](hel.fi) implementation. This is now fixed, **but there is still some empty space at the bottom. Can anybody help pinpoint the problem and fix?**

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1969

## Motivation and Context

Navigation should be accessible in mobile doc site view and styles should work. The Header's mobile navigation menu should encompass the whole screen height.

## How Has This Been Tested?
[Doc site](https://city-of-helsinki.github.io/hds-demo/hds-1969-fix-header-site)
[React storybook](https://city-of-helsinki.github.io/hds-demo/hds-1969-fix-header-react)


laurakarhu:
> but there is still some empty space at the bottom. Can anybody help pinpoint the problem and fix? 
Seems to be something Chrome related, looks good on Edge and Firefox.

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
```diff
@@ -287,9 +303,15 @@ export const HeaderActionBarNavigationMenu = ({
...
+      // Set the focus to the currently active page link
       const linkElement = await getIsElementLoaded(`#${currentActiveLinkId}`);
-
       linkElement.focus();
+
+      // Set the height of the menu container
+      const renderedChildIndex = Math.abs(navContainerRef.current.getBoundingClientRect().left / window.innerWidth);
+      const currentTargetHeight = targetElement.children[renderedChildIndex].clientHeight;
+      navContainerRef.current.style.height = `${currentTargetHeight}px`;
+      navContainerRef.current.style.overflow = 'hidden';
```
NikoHelle, line 314: First I was thinking this should be done with just css, then I tried to do it in the StoryBook demo.  Removing the `height` and `overflow` added by js from the inspector affects nothing.  The height stays the same without styles set by js.

So maybe this code is redundant due some  other css changes?

harriplappalainen: When I took the style setting code block out, the empty space is back after the logo. It can be seen from the demo by going to Sosiaali- ja terveyspalvelut -> Terveydenhoito. But I do agree that pure CSS solution would be the best, I'm just not sure how to do that.

harriplappalainen: > It can be seen from the Full features story by going to Sosiaali- ja terveyspalvelut -> Terveydenhoito. 

This can be seen from the demo as well if you take the element styles away in the inspector.

harriplappalainen:
To do:
- [ ] Fix scrollbar reserved space (?) on 320px wide screens https://city-of-helsinki.github.io/hds-demo/hds-1969-fix-header-react/iframe.html?id=components-header--with-full-features&viewMode=story

harriplappalainen:
To do
- [ ] In doc site the scrollbar is usable in mobile menu. It's scrolling the page content
- [ ] In code examples the mobile menu opens partly visible at the end of the page in fixed position

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
NikoHelle: I'm still struggling to see how this has any effect 😄 

If I comment the code out, the logo always sticks at the bottom anyway. This happens because each navSection fills the viewport vertically anyway and the logo is always at the bottom. 

If I keep the code and remove the "height" (that was set by js) from the inspector, everything is still the same.

Tested with FF & Chrome & Safari on different browser heights and clicked the mobile menus back and forth. The view is unbreakable. 😄 

PS. the `overflow = 'hidden';` could be in the css rather that always resetting it?

NikoHelle (changes requested):

packages/react/src/components/header/components/headerActionBar/HeaderActionBar.module.scss
```diff
@@ -32,6 +33,13 @@
   width: 100%;
 }
 
+.mobileMenuContainer {
+  height: 100%;
+  overflow-x: hidden;
+  overflow-y: auto;
+  position: fixed;
+}
```
NikoHelle, line 41: There is potential problem caused by this. Height of the whole `<header>` is 0px in the dom. Some elements below the header can be positioned incorrectly. The header should at least have the height of the actionBar.

The fixed element should be a descendant, maybe the navigationMenu, which is already absolutely positioned.

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
harriplappalainen: I don't want you to struggle, here's a video to explain why this code block is needed. Or at least I couldn't get it done with CSS alone.
Here's a video what happens when code is not there, i.e. I removed the code block. Notice a varying amount of empty space on the bottom after the logo. This happens if one of the submenus is very long.

https://github.com/City-of-Helsinki/helsinki-design-system/assets/108321134/186dda41-b421-41fd-b1cf-69dd6a61f9c4

And here's a video with the code block in; no extra space after logo

https://github.com/City-of-Helsinki/helsinki-design-system/assets/108321134/078d4dc6-7361-4ed8-b694-30a606ebfb5e

PS. I think you're right that the overflow hidden could be in the CSS.

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.module.scss
```diff
@@ -42,11 +42,9 @@
 }
 
 .headerNavigationMenu {
-  background-color: var(--nav-mobile-menu-bottom-background-color);
-  max-height: 0;
-  overflow: hidden;
+  min-height: 0;
```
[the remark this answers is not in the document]
harriplappalainen: Since you questioned it, I removed it to see what happens and - nothing! Everything's ok. Good find!

packages/react/src/components/header/components/headerActionBar/HeaderActionBar.module.scss
harriplappalainen: Ah, that is true. I set the position: fixed so high in DOM tree because that way the whole actionbar and mobile menu cover the whole page. You can see this in doc site demo and remove the position fixed rule, then the whole page comes into view by scrolling after the menu. But even with the position fixed, the scroll bar appears in doc site (when the page is long), and the parent height: 0, this is not ideal... Sigh. Back to the drawing board.

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
NikoHelle: Thx, my screen was too high to see that! It does not show, if the height is larger than highest menu I guess.

NikoHelle: `.navSection` has `min-height: calc(100vh - var(--action-bar-container-height));`

So it should always fill the whole view.

But it also has `height: min-content;` which shrinks it back to the content minimum.

So if js code block is removed **and** also the `height: min-content;  then it seems to work!

https://github.com/City-of-Helsinki/helsinki-design-system/assets/2639230/464096c3-e36e-4f9b-be97-48e0be6dd05a

The multilevel menu system is quite complex, so maybe I'm missing some scenario where the logo is positioned incorrectly. I'm just sceptical, because there is rarely a need to use js to make an element fill the viewport.  🤔

harriplappalainen: Thanks for helping! But in your video the last menu has way too much empty grey space between universal links and logo. If you compare that menu and the first level menus as they have almost the same amount of links, the last menu is much longer than it should be Maybe the long middle menu is affecting things still.

NikoHelle: 🎶  "I can see clearly now" 🎶 

All menu levels are side by side and the hidden ones take space in the dom. Therefore their container is as high as the highest menu. Which causes all menus to fill the height of the container and the logo jumps at the bottom which is sometimes out of the screen. 

JS sets the height of the menu to fill just the its own size, because otherwise it is not possible to align the logo at the bottom of the container or bottom of the screen - which ever is higher.

In current dom, the js version is the only option 👍

NikoHelle: But.... 😄 If a new container is added, the logo is correctly positioned! In my opinion, that is.

Only bad side-effect is, there is always a scrollbar, if one of the menus is higher than the whole page contents. That probably only happens in Storybook where there is no actual content.

https://github.com/City-of-Helsinki/helsinki-design-system/assets/2639230/37952c16-416c-4798-8b41-b85250c824b8


https://github.com/City-of-Helsinki/helsinki-design-system/assets/2639230/f8aa34e8-74ae-4542-9045-6f1b13c42e59

CSS changes:
image

Dom changes (added just the wrapper)

image

JS was removed too.

The reason this works is that the jew wrapper restricts the "flex-grow:1" in MobileMenuBottom to fill the container of all the menus.

harriplappalainen: Hmm, in that version the last menu has a lot of extra after the logo.. 😆 Might be again because the last menu was so high and that sets the parent containers' height.

packages/react/src/components/header/components/headerActionBar/HeaderActionBar.module.scss
harriplappalainen: I kept the `position: fixed` because it covers the whole page so well and just set min-height for the Header so it takes up the space of the action bar.

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
NikoHelle: Ok 😄  Let's keep the js

NikoHelle (approved):

NikoHelle:
Excellent improvements!

laurakarhu (approved):
Looks good to me, thanks!

PR #1153, opened 2023-10-23
Button.css fixes

Arkkimaagi:
## Description

* Fix syntax problems from button.css
    * ~`::` - notation for pseudo-elements like `::after`~
        * Build test linter was being nasty about this, so, I reverted it. I still think `::after` notation is better, as the last browser that required single-colon-notation was ie8 (R.I.P. 💀) as it did not support the double-colon version. As far as I know, ie8 is not on the HDS supported browsers list nor has it updated within last 6 months (try 14 years). So I'd recommend moving with the ages towards at least allowing the correct notation for pseudo elements.
    * Added newline after selector commas for readability and uniformity with the rest of the code. (I'd add this rule to the linter for future)
* `focus-visible` and `active` had duplicated property-sets.. ..that were simplified in this PR.
* Add better support for buttons of `type="reset"` and `type="button"` as only `type="submit"` was fully supported.
* Avoid duplicate selectors and overriding properties unnecessarily

## Related Issue

Does not close an issue, just fixed things that seemed broken while working with our end of the code.

## Motivation and Context

We're building on top of HDS button styles at hel.fi and having a cleaner and more readable source makes it easier to follow.

There's still plenty wrong on the button.css, but these were the easy to fix issues that came up. 

* We'd really recommend considering to move towards removing requirement for the `<span class="hds-button__label">...</span>` child element.
* We'd recommend strongly getting rid of the multiple margins and paddings inside buttons. 
* The logic of the button inner spacing would benefit of a major redesign towards more uniform and simpler style.
    * Why is the label closer to the edge on oppiside side of the button only when the button has an icon?
    * Why are the supplementary styles so different from other button styles? 
image 
image
    * Why are the button icons so far from the button labels?

We can help with these changes if you wish to move towards a more sane button style as we have a working alternative with matching styles to the current design.

## How Has This Been Tested?

This change is not really tested by us, as we indeed have our own version we're using. I hope the changes speak for themself and to me it looks like this should not be a breaking change.

None related to the changes in this PR, a couple above related to point out the issues with current design (that this PR should not change in any way)

### DIFF
--- a/packages/core/src/components/button/button.css
+++ b/packages/core/src/components/button/button.css
@@ -163,19 +163,25 @@
 }
 
 input[type="submit"].hds-button,
+input[type="reset"].hds-button,
+input[type="button"].hds-button,
 .hds-button__label {
   font-weight: inherit;
   line-height: 1.25em;
   padding: var(--spacing-s);
 }
 
-input[type="submit"].hds-button {
+input[type="submit"].hds-button,
+input[type="reset"].hds-button,
+input[type="button"].hds-button {
   cursor: pointer;
   padding: var(--spacing-s) var(--spacing-l);
 }
 
-/* submit input */
-input[type="submit"].hds-button:focus-visible {
+/* submit/reset/button input */
+input[type="submit"].hds-button:focus-visible,
+input[type="reset"].hds-button:focus-visible,
+input[type="button"].hds-button:focus-visible {
   box-shadow: 0 0 0 var(--outline-gutter) var(--submit-input-focus-gutter-color), 0 0 0 calc(var(--outline-gutter) + var(--outline-width)) var(--focus-outline-color);
 }
 
@@ -185,6 +191,8 @@ input[type="submit"].hds-button:focus-visible {
 }
 
 input[type="submit"].hds-button--small,
+input[type="reset"].hds-button--small,
+input[type="button"].hds-button--small,
 .hds-button--small .hds-button__label {
   line-height: var(--lineheight-s);
   padding: var(--spacing-2-xs) var(--spacing-xs);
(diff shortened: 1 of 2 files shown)


mrTuomoK:
Just for the future, could the branch name be a bit more descriptive (starting with HDS-) and even better, could there be a ticket related to these?

Arkkimaagi:
Github auto-generates the branch name when I edit the contents online. The PR name "Button.css fixes" part is on me, I should have edited the PR name to be better from what github suggested.

Also, I did not have a ticket, only discussion in slack about providing findings about the stuff we went through when applying the new button code on our end. In the future I can contact HDS before making code suggestions to get a proper ticket number. I was just trying to write these things down from my memory from working with the code before I forgot.

NikoHelle:
Thx! These changes look good and visual tests pass. But our visual tests cannot test button pointer events, so I've checked those manually. I can't se any differences, so I'd say this PR is good to go.

NikoHelle (approved):

packages/core/src/components/button/button.css
```diff
@@ -162,32 +155,33 @@
...
 .hds-button__label {
   font-weight: inherit;
   line-height: 1.25em;
   padding: var(--spacing-s);
 }
 
-input[type="submit"].hds-button {
+input[type="submit"].hds-button,
+input[type="reset"].hds-button,
+input[type="button"].hds-button {
```
harriplappalainen, line 176: Quite odd that these selectors weren't included when it seems like they should've been. Adding these now might cause some changes for users but I think this is more of a fix than a change.

harriplappalainen (approved):
Thanks for the contribution! These are good small fixes and you provided good suggestions as well.

I think we should look into configuring the style linter as you proposed. I made [a ticket](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2023) of it.

We'll consider removing the span element from inside the Button component, I'll make a ticket. And as to the styles and unifying the implementation with the designs, we have [a PR of it](https://github.com/City-of-Helsinki/helsinki-design-system/pull/1088) that is waiting for better times as it proved to be a bigger challenge than expected as it requires changes to other HDS components that use HDS Button.

Thanks again for taking the time and effort to improve our dear little Button and giving some solid feedback 🏅

PR #1156, opened 2023-10-26
HDS-1974 align container with header

NikoHelle:
## Description

Container should have same size and responsive spacing as the Header

Closes [HDS-1974](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1974)

## Motivation and Context

Make it easy for devs to align content with Header

## How Has This Been Tested?

Locally and with Loki

[Demo](https://city-of-helsinki.github.io/hds-demo/hds-1974-container/?path=/story/components-container--aligns-with-header)

The css in not in core because Header is not in core. So core users cannot use Header or align content to it.

[HDS-1974]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SillanmakiA:
All good visually!

SillanmakiA (approved):
All good visually!

packages/react/src/components/container/Container.tsx
```diff
@@ -6,11 +6,11 @@ import '../../styles/base.css';
 import styles from './Container.module.scss';
 import classNames from '../../utils/classNames';
 
-export type ContainerProps = React.PropsWithChildren<React.HTMLProps<HTMLDivElement>>;
+export type ContainerProps = React.PropsWithChildren<React.HTMLProps<HTMLDivElement>> & { alignWithHeader?: boolean };
```
harriplappalainen, line 9: Maybe we should add JSDoc here about the prop? Something like "Align element's content with HDS Header's content." ?

NikoHelle: Yes, good point. Added!

laurakarhu (approved):
Seems to be working as expected :) Thanks!

harriplappalainen (approved):
Well done! 👏 Approved!

PR #1162, opened 2023-11-01
Fix login in prod

NikoHelle:
## Description

Login is not working in prod, because middleware does not work there. 

The callback urls are now static html files.

Also added documentation what settings can be used with localhost.


harriplappalainen (approved):
Blind approval 🙈 Let's see if this works.

laurakarhu (approved):
May the force be with you!

PR #1166, opened 2023-11-10
Remove multiple @types/react

NikoHelle:
Some packages have dependency @types/react: "*".

This installs @types/react 17 and 18 and causes ts errors.

Added types to esolutions in react/packages.json do not work, so added them to root.

Site does not use typescript, so it is not a problem that types 17 are fixed in root.

@types/cleave.js is not used, so removed it. It installed React 18 too.

## Related Issue

No Jira issue.

## Motivation and Context

VSCode shows a lot of false errors, because it messes up types/react 17 & 18. This is annoying and may hide actual errors.

## How Has This Been Tested?

Ran all linters and build scripts locally. No errors.

To test this run these commands:
- yarn cleanup
- yarn
- yarn build
- cd packages/react
- yarn ts-check
- yarn build-storybook

Also open some files that used to show errors like Login.stories.tsx

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -61,7 +61,6 @@
     "@testing-library/react": "11.2.0",
     "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/user-event": "^12.8.1",
-    "@types/cleave.js": "1.4.4",
     "@types/jest": "^26.0.1",
     "@types/jest-axe": "^3.5.3",
     "@types/react": "17.0.2",
@@ -154,7 +153,6 @@
     "yup": "^1.0.2"
   },
   "resolutions": {
-    "@types/react": "17.0.2",
     "babel-jest": "^26.0.1",
     "eslint-plugin-react": "7.27.1",
     "jest": "^26.0.1",
--- a/package.json
+++ b/package.json
@@ -23,6 +23,11 @@
     "rimraf": "^5.0.1"
   },
   "resolutions": {
+    "@types/react": "17.0.2",
+    "@types/react-dom": "17.0.2",
+    "@types/react-dom/@types/react": "17.0.2",
+    "@types/react-dom/**/@types/react": "17.0.2",
+    "**/@types/react-dom/**/@types/react": "17.0.2",
     "multer": "1.4.4-lts.1",
     "jpeg-js": "0.4.4",
     "json5": "2.2.2"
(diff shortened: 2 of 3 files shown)


packages/react/package.json
```diff
@@ -61,7 +61,6 @@
     "@testing-library/react": "11.2.0",
     "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/user-event": "^12.8.1",
-    "@types/cleave.js": "1.4.4",
     "@types/jest": "^26.0.1",
     "@types/jest-axe": "^3.5.3",
     "@types/react": "17.0.2",
```
harriplappalainen, line 66: It seems that you can remove this line as well and the root package.jsons resolutions apply and everything works! `"@types/react": "17.0.2",
`

harriplappalainen (commented):
Good work testing, trying and repeating the process 💪 These are annoying problems to solve.

packages/react/package.json
mikkojamG: `@types/react-dom` can also be removed 🤔

mikkojamG (approved):
Seems to work as intended 👍 But do check Harris comment about unnecessary dependencies in `react/package.json` 🙂

packages/react/package.json
NikoHelle: I'm a bit skeptic about removing both of them from React package's devDependencies. After the removal those won't be listed anywhere else than "resolutions". So they are not listed as any kind of project dependencies anymore. 

If those types are removed, @types/react-dom is not installed in any node_modules. But there are no errors in my VSCode.

Maybe the best fix could be removing those types? 

Docs that say they are needed:

https://react.dev/learn/typescript
https://blog.logrocket.com/how-to-use-typescript-react-tutorial-examples/

Google generally says they are not needed until typescript complains about mismatching types.

So which is it 😄

harriplappalainen: Shoot, I made the comment on the wrong line! I meant you can remove `"@types/react": "17.0.2",` from resolutions, not devDependencies, and everything works for me at least. Sorry!

harriplappalainen: I would imagine that those two are needed in devDependencies though...

NikoHelle: Well, it would have worked 😄

NikoHelle: I removed the "@types/react": "17.0.2" from React and got no errors. Thx!

harriplappalainen: Ok! This is good for me. I think it's all well and good that the packages are in React packages dependencies 👍

harriplappalainen (approved):
I approveth 🎩

PR #1169, opened 2023-11-20
HDS-2028-fix-checkbox-label-added-to-DOM-even-if-not-given

mrTuomoK:
## Description
The label is needed to style the checkbox correctly so fixing this issue was to add a class to remove the padding-left of the label when the label is not used.

## Related Issue
[HDS-2028](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2028)

## How Has This Been Tested
- running local tests

[HDS-2028]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/checkbox/checkbox.css
+++ b/packages/core/src/components/checkbox/checkbox.css
@@ -28,6 +28,7 @@
   display: flex;
   flex-wrap: wrap;
   min-height: var(--size);
+  min-width: var(--size);
   position: relative;
 }
 
--- a/packages/core/src/components/checkbox/checkbox.css
+++ b/packages/core/src/components/checkbox/checkbox.css
@@ -133,10 +133,10 @@
   padding-left: calc(var(--size) + var(--label-padding));
   padding-top: calc((var(--size) - (var(--label-font-size) * var(--lineheight-m))) / 2);
   position: relative;
+}
 
-  &.hds-checkbox__no-label {
-    padding-left: 0;
-  }
+.hds-checkbox .hds-checkbox__label--hidden {
+  padding-left: 0;
 }
 
 /* ERROR */
--- a/packages/core/src/components/checkbox/checkbox.css
+++ b/packages/core/src/components/checkbox/checkbox.css
@@ -134,6 +134,10 @@
   padding-left: calc(var(--size) + var(--label-padding));
   padding-top: calc((var(--size) - (var(--label-font-size) * var(--lineheight-m))) / 2);
   position: relative;
+
+  &.hds-checkbox__no-label {
+    padding-left: 0;
+  }
 }
 
 /* ERROR */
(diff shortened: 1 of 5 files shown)


mrTuomoK:
Still needs a fix for table checkboxes, so don't check this yet

packages/core/src/components/checkbox/checkbox.css
```diff
@@ -133,6 +133,10 @@
   padding-left: calc(var(--size) + var(--label-padding));
   padding-top: calc((var(--size) - (var(--label-font-size) * var(--lineheight-m))) / 2);
   position: relative;
+
+  &.hds-checkbox__no-label {
```
mikkojamG, line 137: I believe nesting classes does not work in css, or at least it should not 😀 So should move this outside this selector.

mikkojamG: I pressure we are using BEM? In that case I would interpret this as a modifier instead of a block:
`.hds-checkbox--no-label { ...`

mikkojamG (changes requested):

packages/core/src/components/checkbox/checkbox.css
mrTuomoK: Ah could really be that I named it wrongly 👍 fixing all...

mrTuomoK: And actually nesting classes does work and starts to be supported -> https://caniuse.com/css-nesting ping @mikkojamG But yeah maybe I'll make it the old-fashion-way then.

mikkojamG: I'm old school like that.. 🙂

mrTuomoK: Now this should be corrected as old-school :D

mrTuomoK:
Oops, the core version of this is now broken -> starting to fix

mikkojamG (approved):

harriplappalainen (approved):
Approved! A bit dumb fix at first glance but considering how the core Checkbox CSS is built, I agree that this is better than a complete overhaul.. 😅

mrTuomoK:
> Approved! A bit dumb fix at first glance but considering how the core Checkbox CSS is built, I agree that this is better than a complete overhaul.. 😅 I was thinking that should there be a Story for custom label but maybe it's not needed? Maybe not many projects have custom labels.. Or do they?

Yeah... I don't like this any more than you do but yep, without a larger rewrite this was the simplest solution. I might make a custom example still so I won't merge yet 🦖

harriplappalainen:
Reminder: 
- line to changelog 🗒️ 
- (optional) custom label story

laurakarhu (approved):
Thanks!

harriplappalainen (approved):

PR #1173, opened 2023-11-23
Hds 2031 custom header width

harriplappalainen:
## Description

Header `theme` prop now supports setting max-width and logo-height.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-2031

Closes #

## Motivation and Context

Some projects have a need for defining their own max-width for Header instead of using the default 1440px. Defining logo height was not asked for but it should be made possible since we already offer the possibility for users to use their own logos.

## How Has This Been Tested?
Localhost
[Storybook demo](https://city-of-helsinki.github.io/hds-demo/hds-2031-header-custom-max-width/?path=/story/components-header--with-full-features-custom-theme). Custom story has max-width and logo-height set.
[Site demo](https://city-of-helsinki.github.io/hds-demo/hds-2031-header-custom-max-width-site/components/header/customisation/)


NikoHelle:
Looks good! Easy to use solution!

Our Container component has a quite new prop `alignWithHeader`. It uses fixed max-width, which is the same as header-max-width. Now that the header max-width can be anything, there should be a way to use the same in the Container.

I think it is quite rare that Header max-width is changed and in those cases the users can create a wrapper for the Container which has the same max-width as the Header - or just set className wherever the Container with  `alignWithHeader` is used.

Otherwise the `--header-max-width` should be set to `:root` and the Header and Container should use that value. And that value should be overridden when needed.

harriplappalainen:
> Our Container component has a quite new prop alignWithHeader. It uses fixed max-width, which is the same as header-max-width. Now that the header max-width can be anything, there should be a way to use the same in the Container.
I think it is quite rare that Header max-width is changed and in those cases the users can create a wrapper for the Container which has the same max-width as the Header - or just set className wherever the Container with alignWithHeader is used.
Otherwise the --header-max-width should be set to :root and the Header and Container should use that value. And that value should be overridden when needed.

Yes, I could fiddle with this and put instructions in the docs. The variables are declared in the parent class `.header` so the variable overriding should be happen smoothly by giving the Header component a custom class with a different value for the variable.
To do:
- [ ] Give Header and Container a custom max-width and see how they could work together

NikoHelle (approved):

harriplappalainen:
I don't know why Accordion reference images changed. Loki updated them after I rebased to current development branch 🤷

Riippi (approved):

mrTuomoK (approved):
Shipiiiiit 👍

PR #1175, opened 2023-11-24
Inherit background color for dropdown button with placeholder text

Arkkimaagi:
## Description
Adding a background color to dropdown buttons that have placeholder in order to help working with automated accessibility tools like Siteimprove.

## Related Issue

Sorry, I do not have an issue for this at the moment.

## Motivation and Context

Some automated tools like Siteimprove may have problems checking contrast for absolutely positioned elements with transparent background. HDS dropdown has such an element when a HDS dropdown has placeholder text in it. The button triggering opening the dropdown is absolutely positioned and has a transparent background.

This change will make the button inherit the background color of its parent. It should make it easier for automated accessibility tools to check the color contrast of the placeholder text thus reducing manual work on large sites like helfi.

## How Has This Been Tested?

* The CSS change itself has been tested on the HDS Components page using browser devtools.
* This has been discussed briefly in Slack channel #dev-helfi-hds
    * Tuomo K. could not think of any problems that this could cause.
* This code change has not been tested in React environment

Screenshot of current transparent button background style
image

Screenshot after changing the button to inherit its parents background color. (There should be no visible difference between these two screenshots)
image

Screenshot of the same buttons with rgba(255,0,0,0.5) set as their background to show where the changes could be visible if there were any.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Changes that are not related to specific components
 
 - [Component] What has been changed
+- [Select] Placeholder text inherits background color
 
 #### Fixed
 
--- a/packages/react/src/components/dropdown/select/Select.module.scss
+++ b/packages/react/src/components/dropdown/select/Select.module.scss
@@ -26,6 +26,7 @@
   width: 100%;
 
   &.placeholder {
+    background-color: inherit;
     color: var(--placeholder-color);
   }
 


harriplappalainen (approved):
Looks ok. Loki tests pass and no other component is using Select component, and I can't think of any problems with this 🤔 So Approved! 🎉

laurakarhu (approved):
Nice and easy fix. Thanks!

Arkkimaagi:
So, in what version will this be available in?

laurakarhu:
> So, in what version will this be available in?

We save it for Christmas package 3.3.0 (Sorry, it slipped from 3.2 accidentally, my bad!)

PR #1179, opened 2023-12-01
HDS-2049-searchInput-instructions

mrTuomoK:
## Description

Added instructions not to use SearchInput if no suggestions are shown/used but use a TextInput instead. Added clear button to TextInput to enable clearing it.

## Related Issue

[HDS-2049](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2049)

## Motivation and Context

If SearchInput is used and no suggestions are shown, it will report "no results..." to screen readers etc.

[HDS-2049]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Riippi (approved):

mrTuomoK:
I'll make a demo of how to solve this still...

NikoHelle:
If the input has text, then there are two clear icons on hover

image

NikoHelle:
There should be tests to verify all new props work, like callbacks are called etc.
Br,
Test Maniac

packages/react/src/components/textInput/TextInput.stories.tsx
```diff
@@ -115,3 +117,53 @@ Playground.args = {
...
+    setInputValue(e.target.value);
+  };
+
+  const onClearButtonClick = () => {
+    action('clear search')();
+    const nativeInputValueSetter = Object?.getOwnPropertyDescriptor(window.HTMLInputElement.prototype, 'value').set;
+    nativeInputValueSetter?.call(ref.current, '');
+    ref?.current.focus();
+    ref?.current.dispatchEvent(new Event('change', { bubbles: true }));
+  };
```
NikoHelle, line 143: In my other comment I mentioned that this callback should be discarded.

If still kept, then this could be simplified with:

```
  const onClearButtonClick = () => {
    action('clear search')();
    const input = ref?.current as HTMLInputElement;
    setInputValue('');
    input.focus();
  };
```

 This requires the textInput to be controlled, but that should not be a problem
 
 `value={inputValue}`

NikoHelle (changes requested):

mrTuomoK:
Ok now it should be more like it -> @NikoHelle

NikoHelle (approved):
Looks very good! 💪

NikoHelle:
I think I didn't build earlier and that's why the native clear-button was visible. Didn't see it anymore.

site/src/docs/components/search-input/index.mdx
```diff
@@ -32,6 +32,7 @@ export default ({ children, pageContext }) => <TabsLayout pageContext={pageConte
 - **It is recommended to have only one search per page.** If you need multiple searches, clearly label them to indicate their purpose.
 - If possible, provide search suggestions (see With suggestions section for more info). This helps users to find what they are looking for, especially if the searchable items are complex.
     - Do not use suggestions to present search results. Results should be provided in their own component or a separate page.
+- When no search suggestions are provided, it's more advised to use a simpler search by using a TextInput component <Link openInNewTab openInNewTabAriaLabel="Opens in a new tab" href="/storybook/react/?path=/story/components-textinput--simple-search-input">View in Storybook</Link>.
```
harriplappalainen, line 35: Should we mention navigation to search result page here? And link to TextInput component docs too. So how about something like this (remember to add dot before View in storybook -link):
`- When no search suggestions are provided and search results are shown on another page, we recommend using a simpler search by using a <InternalLink href="/components/text-input">TextInput component</InternalLink>. <Link openInNewTab openInNewTabAriaLabel="Opens in a new tab" href="/storybook/react/?path=/story/components-textinput--simple-search-input">View in Storybook</Link>.`

harriplappalainen (commented):
I think it's reasonable to ask users to use this instead of the SearchInput for the use case where SearchInput has no suggestions and directs to a different search results page. Smart additions here! 🤓

harriplappalainen (approved):
Good work! 🚀

PR #1187, opened 2023-12-12
DRAFT/DEMO: HDS-1742 publish common js

NikoHelle:
## Description

Demo how `hds-js` package could be bundled and published. 

In short: 
- exported code is added to `packages/commonjs/index.ts`
- build compiles it and creates a `package.json` for the `hds-js`
- `lerna` publishes it like other packages (cannot be tested with a dry-run, I guess)

## Tested with
- imported the created bundle to `hds-cra` like other packages and seemed to work.

## Todo:
- some cleaning
- automatic versioning needs tuning, because of `base-package.json`

## Tried also:
- if `packages/commonjs` has its own package.json with build commands, resolving dependencies from "packages/react" does not work
- `rollup.config` cannot be in `packages/commonjs` either or all rollup dependencies must be there too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "build:tokens": "lerna run --scope hds-design-tokens build",
     "build:core": "lerna run --scope hds-core build",
     "build:react": "lerna run --scope hds-react build",
+    "build:commonjs": "lerna run --scope hds-js build",
     "build:site": "lerna run --scope site build",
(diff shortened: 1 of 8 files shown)


harriplappalainen:
This seems promising 👍 I think we could venture forth with this idea. I'm interested to see how the finished build looks like.

harriplappalainen:
This seems promising 👍 I think we could venture forth with this idea.

NikoHelle:
Closed because only a demo

PR #1195, opened 2023-12-27
HDS-2063-remove-unused-props

mrTuomoK:
## Description

There were unused props in Header.Link but after having a closer look, the whole Header LinkProps was useless.

## Related Issue

[HDS-2063](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2063)

## How Has This Been Tested?

- ran local tests

## Add to changelog
- I guess CHANGELOG doesn't need any additions since this change didn't affect anything, just removal of the unused typing. So not used, nowhere seen 🤷‍♂️ 

[HDS-2063]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/header/components/headerLink/HeaderLink.tsx
+++ b/packages/react/src/components/header/components/headerLink/HeaderLink.tsx
@@ -1,4 +1,4 @@
-import React, { MouseEventHandler, cloneElement, useCallback, useEffect, useRef, useState } from 'react';
+import React, { cloneElement, useCallback, useEffect, useRef, useState } from 'react';
 import { v4 as uuidv4 } from 'uuid';
 
 // import base styles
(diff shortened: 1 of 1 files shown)


NikoHelle:
Good clean up! Especially because there is another `LinkProps` exported from `LinkItem`.

I guess the removed LinkProps was not exported, so it could be imported like

`import { LinkProps } from hds-react`

I doubt anyone would have used that, but if possible, this PR would be a breaking change I think.

laurakarhu:
> Good clean up! Especially because there is another `LinkProps` exported from `LinkItem`.
> 
> I guess the removed LinkProps was not exported, so it could be imported like
> 
> `import { LinkProps } from hds-react`
> 
> I doubt anyone would have used that, but if possible, this PR would be a breaking change I think.

According to analytics, LinkProps has not been used (if it was possible in the first place). So let's remove it before anyone does use it.

laurakarhu (approved):
Thanks!

harriplappalainen:
> > Good clean up! Especially because there is another `LinkProps` exported from `LinkItem`.
> > I guess the removed LinkProps was not exported, so it could be imported like
> > `import { LinkProps } from hds-react`
> > I doubt anyone would have used that, but if possible, this PR would be a breaking change I think.
> 
> According to analytics, LinkProps has not been used (if it was possible in the first place). So let's remove it before anyone does use it.

I'll remove these devils you have painted on the walls by assuring that these could not have been used because `packages/react/src/components/header/index.ts` handles Header exports and this `LinkProps` isn't there 👍

NikoHelle (approved):

PR #1112, opened 2023-09-08
HDS-1873 Site Header component preview

mikkojamG:
## Description

Add note on documentation site about Header component, explaining that component is designed to be used full width at the top of the page, and therefore might not scale correctly on preview. Also add links to Storybook to stories that include relevant components. 

## Related Issue

Closes [HDS-1873](https://helsinkisolutionoffice.atlassian.net/browse/HDS-1873)

## How Has This Been Tested?
Tested on local machine

## Screenshots (if appropriate):
Screenshot 2023-09-08 at 13 38 18

[HDS-1873]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-1873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1198, opened 2023-12-28
HDS-1727 Scaling notes to Hero with diagonal koros examples

harriplappalainen:
## Description

Doc site examples of Hero with diagonal Koros don't adjust their content properly on medium sized screens. This happens because Hero uses CSS mediaqueries based on browser width but the component width is shrunk because of Playground component. Problem is similar to [Header scaling in doc site](https://github.com/City-of-Helsinki/helsinki-design-system/pull/1112) which couldn't be fixed and only notes were addes to the examples. I did the same here.

## Related Issue

https://helsinkisolutionoffice.atlassian.net/browse/HDS-1727

## How Has This Been Tested?
Localhost
[Demo](https://city-of-helsinki.github.io/hds-demo/hds-1727-notes-to-hero-diag-koros/components/hero/#diagonal-koro-shape-and-image-as-a-background)

## Add to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ Changes that are not related to specific components
 
 #### Added
 
-- [Component] What is added?
+- [Hero] Note about scaling to diagonal koros examples
 
 #### Changed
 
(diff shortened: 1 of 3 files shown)


laurakarhu (approved):
Thanks!

NikoHelle (approved):

PR #1205, opened 2024-01-04
Loki reference image clearing

mrTuomoK:
Add clearing loki reference images to loki update script and remove currently obsolete reference images

## Description

Remove loki reference images before updating them so they keep in sync with all stories and no obsolete ones are left hanging.

## How Has This Been Tested?

- local machine

## Add to changelog
- No need to add, doesn't affect normal users in any way


Riippi (approved):
LGTM

NikoHelle (approved):

PR #1216, opened 2024-01-19
Fix rounded links buttons on iOS 15.5

Arkkimaagi:
## Description

Fix iOS 15.5 (and hopefully 15.8) rounded link buttons.

## Related Issue

I do not have an issue yet, but I do have an slack message with a thread: https://helsinkicity.slack.com/archives/CHCV3KTHA/p1705669479076289

## Motivation and Context

It seems that at least iOS 15.5 and 15.8 render links in a weird way if `-webkit-appearance: button;` is set. They ignore border-radius and padding properties from css if it is set. More current iOS versions do not have this problem.

This problem comes probably from the original mindset of `.hds-button` only applying to `<button>` elements when the normalize.css code was being merged with the button code. Now that `.hds-button` is also used for `<a href="...`-elements, the assumption fails.

By modifying the button.css code we can get the original intent of normalize.css to work better in this case and exclude link buttons from the `-webkit-appearance: button` issue.

## How Has This Been Tested?

On an iOS simulator for iOS 15.5

Screenshot of a modified hds documentation page with `<button>` and two `<a href` links with similar styles but different rendering on iOS 15.5.

image
image

Screenshot of originating source for normalize.css
image

## Add to changelog
- [ ] Added needed line to changelog

### DIFF
--- a/packages/core/src/components/button/button.css
+++ b/packages/core/src/components/button/button.css
@@ -12,7 +12,6 @@
   * Normalize.css rules
   * Correct the inability to style clickable types in iOS and Safari.
   */
-  -webkit-appearance: button;
   background-color: var(--background-color, transparent);
   border: var(--border-width) solid var(--border-color, transparent);
   border-radius: 0;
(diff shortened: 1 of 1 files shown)


mrTuomoK (approved):
Ok, good fix 👍

laurakarhu (approved):
Looks good to me! Thanks!

PR #1218, opened 2024-01-29
(HDS-2107) remove normalize, fix styles in new builds

mrTuomoK:
- Remove normalize
- Add some normalize-styles to Header styles
- Fix card style compose
- Remove broken hds-core import in Highlight-component (hds-react)

## Description

Fixes the way how hds-core was imported in hds-react -component (Highlight) which broke some build-systems. The correct way is importing the react base.module.css from styles folder (which just imports the hds-core/lib/base.css). For some reason importing the hds-core base.css straight breaks on new build-systems stating it's not ok to bring global css styles. Storybook doesn't nicely cope with the change and needed to add the base.module.css as import in the inde.css currently.

## Related Issue

[HDS-2107](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2107)

## How Has This Been Tested?

- Tested with new hds-cra
- with oncoming hds-vite
- with nextjs test app
- Running visual tests locally

## Add to changelog
- [x] Added needed line to changelog  <-- this still needs to be done, should we instruct importing normalize if it's needed?

[HDS-2107]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/base.css
```diff
@@ -1,4 +1,12 @@
...
 @import url("utils/animations.css");
 @import url("utils/helpers.css");
+
+/*
+ * Normalize.css rule
+ * 1. Prevent adjustments of font size after orientation changes in iOS.
+ */
+
+html {
+  -webkit-text-size-adjust: 100%; /* 1 */
```
NikoHelle, line 11: Maybe some cleaning: remove the "`1.`" and "`/* 1 */`", because there is only one rule

packages/hds-js/package.json
```diff
@@ -9,7 +9,7 @@
   "esnext": "./index.js",
   "types": "./index.d.ts",
   "sideEffects": [
-    "base.min-*"
+    "base.*"
```
NikoHelle, line 12: Now that we know this is for css and "hds-js" has no css, this sideEffects can be removed.

packages/react/package.json
```diff
@@ -9,7 +9,7 @@
   "esnext": "./index.js",
   "types": "./index.d.ts",
   "sideEffects": [
-    "base.min-*"
+    "base.*"
```
NikoHelle, line 12: I'm a bit worried this is too broad rule, as this will affect files named "base.js", "base.component.tsx"

Could be just target the `base.module.css`? I gues it was `base.min-*` because the filename changed between builds?

NikoHelle (commented):
Excellent job! Few cleaning remarks.

PR #1227, opened 2024-02-13
fix: don't override default class names with Table className prop

jorilindell:
## Description
At the moment all default class names are overridden when using Table component `className` prop. The reason for this is the` {...rest}` statement after all the props in the TableContainer component. The `className` prop is added to the class names to solve this issue. `{...rest}` is also moved before other props to avoid other default props to be overridden as well.

## Motivation and Context

Motivation for this PR is to support className prop in the Table component. This can be used to add some minor tweaks for the default styles. Other reason for this PR is to avoid unexpected behaviour when using className prop.

## Add to changelog
- [ ] Added needed line to changelog


mrTuomoK (approved):
ok

PR #1234, opened 2024-03-03
(HDS-2157) Tag mark deprecated and changed properties

mrTuomoK:
## Description

Tag mark deprecated and changed properties in the next major release.

## Related Issue

[HDS-2157](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2157)

## How Has This Been Tested?

- local machine

## Add to changelog

[HDS-2157]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Changes that are not related to specific components
 
 - [Component] What has been changed
+- [Tag] Marked changed and removed properties in the next major release
 
 #### Fixed
 
--- a/packages/react/src/components/tag/Tag.tsx
+++ b/packages/react/src/components/tag/Tag.tsx
@@ -23,10 +23,12 @@ export type TagProps = {
   className?: string;
   /**
    * The aria-label for the delete button
+   * @deprecated Will be removed in the next major release.
    */
   deleteButtonAriaLabel?: string;
   /**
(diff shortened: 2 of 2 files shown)


NikoHelle:
The Navigation component was deprecated with

```
/**
 * Navigation will be removed in the next major release. Upcoming Header component will be the replacement.
 * @deprecated
 */
export const Navigation = ({
```

This might be enough?

https://jsdoc.app/tags-deprecated

mrTuomoK:
Hmm... didn't we agree not to deprecate the whole component? 🤔

NikoHelle:
Ah, true. Just replacing with new one.

NikoHelle:
But shouldn't we still mark the props deprecated?

https://dev.to/dgreene1/how-to-properly-deprecate-3027

Adding plain text to the props won't propably be seen anywhere unless devs actually follow type links or hover over a prop.

mrTuomoK:
Ok let's see... I though I just copied the way Riippi did earlier :D

mrTuomoK:
Now it should be properly done @NikoHelle . @deprecated all which's use will completely be changed or completely removed, so didn't deprecate for example `id` since default value is the only thing that will change.

laurakarhu (approved):
Making space for new shiny tag!

PR #1236, opened 2024-03-04
HDS-2155 table class fix

mrTuomoK:
## Description

At the moment all default class names are overridden when using Table component `className` prop. The reason for this is the` {...rest}` statement after all the props in the TableContainer component. The `className` prop is added to the class names to solve this issue. `{...rest}` is also moved before other props to avoid other default props to be overridden as well.

## Related Issue

[HDS-2155](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2155)

## Motivation and Context

Motivation for this PR is to support className prop in the Table component. This can be used to add some minor tweaks for the default styles. Other reason for this PR is to avoid unexpected behaviour when using className prop.

## How Has This Been Tested?

- running tests on local machine

## Add to changelog

[HDS-2155]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Changes that are not related to specific components
 - Removed old & deprecated individual `lodash` dependencies and replaced with the full package and importing the needed functions only.
 - [HeaderActionBarItemWithDropdown] Removed useless `@layer` css style which caused Jest/jsdom tests output errors.
 - [Checkbox] Layout issue when using external label with htmlLabelFor-attribute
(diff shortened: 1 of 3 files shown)


NikoHelle (changes requested):

laurakarhu (approved):
LGTM

PR #1239, opened 2024-03-08
Fix login tests

NikoHelle:
## Description

OIDC/Login tests failed seemingly randomly. Issue was with some login calls that were actual http-requests instead of mocked requests.

## How Has This Been Tested?

25 successful runs of PR's checks

## Add to changelog
Tests changed, no need for changelog


mrTuomoK (approved):
Good fix 👍

PR #1243, opened 2024-03-21
release-4.0.0 (HDS-2187) copy button styles to cookieConsent not to break new button

mrTuomoK:
## Description

Copy old Button's styles to CookieConsent's LanguageSelector which breaks when new Button is coming.

## Related Issue

[HDS-2187](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2187)

## How Has This Been Tested?

- local machine running tests

## Add to changelog
- Won't be needed since CookieConsent will be renewed

[HDS-2187]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/cookieConsent/languageSwitcher/LanguageSwitcherItem/LanguageSwitcherItem.module.scss
+++ b/packages/react/src/components/cookieConsent/languageSwitcher/LanguageSwitcherItem/LanguageSwitcherItem.module.scss
@@ -1,5 +1,5 @@
 @import '../../../../styles/common.scss';
-@import '../../../button/button.common.scss';
+@import 'button.common.scss';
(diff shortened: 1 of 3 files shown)


arpargo (commented):
Do I understand this correctly, we're duplicating the whole stylesheet? So If we want change to apply everywhere we need to make the change to multiple places? Shouldn't we instead have this file as common import somewhere where both components could access it from?

arpargo (approved):
👍  then this makes sense.

PR #1262, opened 2024-04-16
fix: Add onblur event for Tooltip parent

timwessman:
## Description

Tooltip parent didn't have an on blur event. Now it has. [JIRA HDS-2196](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2196)

## How Has This Been Tested?

Open Tooltip component in Storybooks -> click on (?) -button -> Tooltip comes visible -> press Tab -key -> Tooltip disappers

### DIFF
--- a/packages/react/src/components/tooltip/Tooltip.tsx
+++ b/packages/react/src/components/tooltip/Tooltip.tsx
@@ -127,6 +127,7 @@ export const Tooltip = ({
         aria-label={buttonLabel}
         aria-expanded={isTooltipOpen}
         onClick={onButtonClick}
+        onBlur={() => setIsTooltipOpen(false)}
       >
         <span aria-hidden="true">
           <IconQuestionCircle />


laurakarhu (approved):
Thanks!

packages/react/src/components/tooltip/Tooltip.tsx
```diff
@@ -127,6 +127,7 @@ export const Tooltip = ({
         aria-label={buttonLabel}
         aria-expanded={isTooltipOpen}
         onClick={onButtonClick}
+        onBlur={() => setIsTooltipOpen(false)}
```
mrTuomoK, line 130: BTW is there a performance issue when using like this instead of using for example `const onBlurHandler = () => setToolTipOpen(false);` outside the render and calling it like `onBlur={onBlurHandler}`. If I remember correctly, React recreates the functions when rendering if declared like {() => {}}.

timwessman: interesting, could be.

PR #1269, opened 2024-04-24
feature/new-cookie-consent: vanilla js classes demo

NikoHelle:
## Description

Demo how vanilla-js class can be added to hds-js and used in React.


NikoHelle:
Demo not needed anymore

PR #1278, opened 2024-05-02
Fix "ref" prop in TagProps

NikoHelle:
## Description

The Tag component showed an error when using it like

```

const props:TagProps = {...};
<Tag ̇{...props} />

```

## How Has This Been Tested?

Added TagProps to components rendering Tags with props (available only in Stories).

## Add to changelog
No need


mrTuomoK (approved):
Good good 👍

PR #1280, opened 2024-05-02
HDS-2241 scss to core

NikoHelle:
## Description

Scss-compiling to core package. Same scss stylelint-rules as React. Added test files, which are not added to the compiled bundle. They could be removed when there is an actual scss file in the core.

## Related Issue

Closes [HDS-2241](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2241)

## How Has This Been Tested?
Added a new `core/components/scss` and checked the built css files.

## Demos:
No visible changes or any functionalities that affect demos

## Add to changelog
- internal change


packages/core/src/components/scss/_partial.scss
laurakarhu, line 1: Is this only for test purposes?

packages/core/src/components/scss/test.scss
```diff
@@ -0,0 +1,79 @@
+/* examples from https://sass-lang.com/guide/ */
+
+@use "sass:math";
+@use "partial.scss";
+@use 'button.scss' as Button;
+
+$font-stack: helvetica, sans-serif;
```
laurakarhu, line 7: Is this for test purposes?

laurakarhu (approved):
There are fonts and colors that are not according to HDS but I guess for tests and examples it doesn't matter. But if those scss files are used for actual components, they should be changed.

packages/core/src/components/scss/test.scss
NikoHelle: Yes, these are copied examples from scss home page for testing the features. I can change these to avoid confusion.

packages/core/src/components/scss/_partial.scss
NikoHelle: Removed and added actual scss from hel.fi cookies

packages/core/src/components/scss/test.scss
NikoHelle: Removed and added actual scss from hel.fi cookies

NikoHelle:
Removed test scss and added actual scss from hel.fi cookies

mrTuomoK (approved):
Superrrr 👍

PR #1296, opened 2024-06-06
(HDS-2240) fix read-only input styles

mrTuomoK:
## Description

If the value of a read-only text field is "" (empty), when using tab to navigate through the page, the user loses indication to where the tab focus is. With other values than empty the text is highlighted with white on blue background, but when there is no text, nothing visible changes for user. If there are multiple empty readonly fields in a row, this problem of course increases manyfold

## Related Issue

[HDS-2240](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2240)

## How Has This Been Tested?

- local machine running tests

## Demos:

image

## Add to changelog

[HDS-2240]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nooranen (approved):
Checked and it looks great to me👍

NikoHelle (approved):

SillanmakiA (approved):

PR #1308, opened 2024-07-03
Added "feature/" to target branches of demos

NikoHelle:
## Description

Demos are not created for "feature"-branches. Added them to branch filters.

The branches are single-quoted like Github's own examples with names with slashes.


mrTuomoK (approved):
oook 🥃

PR #1313, opened 2024-07-12
Obsolete: Release-4.0.0: (HDS-2069) New Select part 2

NikoHelle:
## Description

Second part of the new Select. 

Added:

- onChange handler
- disabled, required, invalid states
- text provider
- Assistive text
- Error text

## How Has This Been Tested?

Jest and local testing

## Demos:

## Add to changelog
- Added in last PR


packages/react/src/components/select/Select.module.scss
```diff
@@ -48,6 +57,19 @@
     padding: var(--spacing-s);
     padding-right: var(--spacing-2-xs);
 
+    &.withVisibleFocus {
+      align-self: center;
+      height: calc(calc(var(--dropdown-height)) - 6px);
```
mrTuomoK, line 62: Is this 6px just by eye-testing, could it be like half of something or? If some other height changes this will break?

packages/react/src/components/select/Select.stories.tsx
```diff
@@ -152,10 +162,117 @@ export const WithGroups = () => {
...
+    invalid: false,
+    texts: { ...defaultTexts, label: 'Controlled select' },
+  });
+
+  const resetSelections = () => {
+    updateProps({ ...props, groups: clearAllSelectedOptions(props.groups as Group[]), open: false });
+  };
+
+  const toggleDisable = () => {
+    const current = props.disabled;
```
mrTuomoK, line 204: Should these be written like `const { disabled } = props;` and then `updateProps({ ...props, disabled: !disabled });`. Though dunno which is better 🤷

packages/react/src/components/select/Select.module.scss
NikoHelle: image

The `-6px` makes the focus area of the clear button to fit inside the dropdown button area, which is `--dropdown-height`.

Nothing else is currently affecting the relationship, except if the border sizes change in the future.

packages/react/src/components/select/Select.stories.tsx
NikoHelle: Your suggested way is cleared! Good point! Fixed.

packages/react/src/components/select/dataUpdater.ts
```diff
@@ -10,58 +16,125 @@ import {
...
+        ...returnValue,
+        didDataChange: true,
+      };
+    }
+  }
+
+  return returnValue;
+};
+
+export const changeChandler: ChangeHandler<SelectData, SelectMetaData> = (event, dataHandlers): boolean => {
```
laurakarhu, line 109: just out of curiosity, why is it change**C**handler and not changeHandler?

laurakarhu (approved):
Looks good to me, left one question that is purely curiosity.

packages/react/src/components/select/dataUpdater.ts
NikoHelle: Oh, how blind you can be to your own code. 😄 That misspelling has been there all along. Good catch! Changed to "changeHandler"

mrTuomoK (approved):
With my after-vacation-eyes couldn't find any more to say so good work 💪

PR #1314, opened 2024-07-12
HDS-2324 and HDS-2331 initial PlayWright tests and running in docker

mrTuomoK:
## Description

Adds initial PlayWright tests for new Button and new Tag components. Also had to make possible to run the tests in docker (running `run.sh` in `/e2e/docker` directory should do the trick). Needs Docker to be installed of course. Running the script will run the tests and compares the reference images. If test are ran locally, screenshot comparisons will fail due to font-differences.

Also the corresponding loki-tests are skipped and the reference images for those are removed.

## Related Issue

[HDS-2324](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2324)
[subtask HDS-2331](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2331)

## Motivation and Context

We want to replace the old and less extensive loki tests.

## How Has This Been Tested?

- running locally, in Docker and in CI

## Demos:

## Add to changelog

- no changelog additions needed, but in the future, instructions for developers.

[HDS-2324]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/playwright.config.js
```diff
@@ -36,8 +37,8 @@ export default defineConfig({
 
   projects: [
     {
-      name: 'logged-out',
-      testMatch: [/pages/],
+      name: 'tests',
+      testMatch: [/tests/],
```
NikoHelle, line 41: Could test files be where the components are? Or does mean the resulting images also have to be there?

It would be clearer to have the test files where Jest files are - with the code.

e2e/tests/components/button-spec.ts
```diff
@@ -0,0 +1,79 @@
...
+          });
+        }
+      }
+    });
+
+    test(`subTest-${device.isMobile ? 'mobile' : 'desktop'}`, async ({ page }) => {
+      page.setViewportSize(device.viewport);
+      for (const componentUrl of componentUrls) {
+        await page.goto(`file://${componentUrl}`);
+        const iFrame = await page.frameLocator('iframe[title="storybook-preview-iframe"]');
```
NikoHelle, line 37: If the url is opened without the iframe, the tests should be clearer/faster?

iframe.html?args=&id=<testid>&viewMode=story

e2e/tests/components/tag-spec.ts
```diff
@@ -0,0 +1,77 @@
...
+        const name = `${componentUrl.split('/').pop()}-${device.isMobile ? 'mobile' : 'desktop'}`;
+
+        // take screenshot, also waits for all to render
+        await expect(container).toHaveScreenshot(`${name}.png`);
+
+        const tags = await container.locator('[data-playwright="true"]').all();
+
+        for (const [index, tag] of tags.entries()) {
+          await tag.scrollIntoViewIfNeeded();
+          const tagBoundingBox = await tag.boundingBox() || { x: 0, y: 0, width: 0, height: 0 };
```
NikoHelle, line 49: When can the boundingBox be missing?

e2e/tests/components/button-spec.ts
```diff
@@ -0,0 +1,79 @@
...
+        for (const [index, button] of buttons.entries()) {
+          await button.scrollIntoViewIfNeeded();
+          const buttonBoundingBox = await button.boundingBox() || { x: 0, y: 0, width: 0, height: 0 };
+          const outside = { x: buttonBoundingBox.x - 1, y: buttonBoundingBox.y - 1 };
+          const screenshotPrefix = `${name}-components-${index}`;
+
+          // to make sure nothing is focused
+          await page.mouse.move(outside.x, outside.y);
+          await page.mouse.down();
+          await page.mouse.up();
```
NikoHelle, line 57: When making the final code, it would be clearer to make helper functions like: 
clickOutside(element)
mouseDownInElementCenter(element)

And even:

takeScreenshotOfHoverState(element);

NikoHelle (commented):
Great that this kind of tests can be done! Surprisingly simple even!

This is an initial PR, but looks good. Just remove code repetition with helper functions and it's ready to go!

e2e/playwright.config.js
mrTuomoK: Didn't yet look into that, good point.

e2e/tests/components/button-spec.ts
mrTuomoK: Yeah, we could make a "unfocus" element helper for this.

e2e/tests/components/tag-spec.ts
mrTuomoK: TypeScript says so, so I guess it can return some null etc 🤷

e2e/tests/components/button-spec.ts
mrTuomoK: Checking....

mrTuomoK: Well, it's quite the same to formulae the url from the story button link url than to just locate the iframe 🤷

e2e/tests/components/button-spec.ts
mrTuomoK: Added some helpers

e2e/tests/components/button-spec.ts
mrTuomoK: Done!

e2e/playwright.config.js
mrTuomoK: This might be worth a ticket of it's own since it might need some thinking and restructuring if possinble.

e2e/playwright.config.js
```diff
@@ -22,22 +23,36 @@ export default defineConfig({
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL: process.env.E2E_TESTS_ENV_URL ?? "https://hds.hel.fi",
     ignoreHTTPSErrors: true,
-    screenshot: {
-      fullPage: true,
-      mode: "only-on-failure"
-    },
+    // screenshot: {
```
NikoHelle, line 26: These comments could be removed. And line 34.

NikoHelle (approved):
Very comprehensive interactive tests! Will make tests much better and the code looks easy to use.

PR #1318, opened 2024-07-16
HDS-2095: Added clip-path to Playground previews

NikoHelle:
## Description

Header examples overflow the container and show an ugly grey box.

Overflow:hidden would hide the menus, so clip-path was best way to fix this.

A new wrapped was needed or the container box would also be clipped (borders etc)

## Related Issue

Closes [HDS-2095](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2095)

## How Has This Been Tested?

Locally. 

**NOTE: the menus are still broken. Now just the grey boxes won't show. There is another PR to fix the dropdown menus. Main navigation menu opens outside the whole preview area, because their position is "fixed". **

## Demos:

## Add to changelog
Not an actual documentation change.


site/src/components/Playground.scss
```diff
@@ -118,6 +118,10 @@
   }
 }
 
+.clipped-view-area {
```
mrTuomoK, line 121: Just a nuisance but could/should this class be named something like .playground-clipped-view-area etc?

mrTuomoK: Else looks good though 👍

NikoHelle: The class name is in Playground.scss and used only in Playground, so adding that seems overstating the fact it is for playground. 😄 

But all other classes have that, so I'll add it. Good catch!

NikoHelle: Fixed.

mrTuomoK (approved):
Good work 👍

PR #1338, opened 2024-08-29
Release-4.0.0: (hds-2297) SCSS compiler to core

NikoHelle:
## Description

Compile and export SCSS styles in Core

Converted one component to SCSS. Others should be converted later to keep this PR small. 

The React version uses SCSS from.

Added also a `_bem.scss` that helps to create BEM compliant styles.

## Related Issue

Closes [HDS-2297](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2297)

## How Has This Been Tested?

Checked that the generated core/lib/.css file is identical to the one now in npm. 

Checked that the scss files are copied. 

## Demos:

## Add to changelog

- [x ] Added needed line to changelog

[HDS-2297]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/copyFiles.js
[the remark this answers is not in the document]
NikoHelle: It only part of the build process. Not for external users.

packages/core/copyFiles.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * Copies a file with error handling
+ */
+async function copyFile(sourceLocation, targetLocation) {
+  try {
+    await fs.copyFile(sourceLocation, targetLocation);
+  } catch (e) {
+    Promise.resolve(e);
```
mrTuomoK, line 32: Hmm... why resolving the Promise here (and not Promise.reject(e))? And hmm... why altogether have a separate function at all but just use the fs.copyfile where this additional function is called? Or is there something I'm missing here 🤔

NikoHelle: Rejections must be handled, so always resolving prevents catching again.

But, true, could remove the helper / extra function.

mrTuomoK: Everything else seems fine, just this small refactor might be in place. good work 👍

NikoHelle: Updated.

PR #1342, opened 2024-09-04
HDS-2384: Hide mobilemenu actionItems when menu is closed

NikoHelle:
## Description

The visually hidden items could be navigated with keyboard.

Mobile menu contents are visually hidden when menu is closed and while not animating.

Changes in the visual tests are because the bottom line is shifted a bit, now that menu contents are not taking visual space.

Attached diff images:
chrome_iphone7_Components_Header_Minimal_With_Custom_Logo
chrome_iphone7_Components_Header_Minimal

## Related Issue

Closes [HDS-2384](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2384)

## How Has This Been Tested?

With VoiceOver

## Demos:

## Add to changelog

No need for changelog.

[HDS-2384]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
```diff
@@ -513,12 +513,23 @@ export const HeaderActionBarNavigationMenu = ({
       id="hds-mobile-menu"
       onTransitionEnd={animationDone}
     >
-      {actionBarItems?.map?.((item: HeaderActionBarItemProps) => {
-        if (typeof item === 'object') {
-          return React.cloneElement(item as unknown as React.ReactElement, { fullWidth: true, key: getKey(item) });
-        }
-        return null;
-      })}
+      {actionBarItems && actionBarItems.length && (
```
mrTuomoK, line 516: Hmm... would it work by just modifying the previous implementation with `{ mobileMenuOpen && actionBarItems?.map?.` and so on?

mrTuomoK (commented):
Left a comment

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
NikoHelle: actionBarItems is an empty array by default (created in ActionBar), so the wrapper would be rendered even if there are no items.

mrTuomoK: So the wrapper is needed even if empty?

NikoHelle: I guess I didn't understand the question. 😄 

The new code makes sure the wrapper is **not** added, if there are zero items in the `actionBarItems`. The `actionBarItems` always exists, so length must be checked.

The `mobileMenuOpen` is not checked like it wasn't checked before. I think because menu needs to be rendered for open/close animations to work.

mrTuomoK: Hmm... didn't see any difference after simpifying it to be like this:

```suggestion
    <div
      className={classNames(styles.headerNavigationMenu, mobileMenuOpen && styles.mobileMenuOpen)}
      id="hds-mobile-menu"
      onTransitionEnd={animationDone}
      aria-hidden={!mobileMenuOpen}
    >
      {actionBarItems?.map?.((item: HeaderActionBarItemProps) => {
        if (typeof item === 'object') {
          return React.cloneElement(item as unknown as React.ReactElement, {
            fullWidth: true,
            key: getKey(item),
          });
        }
        return null;
      })}
```

mrTuomoK: ⬆️  would also get rid of the need for the extra wrapper

mrTuomoK (changes requested):
The wanted change was the one we discussed earlier, just made a review out of this so it show's up properly :)

packages/react/src/components/header/components/headerActionBar/HeaderActionBarNavigationMenu.tsx
NikoHelle: This got interesting...

Reverted back with simple "aria-hidden" in the parent. The hidden "Kirjaudu"-button got still focus with tab. There was a error in the console about "Chrome refusing to set aria-hidden to element with active items."

Reverted all changes and just set visibility hidden to the parent when menu is not open. It works now.

NikoHelle: Now it's working again and reviewable.

mrTuomoK (approved):
Looks good to me 👍

laurakarhu (approved):
Looks good to me!

PR #1367, opened 2024-09-23
Hds 2394 core scss v4 for button

timwessman:
## Description

- Convert the CSS of the Button component to SCSS. 
- Documentation added

## Related Issue

- Fixed also Status-label -component to use mixin function on React SCSS classes (and fixed documentation)

Closes [HDS-2394](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2394)

## How Has This Been Tested?

- visual tests
- e2e tests

## Demos:

Links to demos are in the comments

## Add to changelog

- [x] Added needed line to changelog

PR #1369, opened 2024-09-26
Hds 2393 core scss v4 for breadcrumbs

timwessman:
## Description

- Convert the CSS of the Breadcrumb component to SCSS. 
- Documentation added

## Something to think

- In `react/src/components/breadcrumb/Breadcrumb.module.scss` I keeped `composes: from` -notation to split elements from css selectors. It is actually quite clean to do it this way, but its not the same logic that I used in the [Button -component](https://github.com/City-of-Helsinki/helsinki-design-system/pull/1367). So which one is better? or something else?

## Related Issue

Closes [HDS-2393](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2393)

## How Has This Been Tested?

- visual tests
- e2e tests

## Add to changelog

[HDS-2393]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/breadcrumb/_breadcrumb-mixin.scss
```diff
@@ -0,0 +1,160 @@
...
+    &--mobile {
+      display: flex;
+
+      #{BemUtil.get-block(&)}__link {
+        font-size: var(--fontsize-body-l);
+      }
+    }
+  }
+
+  @media (min-width: 768px) {
```
mrTuomoK, line 93: I wonder if some variable is available (breakpoint?) for these?

e2e/tests/core/components/core-breadcrumb-spec.ts-snapshots/core-components-breadcrumb--breadcrumb-desktop.png
mrTuomoK, line 1: Also make sure that the old loki-screenshots are deleted and add to loki config to skip breadcrumb testing (this should always be done when creating new e2e test for components until all loki tests are replaced)

packages/core/src/components/breadcrumb/_breadcrumb-mixin.scss
timwessman: sure, fixed

e2e/tests/core/components/core-breadcrumb-spec.ts-snapshots/core-components-breadcrumb--breadcrumb-desktop.png
timwessman: done

timwessman:
Hups, this is broken now (it renders only the last item of the Breadcrumb), I'll fix it.

timwessman:
Found out that you can't use css variables on media query declarations, so I reverted breakpoints back to numerical values. There have to be some trick, maybe from scss, but I think it's not on scope of this pr.

PR #1377, opened 2024-10-10
(Release 4.0.0): HDS-2069: Select part 5

NikoHelle:
## Description

Added keyboard commands, focus shifting and handling, onFocus/onBlur and active-descendant for accessibility.

Added also Playwright-tests and PW utils.

Missing:
- theming
- storage for making easier for users to memoize props.
- docs

## Related Issue

Closes [HDS-2069](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2069)

## How Has This Been Tested?

Unit and Playwright

## Demos:

## Add to changelog

Changelog added later with last part

[HDS-2069]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react/src/components/select/Select.stories.tsx
```diff
@@ -588,5 +497,124 @@ export const VirtualizedSingleselectWithoutGroups = () => {
...
+          .blurred {
+            padding: 10px;
+          }
+
+          .focused {
+            background-color: #defcde;
+          }
+
+          .blurred {
+            background-color: #ececec;
```
laurakarhu, line 569: We should not put any non-brand-colors to examples, as others might consider it as allowed color. Hds color-black-10 would be #e6e6e6

packages/react/src/components/select/Select.stories.tsx
```diff
@@ -588,5 +497,124 @@ export const VirtualizedSingleselectWithoutGroups = () => {
...
+    <>
+      <style>
+        {`
+          .focused,
+          .blurred {
+            padding: 10px;
+          }
+
+          .focused {
+            background-color: #defcde;
```
laurakarhu, line 565: $color-tram-medium-light | #a3e3c2 would be similar HDS color

laurakarhu (approved):
Very focused. Minor remarks on the example colors, those can be closed after changes.

packages/react/src/components/select/Select.stories.tsx
NikoHelle: Good idea, I'll change these.

PR #1381, opened 2024-10-14
Release-4.0.0 HDS-2419 remove useless native attributes

mrTuomoK:
## Description

Remove useless native props documentations

## Related Issue

Closes [HDS-2419](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2419)

## How Has This Been Tested?

- local tests

## Demos:

## Add to changelog

- Not needed

[HDS-2419]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/textInput/TextInput.tsx
+++ b/packages/react/src/components/textInput/TextInput.tsx
@@ -33,10 +33,6 @@ export type TextInputProps = MergeAndOverrideProps<
      */
     defaultValue?: string;
     /**
-     * If `true`, the input will be disabled
-     */
-    disabled?: boolean;
-    /**
      * The error text content that will be shown below the input
      */
     errorText?: string;
@@ -72,14 +68,6 @@ export type TextInputProps = MergeAndOverrideProps<
      * Short hint displayed in the input before the user enters a value
      */
     placeholder?: string;
-    /**
-     * If `true`, prevents the user from changing the value of the field (not from interacting with the field)
-     */
-    readOnly?: boolean;
-    /**
-     * If `true`, the label is displayed as required and the `input` element will be required
-     */
-    required?: boolean;
     /**
      * Override or extend the styles applied to the component
      */
--- a/packages/react/src/components/textarea/TextArea.tsx
+++ b/packages/react/src/components/textarea/TextArea.tsx
@@ -18,10 +18,6 @@ export type TextAreaProps = MergeAndOverrideProps<
      */
     defaultValue?: string;
     /**
-     * If `true`, the textarea will be disabled
-     */
-    disabled?: boolean;
-    /**
      * The error text content that will be shown below the textarea
      */
     errorText?: string;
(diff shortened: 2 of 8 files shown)


packages/react/src/components/textInput/TextInput.tsx
```diff
@@ -72,14 +68,6 @@ export type TextInputProps = MergeAndOverrideProps<
      * Short hint displayed in the input before the user enters a value
      */
     placeholder?: string;
-    /**
-     * If `true`, prevents the user from changing the value of the field (not from interacting with the field)
-     */
-    readOnly?: boolean;
-    /**
-     * If `true`, the label is displayed as required and the `input` element will be required
-     */
```
NikoHelle, line 81: The prop is native and therefore useless to list here, but setting the value `true` affects also the visuals (adds the "*"), so that should be mentioned somewhere. It is not in docs at the moment.

Perhaps it should be?

packages/react/src/components/textarea/TextArea.tsx
```diff
@@ -45,18 +41,10 @@ export type TextAreaProps = MergeAndOverrideProps<
...
-     */
-    onChange?: React.ChangeEventHandler<HTMLTextAreaElement>;
     /**
      * Short hint displayed in the textarea before the user enters a value
      */
     placeholder?: string;
-    /**
-     * If `true`, the label is displayed as required and the `textarea` element will be required
-     */
-    required?: boolean;
```
NikoHelle, line 59: Same "*" thing with this one.

packages/react/src/components/textInput/TextInput.tsx
mrTuomoK: I'll add

packages/react/src/components/textarea/TextArea.tsx
mrTuomoK: same

packages/react/src/components/textInput/TextInput.tsx
mrTuomoK: Hmm, on site there's this "To decrease a cognitive load of the user, it is recommended to mark all required fields. Remember to provide explanation of the required field indicator at the top of your form." ping @NikoHelle

NikoHelle: Yes, but I think we should somewhere clearly say that if required=true there will be an asterix added after the label.

This part of JSDocs is nowhere after deletion "the label is displayed as required".  Should it be moved to docs?

mrTuomoK: Looks like this now on the doc-site 
image

mrTuomoK: ping @laurakarhu is this ☝️ ok?

NikoHelle (approved):

PR #1389, opened 2024-10-25
(Release-4.0.0) HDS-2069: Select docs

NikoHelle:
## Description

Docs, migration guide and changelog

## Related Issue

Closes [HDS-2069](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2069)

## Demos:

Links to demos are in the comments

## Add to changelog

- [ x] Added needed line to changelog

PR #1401, opened 2024-11-04
(Release-4.0.0) HDS-2069 Select accessibility fixes

NikoHelle:
## Description

Latest accessibility auditing found these issues fixed in this PR:
- search errors are not announced to screen reader like other filter/search notifications
- single select group labels can be left out of navigation
- "show less" -button should not lose focus on click. Also a screen reader notification should be announced.

Also added:
- clearable -prop, so select cannot be cleared after selection
- themes for tags. Added also same `--computed-` that are in new scss components.
- fixed some translation keys
- hidden selected option counts must be re-calculated with resizeObserver when component is resized.

## Related Issue

Closes [HDS-2069](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2069)

## How Has This Been Tested?

Jest, e2e

## Demos:

## Add to changelog

- changelog in [docs PR](https://github.com/City-of-Helsinki/helsinki-design-system/pull/1389)

[HDS-2069]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react/src/components/select/Select.module.scss
```diff
@@ -252,23 +255,15 @@
...
-        --menu-item-background-hover: var(--color-bus-light);
-        --menu-item-background-selected: var(--color-bus);
-        --menu-item-background-selected-hover: var(--color-bus);
-        --menu-item-color-selected: var(--color-white);
-        --menu-item-color-selected-hover: var(--color-white);
-
-        &.disabledOption.disabledOption {
-          color: var(--color-black-40);
-        }
+        --computed-menu-item-background-hover: var(--menu-item-background-hover, var(--color-bus-light));
```
mrTuomoK, line 262: hmm... should all these `*-background-*` be `*-background-color-*`? and if there's more of these in which it only affects a -color?

mrTuomoK (changes requested):
Commented about the -background- ->  -background-color-

packages/react/src/components/select/Select.module.scss
NikoHelle: The naming is copied from the old Select, so haven't really thought about it. The only thing that can change in the background is the color.

mrTuomoK: So then I'd prefer it to be *-color* :D everywhere

NikoHelle: Changed!

mrTuomoK (approved):
Good work again!

PR #1403, opened 2024-11-04
fix: scss include bug

timwessman:
## Description

Styles of some components didn't work on the documentation site code/Core -tab. The components that were not working were those with scss styles. They were fixed by using `@use` instead of `@import` when including stylesheets to `all.css`.


mrTuomoK (approved):
Nice 🧑‍🚒

PR #1404, opened 2024-11-06
Release-4.0.0 HDS cookie small fixes, timeout 50 -> 500ms, fix tests...

mrTuomoK:
Small fixes, timeout 50 -> 500ms, fix tests clean code a bit, rename typoted class

## How Has This Been Tested?

- running tests

## Demos:

### DIFF
--- a/packages/react/src/components/cookieConsentCore/cookieHandler.js
+++ b/packages/react/src/components/cookieConsentCore/cookieHandler.js
@@ -466,16 +466,10 @@ export default class CookieHandler {
     }
 
     const siteSettingsGroups = [...this.#siteSettings.requiredGroups, ...this.#siteSettings.optionalGroups];
+    const groupIds = siteSettingsGroups.map((group) => group.groupId);
+    const duplicateGroupNames = groupIds.filter((groupId, index) => groupIds.indexOf(groupId) !== index);
 
-    const cookieNames = new Set();
-    const duplicateGroupNames = new Set();
-    siteSettingsGroups.forEach((group) => {
-      if (cookieNames.has(group.groupId)) {
-        duplicateGroupNames.add(group.groupId);
-      }
-      cookieNames.add(group.groupId);
-    });
-    if (duplicateGroupNames.size > 0) {
+    if (duplicateGroupNames.length > 0) {
       throw new Error(
         `Cookie consent: Groups '${Array.from(duplicateGroupNames).join(', ')}' found multiple times in settings.`,
       );
--- a/packages/react/src/components/cookieConsentCore/monitorAndCleanBrowserStorages.js
+++ b/packages/react/src/components/cookieConsentCore/monitorAndCleanBrowserStorages.js
@@ -6,7 +6,7 @@ import { parse } from 'cookie';
 
 import deleteCookie from './deleteCookie';
 
-export default class MonitorAncCleanBrowserStorages {
+export default class MonitorAndCleanBrowserStorages {
   // MARK: Public properties
   BROWSER_STORAGES = ['cookie', 'localStorage', 'sessionStorage', 'indexedDB', 'cacheStorage'];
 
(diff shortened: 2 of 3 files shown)


packages/react/src/components/cookieConsentCore/cookieHandler.js
```diff
@@ -466,16 +466,10 @@ export default class CookieHandler {
     }
 
     const siteSettingsGroups = [...this.#siteSettings.requiredGroups, ...this.#siteSettings.optionalGroups];
+    const groupIds = siteSettingsGroups.map((group) => group.groupId);
+    const duplicateGroupNames = groupIds.filter((groupId, index) => groupIds.indexOf(groupId) !== index);
```
NikoHelle, line 471: I think the fastest would be

```
const uniqueGroupIds = new Set(groupIds); // or(...groupIds)
if(uniqueGroupIds.size != groupIds.length){....}
```

packages/react/src/components/cookieConsentCore/monitorAndCleanBrowserStorages.js
```diff
@@ -45,59 +45,50 @@ export default class MonitorAncCleanBrowserStorages {
...
-        key === '' || // If the key is empty, filter it out
-        this.#isKeyConsented(key, consentedKeysArray) // If key is consented (with possible wildcards), filter it out
-      ) {
-        return false;
-      }
-      return true;
+      return !(key === '' || this.#isKeyConsented(key, consentedKeysArray));
     });
+    // console.log('deleteKeys', deleteKeys, deleteKeys.length);
+    deleteKeys.forEach((key) => {
```
NikoHelle, line 51: Remove the .logs?

packages/react/src/components/cookieConsentCore/monitorAndCleanBrowserStorages.js
```diff
@@ -297,9 +285,9 @@ export default class MonitorAncCleanBrowserStorages {
    * @param {number} [monitorInterval=500] - The interval in milliseconds at which to monitor the stored keys.
    * @param {boolean} [remove=false] - Indicates whether to remove the stored keys or not.
    */
-  init(cookieHandler, monitorInterval = 0, remove = false) {
+  init(cookieHandler, monitorInterval = 500, remove = false) {
```
NikoHelle, line 288: This enables the monitoring by default. Should it still be "0" by default?

NikoHelle (changes requested):

packages/react/src/components/cookieConsentCore/cookieHandler.js
mrTuomoK: Hah thought of the exact same thing later but didn't change it :D

packages/react/src/components/cookieConsentCore/monitorAndCleanBrowserStorages.js
mrTuomoK: will do

packages/react/src/components/cookieConsentCore/monitorAndCleanBrowserStorages.js
mrTuomoK: Hmmm I actually don't know should or shouldn't it be since all the JSONs set the monitorInterval and the default value as documented right above says it's 500?

mrTuomoK: Although it won't remove anything if remove = false (default)

NikoHelle: Ok, both ways are good, as long as it is not 50ms. We'll see how many complain about tons of console.log warnings. 😄

NikoHelle (approved):

PR #1410, opened 2024-11-07
feat: use scss styles from hds core components in cookieConsent

timwessman:
## Description

Use scss styles from hds core components in cookieConsent


mrTuomoK (approved):
ok, good work!

mrTuomoK (approved):
Ready for beta!

PR #1417, opened 2024-11-12
HDS-2495 cookie hds event namings

mrTuomoK:
## Description

Change event strings like below:

`
export const cookieEventType = {
  MONITOR: 'hds-cookie-consent-unapproved-item-found',
  READY: 'hds-cookie-consent-ready',
  CHANGE: 'hds-cookie-consent-changed',
};
`
Also changes `submitEvent` prop from string type to boolean and makes according changes across. It's also always set to true on React-side now.

Documentation changes are in another PR (cookie docs PR). 

## Related Issue

Closes [HDS-2495](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2495)

## Demos:

## Add to changelog

- not needed

[HDS-2495]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/cookieConsent/hooks/useCookieConsentEvents.ts
+++ b/packages/react/src/components/cookieConsent/hooks/useCookieConsentEvents.ts
@@ -1,6 +1,7 @@
 import { MutableRefObject, useEffect, useMemo, useRef } from 'react';
 
 import { isSsrEnvironment } from '../../../utils/isSsrEnvironment';
+import { cookieEventType } from '../../cookieConsentCore/cookieConsentCore';
 
 export type CookieConsentChangeEvent = {
   type: string;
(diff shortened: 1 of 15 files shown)


NikoHelle:
Just came to mind that the cookie consent core show export these, not React. React should only import them. 

And the `submitEvent` could only be a boolean. I see no need to make it changeable. Makes stuff harder for no reason when other event types are anyway locked.

If these are changed later, it's a breaking change 😬

packages/react/src/components/cookieConsent/hooks/useCookieConsentEvents.ts
```diff
@@ -12,17 +13,13 @@ export type CookieConsentEventsProps = {
...
 
-export const defaultSubmitEvent = 'cookie-consent-changed';
-export const monitorEvent = 'hds-cookie-consent-unapproved-item-found';
-export const readyEvent = 'hds_cookieConsent_ready';
-
 export function useCookieConsentEvents(props: CookieConsentEventsProps): CookieConsentEventsReturnType {
   const { onChange, onReady, onMonitorEvent } = props;
-  const submitEvent = props.submitEvent || defaultSubmitEvent;
+  const submitEvent = props.submitEvent || false;
   const listenerDisposer: MutableRefObject<(() => void) | null> = useRef(null);
```
NikoHelle, line 23: When submitEvent was a string, this had to be settable with props. Now that it is a boolean, it should (practically must be) always `true`. There is no scenario where React needs to reload whole page when cookies change or not to listen events.

NikoHelle (changes requested):
All good, expect in React submitEvent needs to always true

NikoHelle (approved):
Very nice! Looks minor changes, but was a battle 💪

PR #1436, opened 2024-11-21
remove Sketch & Abstract from CHANGELOG.template since obsolete

mrTuomoK:
Self explanatory :D remove Sketch & Abstract from CHANGELOG.template since obsolete


nooranen (approved):
Vihdoin ja viimein👍 Finally. Approved.

PR #1437, opened 2024-11-21
Hds 2352 fix missing component urls

timwessman:
## Description

Found small annoyance, if documentation version is changed on component page to version where it doesn't exists, documentation leads to error. Found out that there is not many of these kind of pages. So I made small routing fix to route these links to the root page of components (better than error page)

## How Has This Been Tested?

goto to page `components/button/` or `components/dropdown/` or `components/header/` or `components/error-summary/`and change documentation version, it should not crash.

### DIFF
--- a/site/src/components/layout.js
+++ b/site/src/components/layout.js
@@ -54,7 +54,7 @@ const fixUrlExceptions = (href, version) => {
 const hrefWithVersion = (href, version, withoutPrefix = false) => {
   const hrefWithFixedExceptions = fixUrlExceptions(href, version);
 
-  if (!version || version === '' || hrefWithFixedExceptions === ''
+  if (!version || hrefWithFixedExceptions === ''
     || hrefWithFixedExceptions.startsWith('mailto:')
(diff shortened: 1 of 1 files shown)


laurakarhu (approved):
Thx!

site/src/components/layout.js
```diff
@@ -22,14 +22,47 @@ import versions from '../data/versions.json';
...
+}
+
 const hrefWithVersion = (href, version, withoutPrefix = false) => {
-  if (!version || version === '' || href === ''
-    || href.startsWith('mailto:') || href.startsWith('#') || href.startsWith('http'))
-    return href;
+  const hrefWithFixedExceptions = fixUrlExceptions(href, version);
+
+  if (!version || version === '' || hrefWithFixedExceptions === ''
+    || hrefWithFixedExceptions.startsWith('mailto:')
```
NikoHelle, line 58: Just to make the long list shorter, `!version` includes `version === '' `

NikoHelle (approved):
One tiny thing that could be fixed.

PR #1443, opened 2024-11-25
HDS-2535: external modules listeners fix in Strict mode

NikoHelle:
## Description

Fix event listeners in external modules if beacon is re-created

In a scenario where external modules are used and the component is created multiple times, the listeners are set only once in the modules.

This happens in StrictMode.

On the second storeBeacon call the listeners in dedicated beacon are not added back and old ones are not removed.

The list of listeners must be kept and disposers must be called.

Now listeners are added again.

Listeners added after a beacon is created are not stored, because those are logically added when second beacon is created - as in first run.

## Related Issue

Closes [HDS-2535](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2535)

## How Has This Been Tested?

Added tests.

## Demos:

Login does not work in demo site.

## Add to changelog

- [x ] Added needed line to changelog

[HDS-2535]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Changes that are not related to specific components
 
 - [Login] External modules with Strict mode throwed an error
 - [ssr] Login component works with SSR rendering
+- [Login] Event listeners with Strict mode were lost in double render
 
 ### Core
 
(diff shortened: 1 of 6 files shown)


mrTuomoK (approved):
Good job!

timwessman (approved):

PR #1447, opened 2024-11-29
fix: buttons with icons renders incorrectly

timwessman:
## Description

Buttons with icons don't render correctly on core examples on the documentation page. These buttons render correctly in Storybook. 

I found out that the `::before` pseudo selector (which attaches icons to these buttons) doesn't work correctly because CSS rules are loaded in the wrong order. But only on the documentation page. 

The problem was in Webpack chunking configuration. Currently chunking optimization was turned off, because I haven't found any configuration that would work correctly (I have tried quite many). Now I found minimal configuration that do some chunk split optimization and still works. 

## Related Issue

Closes [HDS-2540](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2540)

## Demos:

Links to demos are in the comments.

## Add to changelog

Just change the documentation site build process, not the actual change to the code.

[HDS-2540]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -44,6 +44,8 @@ jobs:
       - name: build documentation site
         run: yarn build
         working-directory: ./site
+        env:
+          NODE_OPTIONS: "--max_old_space_size=8192"
 
       - name: Copy CNAME file
         run: cp CNAME ./site/public/
--- a/.github/workflows/hds-demo-preview.yml
+++ b/.github/workflows/hds-demo-preview.yml
@@ -81,6 +81,8 @@ jobs:
       - name: build site package
         run: yarn build --prefix-paths
         working-directory: ./site
+        env:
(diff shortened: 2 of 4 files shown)


timwessman:
Build breaks on `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` during site build process in Github's pipeline and in my local computer if it is built from root folder with Lerna. But if `yarn build` script is run from the `site` folder there is no problems. I'll investigate what is the difference

NikoHelle (approved):

mrTuomoK (approved):
Good fix 👍

PR #1471, opened 2025-02-06
Fix issues with controlled Select 

NikoHelle:
## Description

Fixed following issues:

- multiselect memoization issue: when selecting an option calls the `onChange` which then refreshes props and re-renders the component, the selected option keeps has old data in the memoized "trigger" function. When an option with old "trigger" reference is unselected, the whole data set is reverted to old state and selections made after it are cleared.
 
- crash if keyboard is used after `onChange` re-renders the component. Memoization issue in the `useKeyboard`. `onKeyDown` did not have correct dependencies.

- "open" state was not kept consistently when `onChange` caused re-render. When component props change, the `open` can be reverted to closed-state even if the component has internally opened it. This causes the multiselect close on each selection. The reset of the internal state is now handled by comparing old state with new props after a re-render.

- Multiselect groups with all options selected are not marked selected after re-render and clicking them had no effect.


timwessman (approved):
Excellent work, Niko! This helps us a lot. Thanks for your contribution. Now we have a state that doesn't mix up. Yep, the main problem was `onChange` in the state.

I rebased this on the latest `development`, fixed the conflict and updated reference images (as the  `clearable` property is now `false` as a default)

mrTuomoK (approved):
Good work, seems to fix the issues 👍

PR #1478, opened 2025-02-18
fix: (HDS-2651) only latest version of documentation for development …

timwessman:
## Description

Documentation site development got broken. 

FIX:
- moved heavy graphql queries out of `layout.js`. It's enough to query the data only once in `gatsby-node.js` where the query is done anyway. So now `siteData`, `currentMenuItem`, `uiMenuLinks` and `uiSubMenuLinks` is calculated in `gatsby-node.js` and transferred in `pageContext` to `layout.js`
- renamed `versions.json` to `versionsFromGit.json` to make it more logical to list only old versions here
- current documentation version is now only written in `package.json` and fetched from there to code
- when in development mode, only latest documentation is used 

## Related Issue

Closes [HDS-2651](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2651)

## Motivation and Context

To make development mode work. And when it works it should be lighter and faster.

## Demos:

## Add to changelog

- [] Added needed line to changelog

Only fixes for development mode


mrTuomoK (approved):
Seems to work, good job 💪

PR #1479, opened 2025-02-20
fix: (HDS-2561) cache site files for development

timwessman:
Adds cache for development build. So, documentation site build time files are cached


mrTuomoK (approved):
Nice 👍

PR #1512, opened 2025-04-29
HDS-2644: HDS Notification updates (for Helsinki Notification)

arpargo:
## Description

- [x] New body text size --fontsize-body-m all across the component
- [ ] Heading size and gaps in default/basic size inline notification
- [ ] Re-done Dismissable prop
- [ ] Body text content area needs to host a link

## Related Issue

Closes [HDS-2644](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2644)


site/src/docs/components/notification/code.mdx
mrTuomoK, line 1: The table is messed up, looks like this now :D 
CleanShot 2025-05-20 at 14 30 00@2x

packages/react/src/components/notification/Notification.tsx
```diff
@@ -66,6 +67,15 @@ type CommonProps = React.PropsWithChildren<
...
     label?: string | React.ReactNode;
+    /**
+     * Link to be displayed in the notification. If not provided, the link will not be displayed.
+     */
+    link?: string;
+    /**
+     * Link text to display. If not provided, the link will be displayed as plain text.
+     * Note: This prop is only used when the `link` prop is provided.
+     */
+    linkText?: string;
```
mrTuomoK, line 78: Was there a discussion of just giving the link as `Link`? We're now implementing the Tooltip that way too (a PR is open) in which the tooltip is given like `tooltip?: ReactElement<TooltipProps, typeof Tooltip>;` instead of all the separate tooltip-props.

mrTuomoK (changes requested):
Check the comments, at least the table is broken and I think we'd prefer to give the `Link` as a link-component instead of separate props.

mrTuomoK:
Also CHANGELOG changes are missing, no release-please in use in HDS.

packages/react/src/components/notification/Notification.tsx
arpargo: No but as the specs specified that this should only come supporting the medium size link I figured it's simpler to implement it this way instead of making it take in a Link element and then create restrictions on what type of a Link element it has to be.

mrTuomoK:
BTW does this support giving separate aria-label for the link? (by using the `Link` it would be possible 🤔 )

nooranen:
Also checked this and looks good to me👍 Good job! 

One thing I noticed on the docsite (not from this PR, it seems to be a published feature😅) Variations/Inline example has different sizes. These would look neater with just the default size:

Screenshot 2025-06-02 at 14 49 43

mrTuomoK (approved):
I approve but yeah I have my own commits here so a bit biased.

mikkojamG (approved):

PR #1517, opened 2025-05-20
fix: (FileInput) defaultValue typings

mrTuomoK:
## Description

There was an error when importing and using `hds-react` in other projects with error:

```shell
Type error: Type 'File[]' is not assignable to type '((string | number | readonly string[]) & File[]) | undefined'.
  Type 'File[]' is not assignable to type 'readonly string[] & File[]'.
    Type 'File[]' is not assignable to type 'readonly string[]'.
      Type 'File' is not assignable to type 'string'.
``` 

## How Has This Been Tested?

- importing and using in `hds-next`

## Add to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Changes that are not related to specific components
 #### Fixed
 
 - [Component] What bugs/typos are fixed?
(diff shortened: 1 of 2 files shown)


Riippi (approved):
👍

PR #1521, opened 2025-07-10
fix: issues with selections suddenly lost and mounting bug

mrTuomoK:
- issue with search & groupsLabels messing up selections
- more robust mount detection

Refs: HDS-2704

## Description

Selections could suddenly be lost at least when first selecting something searched and the searching something else and selecting more.

## Related Issue

Closes [HDS-2704](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2704)

## How Has This Been Tested?

- running all tests and manually doing the mentioned selections

## Demos:

## Add to changelog

[HDS-2704]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrTuomoK:
Still needs tests and examples... and did not even fix the issue in linkedcomponents-ui so it must be something with the project itself (looking into that now)

timwessman:
Good fix Tuomo! But as you said, this doesn't fix the problem in `linkedcomponents-ui`. I discovered that the problem is indeed a bug in our `Select` -component.

If a user selects some items from a search query and does another search, selections are removed if they are not part of the original `options` list or in the last search results. Now, this is fixed, so all selected items are preserved in the options list after new updates from search queries or when Formik or some other process updates the `options` argument.

- I made a new story `MultiselectWithSearch` to test this case. 
- Two tests for automatic tests. 
- I also run `linkedcomponents-ui` with this fixed HDS version to verify that it works now.

mikkojamG (approved):
Wonderful!

PR #1538, opened 2025-09-18
fix: stricter yarn + CODEOWNERS file

arpargo:
Quite simple .yarnrc change to prevent unmanaged package upgrades through dependencies.
Added CODEOWNERS file as is practice in Ratkis projects.


mrTuomoK (approved):
All good, thx!

PR #1547, opened 2025-10-14
fix: (HDS-2762) select keyboard navigation with virtualized mode

timwessman:
## Description

Jumping to the end of the option list didn't work whenever the Select component was used in virtualized mode. The last option is to read from the data, not from the rendered DOM items. So, as the component is rendered in chunks, it retries scrolling to the end until the last item is there.

## Related Issue

Closes [HDS-2762](https://helsinkisolutionoffice.atlassian.net/browse/HDS-2762)

[HDS-2762]: https://helsinkisolutionoffice.atlassian.net/browse/HDS-2762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react/src/components/select/hooks/useKeyboard.ts
+++ b/packages/react/src/components/select/hooks/useKeyboard.ts
@@ -2,7 +2,7 @@ import { KeyboardEvent, useCallback, useEffect, useMemo, useRef } from 'react';
 
 import { eventIds, eventTypes } from '../events';
 import { KnownElementType } from '../types';
-import { defaultFilter, filterSelectableOptions } from '../utils';
+import { defaultFilter, filterSelectableOptions, getAllOptions } from '../utils';
 import {
   useElementDetection,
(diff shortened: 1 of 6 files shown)


packages/react/src/components/select/hooks/useKeyboard.ts
```diff
@@ -11,6 +11,7 @@ import {
   isAnyListChildType,
 } from './useElementDetection';
 import { useSelectDataHandlers } from './useSelectDataHandlers';
+import { getAllOptions } from '../../modularOptionList/utils';
```
mikkojamG: Looks like this incorrect import is breaking the build.

mrTuomoK:
I'm still not very fond of polling the elements from the dom 🤔

